PR #39, opened 2022-08-07
Compare to other semver-checking tools.

### LINKED ISSUE #21: Doc request: compare semverver
[semverver](https://lib.rs/crates/semverver) is another tool for checking semver violations. Its approach at a high level is to compare rlibs rather than rustdocs.

semverver was originally published as part of GSoC 2017, but has recently been adopted into the rust-lang org to hopefully keep it functional.

Obviously there will be differences in what changes both tools do and don't catch, and keeping a scorecard of that probably isn't really worth the effort. It would be nice, however, to call out semverver as an alternative and list some high-level differences resulting from the different approaches. As an initial intuition, that would include at least

- semver-checks by design only checks rustdoc-exposed information, so ignores any `#[doc(hidden)]` items, which semverver has access to to check.
- semver-checks works off of (mostly?) stable metadata from rustdoc; semverver is by design tied to only work on a specific nightly.

Personally, I think diffing the documented surface *is* the "more correct" approach here, as semver is about public exposed API. Relying on details not contained in the documentation is relying on details outside of the semver guarantee.

obi1kenobi:
Resolves #21.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,17 +61,115 @@ If you'd like to perform those steps manually, here they are:
    published v1.2.2 and v1.3.0, and you need to backport some fixes and release v1.2.3:
    1.2.2 would be your baseline, and you'd compare 1.2.3 -> 1.2.2 and not 1.2.3 -> 1.3.0.
 
-Notes:
-- If using it on a massive codebase (multiple hundreds of thousands of lines of Rust),
-  the queries may be a bit slow: there is some `O(n^2)` scaling for `n` items in a few places that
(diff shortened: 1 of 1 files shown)


obi1kenobi:
cc @CAD97 if you have a few minutes. I'm inclined to merge this promptly on the basis of "something is better than nothing" especially since "how does this compare" is everyone's first question, but I'd love your feedback even post-merge and would be happy to address it with future PRs.

PR #64, opened 2022-08-10
Add mentorship issue label.

obi1kenobi:


obi1kenobi:
Am I doing this right?

epage:
> Inspired by the rust-lang/rust label. I'm interested in offering mentoring for writing some of the easier remaining semver queries.

I leave this to you for what you want.  Personally, I don't tend to split out mentorship as a label because, at my scale of projects, that is implied with `E-help-wanted`.

For a more complicated, like clap, I do the following
- Several `S-` (status) labels are used to track where in the decision making process an issue is
- Once a decision is made, I switch to an `E-` effort tag to indicate that its "accepted"
  - `E-help-wanted` are meant for new contributors

See https://github.com/clap-rs/clap/blob/master/.github/settings.yml for the full list (though currently not using Settings bot for it)

> Am I doing this right?

Yup

obi1kenobi:
I like the idea of explicitly mentioning mentorship. You and I might understand that `E-help-wanted` means "I'm happy to mentor folks" but I'm not confident that an arbitrary person visiting the repo for the first time might see it the same way — especially if this is the first repo with an elaborate label system they're considering contributing to.

Thanks for checking it out! I might add more `S-` and `E-` difficulty labels as they become useful.

PR #66, opened 2022-08-11
docs(readme): Focus on user/contrib entry points

epage:
At least to me, some of the README topics seemed to jump between the
target audience being a user vs being a contributor.  This splits them
up and tries to establish a flow.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,101 @@
+# Contributing
+
+## Design goals
+
+In short:
+- Checks should be configuration, not code.
+- That helps us ensure we don't have to trade off ergonomics versus maintainability.
+
+To make a semver-checker that is pleasant to use (and therefore gets widely adopted),
+we have to go beyond being merely "technically correct" when reporting problems.
+
+For example, say the tool has discovered that a `pub struct` no longer implements some trait:
+this is a breaking change and semver requires a major version bump. It's technically correct to
+state this fact and move on, but it's more helpful to have contextually-appropriate advice and
+reference links based on whether the trait in question is:
+- an auto-trait like `Send`, `Sync`, or `Sized`
+- a trait that is usually added via `#[derive(...)]`, like `Debug` or `Clone`
+- a built-in trait that is usually not derived, like `From`
+- one of the crate's own traits
+
+If all our semver checks were written imperatively, it would have been difficult to reuse code
+and optimizations across different checks. This would have incentivized having a single overarching
+"trait is missing" check with a ton of special cases, i.e. complex code with a maintainability
+hazard.
+
+Checks should be configuration, not code, and that's what `cargo-semver-checks` does.
+It uses a datasource-agnostic query engine called
+[Trustfall](https://github.com/obi1kenobi/trustfall) to allow writing semver checks as
+[declarative strongly-typed queries](https://twitter.com/PredragGruevski/status/1550135974499438592)
+over a
+[schema](https://github.com/obi1kenobi/cargo-semver-check/blob/main/src/rustdoc_schema.graphql).
+Adding a new semver check is as simple as adding a new file that specifies the query to run and
+metadata like the error message to display in case the query finds any results (errors).
+
+This has several advantages:
+- **It's easy to write more checks or specialize existing ones.** Just duplicate an existing
+  query file and edit it to your liking. The strongly-typed query language doesn't prevent
+  logic errors (neither does Rust 😅), but like Rust it has a strong tendency
+  to "work correctly as soon as it compiles."
+- **Fast performance without complex code.** Trustfall enables efficient lazy evaluation of queries
+  without any cloning of rustdoc JSON data and without `unsafe`. The obvious way to write queries
+  is also the fast way.
+- **Optimizations are decoupled from queries.** When a new optimization (e.g. some caching)
+  is added to  `cargo-semver-checks` (or even Trustfall itself), all queries automatically
+  benefit from it without needing any changes.
+
+### Future functionality made easier by the query-based approach
+
+In principle, `cargo-semver-checks` could be extended to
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,107 @@
...
+
+`cargo-semver-checks` queries `rustdoc`-generated crate documentation using the `trustfall`
+["query everything" engine](https://github.com/obi1kenobi/trustfall).
+Each query looks for a particular kind of semver violation, such as:
+- public struct was removed
+- public enum's variant was removed
+- public struct is no longer `Send/Sync/Debug/Clone` etc.
+- public enum has a new variant, but wasn't non-exhaustive in the last version
+
+## Design goals
```
obi1kenobi, line 11: What do you think about moving the design goals to a separate file, maybe called `DESIGN.md` or similar? It feels a bit strange that the contributing doc starts with a long section that isn't really directly about contributing, and doesn't really address why that is the case.

README.md
```diff
@@ -1,177 +1,78 @@
...
+sync](https://github.com/rust-lang/rust-semverver/search?q=Rustup+to&type=commits).
+
+[cargo breaking](https://github.com/iomentum/cargo-breakinga) effectively runs
+`cargo expand` and re-parses the code using
+[`syn`](https://crates.io/crates/syn) which requires re-implementing large
+swaths of rust's semantics to then lint the API for changes.
+This is limited to the feature and target the crate was compiled for.
+
+`cargo-semver-checks` sources its data from rustdoc's json output.  While
+unstable, the rate of change is fairly low, reducing the churn in keeping up.
```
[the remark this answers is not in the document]
obi1kenobi: Just trying to clarify that it's the JSON format that is unstable, not `cargo-semver-checks`.

README.md
```diff
@@ -1,177 +1,78 @@
...
+The lints are also written ontop of the `trustfall` ["query everything"
+engine](https://github.com/obi1kenobi/trustfall), reducing the work for writing
+and maintaining lints.  Because of the extra data that rustdoc includes, some
+level of feature/target awareness might be able to be introduced.
```
obi1kenobi, line 68:
```suggestion
The lints are also written as queries for the `trustfall` ["query everything"
engine](https://github.com/obi1kenobi/trustfall), reducing the work for adding
and maintaining lints.  Because of the extra data that rustdoc includes, some
level of feature/target awareness might be able to be introduced.
```
Just trying to clarify a bit. English is harder than Rust :)

If you see a way to reword this so the first sentence doesn't both start and end with `lints`, that'd be great. I'm getting tripped up on it but I couldn't figure out a great way out of that.

README.md
```diff
@@ -1,177 +1,78 @@
...
+
+There is interest in
 [hosting rustdoc JSON](https://github.com/rust-lang/docs.rs/issues/1285) on `docs.rs` meaning
 that semver-checking could one day download the baseline rustdoc JSON file instead of generating it.
 Also, generally speaking, inspecting JSON data is likely going to be faster than full compilation.
 
-### Comparing `cargo-semver-checks` vs `cargo-crate-api` and `cargo-public-api`
```
obi1kenobi, line 94: You don't think it's worth comparing to `cargo-crate-api` and `cargo-public-api`? I like the "one short paragraph per tool" format you set up, and it might not be too verbose to have a single extra paragraph for the two tools.

CONTRIBUTING.md
```diff
@@ -0,0 +1,107 @@
+# Contributing
```
obi1kenobi, line 1: Either here or in a future PR, it might be good to structure this doc around the different things people could contribute with and where to find them (e.g. improving docs vs adding new lints vs improving the test setup) and meta advice like "look for `E-mentor` and `E-help-wanted` labels."

What do you think?

obi1kenobi (approved):
This is definitely an improvement in the docs, and I'd be happy to merge it. I left a few comments and suggestions, which you're free to ignore, address in this PR, or leave for a future PR, as you think is best.

CONTRIBUTING.md
epage: Yup, we could definitely use more stuff like that.  This was focused more on restructuring what we already had.

README.md
```diff
@@ -1,177 +1,78 @@
...
+- If using it on a massive codebase (multiple hundreds of thousands of lines of Rust),
+  the queries may be a bit slow: there is some `O(n^2)` scaling for `n` items in a few places that
+  I haven't had time to optimize down to `O(n)` yet. Apologies! I have temporarily prioritized
+  features over speed, and the runtime will improve significantly with a small amount of extra work.
+- **No false positives**: Currently, all queries report constructive proof of semver violations:
+  there are no false positives. They always list a file name and line number for the baseline item
+  that could not be found in the new code.
+- **There are false negatives**: This tool is a work-in-progress, and cannot check all kinds of
+  semver violations yet. Just because it doesn't find any semver issues doesn't mean
+  they don't exist.
```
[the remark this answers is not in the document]
epage: You actually had this twice in two different forms, so figured it was important :)

I think these kind of details are useful.  We just might polish it further in the future.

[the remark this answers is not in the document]
obi1kenobi: More likely the result of rushing to get everything written down than actual intent in this case 😅

README.md
epage: `cargo-crate-api` was a PoC for a rustdoc-based tool.  I've already archived the repo and will just move into this what I find useful.

I'll add back in `cargo-public-api`

CONTRIBUTING.md
epage: Maybe in the future?  DESIGN is a subset of CONTRIBUTING information so as we grow, it might become important but then we have to help people discover these files (for some use cases) so keeping things focused when smaller I think helps.

PR #71, opened 2022-08-15
Rerun tests every workday to check for nightly JSON breaks.

### LINKED ISSUE #70: Set up Action to periodically check compatibility with new nightly rustdoc JSON formats
Since the rustdoc JSON data format is unstable, new nightly Rust releases will change the output and make this tool no longer work until it is adapted for the new output format.

Set up a GitHub Action to periodically re-run the test suite with the latest Rust nightly so we can discover these problems as early as possible.

obi1kenobi:
Resolves #70.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,8 @@ on:
   push:
     branches:
       - main
+  schedule:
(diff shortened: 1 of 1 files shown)


obi1kenobi:
@epage wondering if you have opinions on #70 and this PR. In the spirit of "something is better than nothing" I'm inclined to merge first and update afterward as needed.

Issue #54, opened 2022-08-09
Mark the rustdoc flags as perma-unstable, behind a `-Z` flag


epage:
Requires at least
- #53 
- #52 

But would also benefit from #51, #50

PR #77, opened 2022-08-17
refactor(cli): Switch to clap_derive

epage:
This will make it easier to evolve the CLI for issues like #54

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -7,10 +7,9 @@ mod query;
 mod templating;
 mod util;
 
-use std::env;
 use std::path::PathBuf;
 
-use clap::{crate_version, value_parser, Arg, ArgAction, Command};
+use clap::{Args, Parser, Subcommand};
 use termcolor::{ColorChoice, StandardStream};
 
 use crate::{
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -52,117 +51,66 @@ impl GlobalConfig {
...
+
+#[derive(Args)]
+struct CheckRelease {
+    /// The current rustdoc json output to test for semver violations. Required.
+    #[clap(short, long = "current", value_name = "CURRENT_RUSTDOC_JSON")]
+    current_rustdoc_path: PathBuf,
+
+    /// The rustdoc json file to use as a semver baseline. Required.
+    #[clap(short, long = "baseline", value_name = "BASELINE_RUSTDOC_JSON")]
+    baseline_rustdoc_path: PathBuf,
```
obi1kenobi, line 109: Since these two subcommands are identical (though poorly documented as such), is there a way to have them use an identical struct instead of two different ones?

epage: https://github.com/obi1kenobi/cargo-semver-check/pull/78 is the follow up to this comment

PR #78, opened 2022-08-17
fix: Soft-deprecate 'diff-files' subcommand

epage:
This makes it a hidden alias of `check-release`.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -56,12 +56,6 @@ fn main() -> anyhow::Result<()> {
     let config = GlobalConfig::new();
 
     match args {
-        SemverChecks::DiffFiles(args) => {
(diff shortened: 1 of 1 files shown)


obi1kenobi (commented):
Awesome! I'm learning a lot just reading your PRs — thank you! :+1:

epage:
Yeah, clap's API is too big, hiding many features.  One of the things we are working towards is providing fewer but composable features so its easier to see what is possible.

obi1kenobi:
I'm also just not that familiar with clap or many of the other "common" crates in the Rust ecosystem yet, since I'm largely self-taught in terms of Rust and don't really use it in my day job. This tool is actually my first time using clap at all.

PR #79, opened 2022-08-17
refactor(cli): Reduce number of exit paths

epage:
In working on other changes, I found `run_check_release` exiting in some cases hard to account for.  I've generally found it best to centralize all exits.

If we want to extend this further, a pattern I've found useful is to use the [proc_exit](https://docs.rs/proc-exit/) crate.


obi1kenobi (commented):
:+1:

PR #80, opened 2022-08-17
feat(cli): Generate rustdoc automatically

### LINKED ISSUE #52: Add `--base-manifest-path <path>` argument
Point to another path to compare APIs

This is exclusive with #50, #51

epage:
The original command-line will work as usual but now
- If you don't specify `--current`, it will infer the packages like any other cargo command
- Instead of `--baseline`, you can specify `--baseline-root` and it will search it for relevant manifests for what is in current

clap is enforcing the relationships to make this all work.

I've played with integration tests for this but we'll have to decide how well we want to handle that going forward.  I was originally going to do integration tests against all of cargo-semverver's test cases but it was pretty slow :).  I did hand test by having two clap directories set to different commits.

`--baseline-git` will build on this, we'll just clone first.  `--baseline-version` will require a bit more work.

This is best review per-commit to see how it all builds up

Fixes #52

### DIFF
--- a/src/dump.rs
+++ b/src/dump.rs
@@ -87,7 +87,7 @@ impl RustDocCommand {
             }
         }
 
-        let json_path = target_dir.join(format!("doc/{}.json", crate_name));
+        let json_path = target_dir.join(format!("doc/{}.json", crate_name.replace('-', "_")));
         if !json_path.exists() {
             anyhow::bail!(
                 "Could not find expected rustdoc output: {}",
--- a/src/dump.rs
+++ b/src/dump.rs
@@ -90,7 +90,8 @@ impl RustDocCommand {
         let json_path = target_dir.join(format!("doc/{}.json", crate_name.replace('-', "_")));
         if !json_path.exists() {
             anyhow::bail!(
-                "Could not find expected rustdoc output: {}",
+                "Could not find expected rustdoc output for `{}`: {}",
+                crate_name,
                 json_path.display()
             );
         }
--- a/src/dump.rs
+++ b/src/dump.rs
@@ -1,0 +1,105 @@
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct RustDoc {
+    deps: bool,
+    silence: bool,
+}
+
+impl RustDoc {
+    pub fn new() -> Self {
+        Self {
+            deps: false,
+            silence: false,
+        }
+    }
+
+    /// Include dependencies
+    ///
+    /// Reasons to have this disabled:
+    /// - Faster API extraction
+    /// - Less likely to hit bugs in rustdoc, like
+    ///   - rust-lang/rust#89097
+    ///   - rust-lang/rust#83718
+    ///
+    /// Reasons to have this enabled:
+    /// - Check for accidental inclusion of dependencies in your API
+    /// - Detect breaking changes from dependencies in your API
+    pub fn deps(mut self, yes: bool) -> Self {
+        self.deps = yes;
+        self
+    }
+
+    /// Don't write progress to stderr
+    pub fn silence(mut self, yes: bool) -> Self {
+        self.silence = yes;
+        self
+    }
+
+    pub fn dump(&self, manifest_path: &std::path::Path) -> anyhow::Result<std::path::PathBuf> {
+        let crate_name = crate::manifest::get_package_name(manifest_path)?;
+
+        let metadata = cargo_metadata::MetadataCommand::new()
+            .manifest_path(manifest_path)
+            .no_deps()
+            .exec()?;
+        let manifest_target_directory = metadata
+            .target_directory
+            .as_path()
+            .as_std_path()
+            // HACK: Avoid potential errors when mixing toolchains
+            .join("semver-check/target");
+        let target_dir = manifest_target_directory.as_path();
+
+        let stderr = if self.silence {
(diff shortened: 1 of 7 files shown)


src/dump.rs
```diff
@@ -0,0 +1,105 @@
...
+        } else {
+            // Print cargo doc progress
+            std::process::Stdio::inherit()
+        };
+
+        let mut cmd = std::process::Command::new("cargo");
+        cmd.env(
+            "RUSTDOCFLAGS",
+            "-Z unstable-options --document-hidden-items --output-format=json",
+        )
```
obi1kenobi, line 63: Interesting! I don't know enough to have an opinion on this, I'm just curious why these are passed through the env instead of in `.args()`

src/dump.rs
```diff
@@ -0,0 +1,105 @@
...
+            }
+        }
+
+        let json_path = target_dir.join(format!("doc/{}.json", crate_name));
+        if !json_path.exists() {
+            anyhow::bail!(
+                "Could not find expected rustdoc output: {}",
+                json_path.display()
+            );
+        }
```
obi1kenobi, line 96: Do we need to worry about dashes vs underscores here? The GitHub Action has to replace dashes with underscores in the crate name to get the JSON file name: https://github.com/obi1kenobi/cargo-semver-checks-action/blob/main/action.yml#L56-L68

src/main.rs
```diff
@@ -61,15 +60,20 @@ fn main() -> anyhow::Result<()> {
...
+                };
             let rustdoc = dump::RustDoc::new().deps(false).silence(false);
 
             let rustdoc_paths =
                 if let Some(current_rustdoc_path) = args.current_rustdoc_path.as_deref() {
                     vec![(
-                        loader.load_rustdoc("<unknown>")?,
+                        loader.load_rustdoc(&rustdoc, "<unknown>")?,
                         current_rustdoc_path.to_owned(),
                     )]
```
obi1kenobi, line 78: If I'm reading this correctly, it seems like this is always generating new rustdoc JSON since new line 71 is always hit, even if there's a `current_rustdoc_path` set. That doesn't seem right to me?

src/dump.rs
```diff
@@ -0,0 +1,105 @@
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct RustDoc {
+    deps: bool,
+    silence: bool,
+}
+
+impl RustDoc {
```
obi1kenobi, line 7: Consider tweaking the name here to communicate that this is a type that can _generate_ rustdoc JSON, rather than a type that represents _already-generated_ rustdoc JSON?

Otherwise, it feels strange to see what appears to be the "current" rustdoc JSON instance being given to the baseline rustdoc JSON generator.

obi1kenobi (commented):
Lots of good stuff here! Excited to get it merged. Left a few minor suggestions here and there.

src/main.rs
epage: line 71 just creates the rustdoc wrapper.  The active `loader` is what determines whether rustdoc is generated or not.

src/dump.rs
epage: Modeling off of `cargo_metadata`, calling this `RustDocCommand` seems appropriate.

src/dump.rs
epage: At the time, that was how I was used to sending arguments directly to rustdoc through cargo-doc.  I'm assuming it supports `--` like `cargo test` and everything else.

I can go either way on how its implemented in practice.

src/main.rs
obi1kenobi: Changing `RustDoc` to `RustDocCommand` helps a lot here, thanks! To simplify the thought process of the next person to come across this, consider leaving a comment next to `let rustdoc` mentioning that the command is prepared but not executed, and that the loader gets to choose whether to run it or not.

src/dump.rs
```diff
@@ -87,7 +87,7 @@ impl RustDocCommand {
             }
         }
 
-        let json_path = target_dir.join(format!("doc/{}.json", crate_name));
+        let json_path = target_dir.join(format!("doc/{}.json", crate_name.replace('-', "_")));
```
obi1kenobi, line 90: This looks good. In case there's a problem with the crate name -> JSON file lookup expectations themselves, consider including the crate name in the `anyhow::bail!()` call together with the `json_path`.

src/dump.rs
obi1kenobi: No preference either way. If this works without issues, no need to change it.

src/main.rs
epage: I renamed the variable to clarify this is the command.  This should clarify that it still needs a call done on it to actually execute.

obi1kenobi:
All this looks good, my remaining comments are nits that I feel like are pretty straightforward and don't need an in-depth review.

If you'd like to merge this, go for it :+1: If you'd prefer to keep polishing (docs, CLI help text, etc.) on this branch before merging, that's fine by me too.

epage:
> If you'd like to merge this, go for it 👍 If you'd prefer to keep polishing (docs, CLI help text, etc.) on this branch before merging, that's fine by me too.

I lean towards smaller, independent iterations rather than upfront perfection.  Feel free to fix or open issues in anything you see.  Otherwise, I'll probably move on to git support and then registry support

PR #83, opened 2022-08-18
fix(cli): Make output behave more like cargo

epage:
Main changes
- Move status messages to stderr
- rustdoc output is now behind `--verbose`'
- For users concerned over long rustdoc run without output, we now print a status message for it
- Add crate name to per-crate start message
- Put per-crate check counts behind `--verbose`
- Put each check's run behind `--verbose`

To more easily match cargo's behavior (and make it easier to integrate with cargo in the future), this uses an API inspired by `cargo::core::Shell` to manage common message formats and behavior.  This was pulled out of cargo-edit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -149,8 +149,10 @@ dependencies = [
  "cargo_metadata",
  "clap",
  "clap-cargo",
+ "clap-verbosity-flag",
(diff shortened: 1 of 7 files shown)


obi1kenobi (commented):
👍

PR #85, opened 2022-08-18
fix(cli): Add more parsing details

epage:
- Now both parses get a message, not just the second
- Now we clarify which parse operation is happening

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -61,8 +61,9 @@ fn main() -> anyhow::Result<()> {
                 let mut rustdoc_paths = Vec::with_capacity(selected.len());
                 for selected in selected {
                     let manifest_path = selected.manifest_path.as_std_path();
-                    let rustdoc_path = rustdoc_cmd.dump(manifest_path)?;
                     let crate_name = manifest::get_package_name(manifest_path)?;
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -61,8 +61,9 @@ fn main() -> anyhow::Result<()> {
                 let mut rustdoc_paths = Vec::with_capacity(selected.len());
                 for selected in selected {
                     let manifest_path = selected.manifest_path.as_std_path();
-                    let rustdoc_path = rustdoc_cmd.dump(manifest_path)?;
                     let crate_name = manifest::get_package_name(manifest_path)?;
+                    config.shell_status("Parsing", format_args!("{} (current)", crate_name))?;
```
obi1kenobi, line 65: It feels a bit strange to have a "Parsing" message right before generating the JSON which is to be parsed. Consider "Generating" or another word choice?

epage: "Generating rustdoc" and then "Parsing rustdoc" feels like an implementation detail that doesn't matter to the end user.

If I run `cargo semver-checks check-release --baseline-root ../foo`, I have no idea rustdoc is involved.  I do know it needs to parse the source base to extract data about the API.

This doesn't mean "Parsing" is the ideal word, just trying to put the focus on the users perspective.

Issue #98, opened 2022-08-23
cargo semver-checks fails with "Failed to parse rustdoc JSON output"


ericseppanen:
I was trying cargo-semver-check for the first time on a small project of mine and this was the result:

```txt
$ cargo semver-checks check-release
    Updating index
     Parsing escape8259 v0.5.1 (current)
     Parsing escape8259 0.5.0 (baseline)
Error: Failed to parse rustdoc JSON output file "/home/eric/work/rust/escape8259/target/semver-checks/registry-escape8259-0_5_0/target/semver-checks/target/doc/escape8259.json"

Caused by:
    missing field `name` at line 1 column 774
```

My project is [here](https://github.com/ericseppanen/escape8259); my working directory is clean and I have the top of the master branch checked out (09529d6).

My default toolchain is 1.63, and I cargo-semver-check is 0.9.0.

epage:
What is your nightly version?

obi1kenobi:
I think #88 is likely, because:
- this is the same kind of error I got when triaging what happened the last time the rustdoc JSON format had a breaking change
- the JSON format hasn't changed since that nightly, since our scheduled builds with latest nightly have all passed: https://github.com/obi1kenobi/cargo-semver-check/actions/runs/2909680210

@ericseppanen, `cargo-semver-checks` needs nightly because the rustdoc JSON capability we use is still unstable and nightly-only. On your end, upgrading to a newer nightly should fix the problem. On our end, we need to fix #88 and also provide a better error message here.

If you're able to check if upgrading to the latest nightly resolves this problem for you and let us know, that'd be great. If that's the case, then I'd like to use this issue as the "improve the error message when the JSON format doesn't match" issue.

ericseppanen:
Thanks for the help!

I was on `rustc 1.65.0-nightly (29e4a9ee0 2022-08-10)`, and after updating to `rustc 1.65.0-nightly (060e47f74 2022-08-23)` it works.

PR #99, opened 2022-08-23
feat(cli): Help users in writing bug reports

epage:
This does two things
- Provides a crash handler
- Adds a `--bugreport` flag for capturing extra information

The hope is to help with cases like #98


obi1kenobi (commented):
Amazing! I was just thinking that we should add something like this, but I didn't know about these awesome crates so I couldn't have done this good of a job 💯

PR #100, opened 2022-08-23
chore(gh): Add issue templates

epage:
The main goal is to raise awareness of `--bugreport` once its released.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,30 @@
+name: Bug report
+description: Things didn't meet expectations
+labels: 'C-bug'
+body:
+  - type: textarea
+    attributes:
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,30 @@
...
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Additional Context
+      description: Add any other context about the problem here.
+  - type: textarea
+    attributes:
+      label: Debug Output
+      description: "Run `cargo semver-checks --bugreport`'
```
epage, line 30: We need to do a release first

obi1kenobi: Looks good to me. I'm in favor of releasing whenever you think is right :+1:

PR #103, opened 2022-08-23
Fix mismatched quotes in bug_report.yml template.

obi1kenobi:


obi1kenobi:
This is a new bit of UI that I don't recall noticing before and that was helpful:
image

I just wish it appeared directly on the PR itself where this file was added.

Issue #88, opened 2022-08-18
Document nightly requirements


epage:
At the moment, rustdoc is coupled to a specific range of nightly versions.  We should clearly document this for users (and maybe provide helpful runtime errors about it) so users who let their nightly versions go stale have a positive path forward.

obi1kenobi:
`cargo-semver-checks` no longer needs nightly, and its support for current stable + beta is documented in the README.

This issue seems okay to close to me, but if there's something I've missed then feel free to reopen.

PR #106, opened 2022-08-24
Give actionable advice when rustdoc JSON files fail to parse.

### LINKED ISSUE #105: Improve error message when failing to parse rustdoc output
Every so often, new Rust nightlies might cause the rustdoc JSON output format to change in a backward-incompatible way.

`cargo-semver-checks` will remain compatible with the most recent nightly's output, and we have a test that ensures that. However, users may still be using an older (or newer!) nightly that is incompatible with their `cargo-semver-checks` version.

If the JSON fails to parse, we should:
- check the nightly version to see if it's known to not work because it's too old (related: #88), then advise the user to upgrade to a newer nightly Rust
- otherwise, the nightly may be too new for the `cargo-semver-checks` version and we should advise the user to try upgrading to a newer `cargo-semver-checks` (as in #98)

obi1kenobi:
Resolves #105.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -36,23 +36,17 @@ pub(crate) fn load_rustdoc_from_file(path: &Path) -> anyhow::Result<Crate> {
 
             // The error on this line is case (1).
             let version = serde_json::from_str::<RustdocFormatVersion>(&s).with_context(|| {
-                format!(
-                    "parsing rustdoc JSON failed, and could not determine rustdoc format version \
-                    for file {}",
-                    path.display(),
-                )
+                format!("unrecognized rustdoc format for file {}", path.display(),)
             })?;
 
             match version.format_version.cmp(&rustdoc_types::FORMAT_VERSION) {
                 std::cmp::Ordering::Less => {
                     // The error here is case (2).
                     bail!(
-                        "rustdoc output format is too old (v{1}, need v{0}) for file \
-                        {2}\n\n\
-                        \
-                        note: Using a newer Rust nightly version should help. \
-                        The latest cargo-semver-checks and Rust nightly should always be \
-                        compatible with each other.",
+                        "\
+rustdoc output format is too old (v{1}, need v{0}) for file {2}
+
+note: using a newer Rust nightly version should help",
                         rustdoc_types::FORMAT_VERSION,
                         version.format_version,
                         path.display(),
@@ -61,12 +55,10 @@ pub(crate) fn load_rustdoc_from_file(path: &Path) -> anyhow::Result<Crate> {
                 std::cmp::Ordering::Greater => {
                     // The error here is case (3).
                     bail!(
-                        "rustdoc output format is too new (v{1}, need v{0}) for this \
-                        cargo-semver-checks when parsing {2}\n\n\
-                        \
-                        note: A newer version of cargo-semver-checks is likely available. \
-                        The latest cargo-semver-checks and Rust nightly should always be \
-                        compatible with each other.",
+                        "\
+rustdoc output format is too new (v{1}, need v{0}) when parsing {2}
+
+note: a newer version of cargo-semver-checks is likely available",
                         rustdoc_types::FORMAT_VERSION,
                         version.format_version,
                         path.display(),
@@ -74,11 +66,11 @@ pub(crate) fn load_rustdoc_from_file(path: &Path) -> anyhow::Result<Crate> {
                 }
                 std::cmp::Ordering::Equal => Err(e).with_context(|| {
                     format!(
-                        "parsing rustdoc failed even though the format was the expected \
-                            version (v{}) for file {}\n\n\
-                            \
-                            note: This is a bug. Please report it on the cargo-semver-checks \
-                            GitHub together with the output of `cargo-semver-checks --bugreport`.",
+                        "\
+unexpected parse error for v{} rustdoc for file {}
+
+note: this is a bug, please report it on the tool's GitHub together with \
+the output of `cargo-semver-checks --bugreport`",
                         rustdoc_types::FORMAT_VERSION,
                         path.display(),
                     )
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,10 +1,16 @@
 use std::path::Path;
 use std::{fs::File, io::Read};
 
-use anyhow::Context;
+use anyhow::{bail, Context};
 use rustdoc_types::Crate;
+use serde::Deserialize;
 
 pub(crate) const SCOPE: &str = "semver-checks";
+
+#[derive(Deserialize)]
+struct RustdocFormatVersion {
+    format_version: u32,
+}
 
 pub(crate) fn load_rustdoc_from_file(path: &Path) -> anyhow::Result<Crate> {
     // Parsing JSON after fully reading a file into memory is much faster than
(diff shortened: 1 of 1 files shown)


obi1kenobi:
@epage would you say this is also sufficient for #88 or did you have something else in mind for that one? I wasn't sure if information about which nightly ranges use which rustdoc format versions is available somewhere, for example.

src/util.rs
```diff
@@ -16,8 +22,69 @@ pub(crate) fn load_rustdoc_from_file(path: &Path) -> anyhow::Result<Crate> {
...
+                        both cargo-semver-checks and Rust nightly, \
+                        which should always be compatible with each other.",
+                        rustdoc_types::FORMAT_VERSION,
+                        version.format_version,
+                    )
+                }
+                std::cmp::Ordering::Greater => {
+                    // The error here is case (3).
+                    bail!(
+                        "Failed to parse rustdoc JSON file: {path:?}\n\n\
```
epage, line 66: As a reviewer, escaped newlines make it harder to review the legibility of these error messages.

src/util.rs
```diff
@@ -16,8 +22,69 @@ pub(crate) fn load_rustdoc_from_file(path: &Path) -> anyhow::Result<Crate> {
...
+                        This version of cargo-semver-checks requires rustdoc JSON version {0}, \
+                        but the JSON file uses the newer version {1}.\n\n\
```
epage, line 68: I feel like these messages could be more concise so the user can more easily find the information they need.

Something like
```
rustdoc output is too new (version {}) for cargo-semver-checks when parsing {}

note: a newer version of cargo-semver-checks is likely available
```
- cargo's style guide says to not capitalize error messages
- rustc's style guide says to avoid "did you mean" or "suggestions" but to instead offer people alternatrives (ie be less prescriptive) which is why I made it a "note:".  cargo doesn't follow this atm but I suspect we should conform to it.

src/util.rs
```diff
@@ -16,8 +22,69 @@ pub(crate) fn load_rustdoc_from_file(path: &Path) -> anyhow::Result<Crate> {
...
+            // Several possible options and their resolutions:
+            // (1) The file isn't actually a rustdoc JSON file. The user should supply a valid file.
+            // (2) The rustdoc JSON file has a version number that is too old, and isn't supported.
+            //     The user should upgrade to a newer nightly Rust version and regenerate the file.
+            // (3) The rustdoc JSON file has a version number that is too new, and isn't supported.
+            //     The user should attempt to upgrade to a newer cargo-semver-checks version
+            //     if one is already available, or open a GitHub issue otherwise.
+
+            // The error on this line is case (1).
+            let version = serde_json::from_str::<RustdocFormatVersion>(&s).with_context(|| {
```
epage, line 38: > @epage would you say this is also sufficient for https://github.com/obi1kenobi/cargo-semver-check/issues/88 or did you have something else in mind for that one? I wasn't sure if information about which nightly ranges use which rustdoc format versions is available somewhere, for example.

The core of #88 is the install instructions.

Something to keep in mind is some users might be stuck on a specific nightly version
- Ideally, we'd let them know what range of versions work rather than just saying "use latest"
- Ideally, they'd have a path for finding what older version would work

While the error reporting is good and important, the reason the documentation is important is that an error is a road bump and the more road bumps someone hits to accomplishing their task, the more frustrated they get.  #88 is about avoiding that.

src/util.rs
```diff
@@ -16,8 +22,69 @@ pub(crate) fn load_rustdoc_from_file(path: &Path) -> anyhow::Result<Crate> {
...
+                format!(
+                    "Failed to parse the rustdoc JSON file, and could not determine its \
+                    format version. Are you sure this is a valid rustdoc JSON file? \
+                    File path: {:?}",
+                    path
```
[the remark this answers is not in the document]
obi1kenobi: `Path` isn't `Display`, and its `Debug` implementation just prints the path in quotes. This is what the errors currently look like:
image

Is there a way to turn this `Path` value into something that is `Display`? I know my path handling is sloppy :)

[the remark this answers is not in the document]
epage: `path.display()`

src/util.rs
obi1kenobi: There's a bit of tension here and I'm not sure how to navigate it:
- If I use unescaped newlines and omit the trailing `\`, then the Rust string contains 24 spaces' worth of indentation which is not removed.
- If I escape the newlines and used a trailing `\`, then there's no indentation in the final string but as you point out, the legibility is not ideal.

One hybrid option is something like this:
```rust
bail!(
    "rustdoc output format is too old (v{1}, need v{0}) for file \
    {path:?}\n\n\
    \
    note: Using a newer Rust nightly version should help. \
    The latest cargo-semver-checks and Rust nightly should always be \
    compatible with each other.",
    rustdoc_types::FORMAT_VERSION,
    version.format_version,
)
```

Is that preferable?

src/util.rs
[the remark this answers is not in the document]
obi1kenobi: I didn't realize that was a thing, I thought you had meant the trait. Thanks for bearing with me. Fixed!

src/util.rs
obi1kenobi: I tried my hand at mirroring your style, let me know if this is better.

src/util.rs
epage: Personally, I wouldn't indent it.

Getting fancy with escapes means you are artificially making the format look like the real thing which can drift apart.

I would put this as non-blocking to be clear.

src/util.rs
```diff
@@ -1,23 +1,91 @@
...
+            // (2) The rustdoc JSON file has a version number that is too old, and isn't supported.
+            //     The user should upgrade to a newer nightly Rust version and regenerate the file.
+            // (3) The rustdoc JSON file has a version number that is too new, and isn't supported.
+            //     The user should attempt to upgrade to a newer cargo-semver-checks version
+            //     if one is already available, or open a GitHub issue otherwise.
+
+            // The error on this line is case (1).
+            let version = serde_json::from_str::<RustdocFormatVersion>(&s).with_context(|| {
+                format!(
+                    "parsing rustdoc JSON failed, and could not determine rustdoc format version \
```
epage, line 40: nit: Personally, I would leave out "and could not determine rustdoc format version", that is an implementation detail that the user can't act upon.

As hybrid, you could instead make it "Unrecognized rustdoc output" or something

src/util.rs
```diff
@@ -1,23 +1,91 @@
...
+            match version.format_version.cmp(&rustdoc_types::FORMAT_VERSION) {
+                std::cmp::Ordering::Less => {
+                    // The error here is case (2).
+                    bail!(
+                        "rustdoc output format is too old (v{1}, need v{0}) for file \
+                        {2}\n\n\
+                        \
+                        note: Using a newer Rust nightly version should help. \
+                        The latest cargo-semver-checks and Rust nightly should always be \
+                        compatible with each other.",
```
epage, line 55: nit: this feels superfluous

src/util.rs
```diff
@@ -1,23 +1,91 @@
...
+                }
+                std::cmp::Ordering::Greater => {
+                    // The error here is case (3).
+                    bail!(
+                        "rustdoc output format is too new (v{1}, need v{0}) for this \
+                        cargo-semver-checks when parsing {2}\n\n\
+                        \
+                        note: A newer version of cargo-semver-checks is likely available. \
+                        The latest cargo-semver-checks and Rust nightly should always be \
+                        compatible with each other.",
```
epage, line 69: nit: this feels superfluous

src/util.rs
```diff
@@ -1,23 +1,91 @@
...
+                        rustdoc_types::FORMAT_VERSION,
+                        version.format_version,
+                        path.display(),
+                    )
+                }
+                std::cmp::Ordering::Equal => Err(e).with_context(|| {
+                    format!(
+                        "parsing rustdoc failed even though the format was the expected \
+                            version (v{}) for file {}\n\n\
```
epage, line 78: Nit: To put this more in user terms what about `Unexpected parse error for v{} rustdoc file: {}`?

src/util.rs
```diff
@@ -1,23 +1,91 @@
...
+                        rustdoc_types::FORMAT_VERSION,
+                        version.format_version,
+                        path.display(),
+                    )
+                }
+                std::cmp::Ordering::Greater => {
+                    // The error here is case (3).
+                    bail!(
+                        "rustdoc output format is too new (v{1}, need v{0}) for this \
+                        cargo-semver-checks when parsing {2}\n\n\
```
epage, line 65: nit: "for this cargo-semver-checks" is superfluous and sounds a bit strange

epage (approved):

PR #109, opened 2022-08-30
feat(cli): Flags for looking up lints

epage:
Longer term, we should probably have
- Brief descriptions
- Full explanations, independent of the reference

But this is a start

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -37,6 +37,22 @@ fn main() -> anyhow::Result<()> {
             .info(CompileTimeInformation::default())
             .print::<Markdown>();
         std::process::exit(0);
+    } else if let Some(id) = args.explain.as_deref() {
+        let queries = query::SemverQuery::all_queries();
+        let query = queries.get(id).ok_or_else(|| {
+            let ids = queries.keys().cloned().collect::<Vec<_>>();
+            anyhow::format_err!(
+                "Unknown id `{}`, available id's:\n  {}",
+                id,
+                ids.join("\n  ")
+            )
+        })?;
+        println!("{}", query.description);
+        if let Some(link) = &query.reference_link {
+            println!();
+            println!("See also {}", link);
+        }
+        std::process::exit(0);
     }
 
     match args.command {
@@ -157,6 +173,9 @@ enum Cargo {
 struct SemverChecks {
     #[clap(long, global = true)]
     bugreport: bool,
+
+    #[clap(long, global = true)]
+    explain: Option<String>,
 
     #[clap(subcommand)]
     command: Option<SemverChecksCommands>,
--- a/src/main.rs
+++ b/src/main.rs
@@ -37,6 +37,28 @@ fn main() -> anyhow::Result<()> {
             .info(CompileTimeInformation::default())
             .print::<Markdown>();
         std::process::exit(0);
+    } else if args.list {
+        let queries = query::SemverQuery::all_queries();
+        let mut rows = vec![["id", "type"], ["==", "===="]];
+        for query in queries.values() {
+            rows.push([query.id.as_str(), query.required_update.as_str()]);
+        }
+        let mut widths = [0; 2];
+        for row in &rows {
+            widths[0] = widths[0].max(row[0].len());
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -37,6 +37,22 @@ fn main() -> anyhow::Result<()> {
...
+                "Unknown id `{}`, available id's:\n  {}",
+                id,
+                ids.join("\n  ")
+            )
+        })?;
+        println!("{}", query.description);
+        if let Some(link) = &query.reference_link {
+            println!();
+            println!("See also {}", link);
+        }
```
obi1kenobi, line 76: I'm assuming you've thought about showing the implementation link, and have decided it isn't worth it? No strong feelings one way or the other on my end, just wanted to get your thoughts.

I had originally included it in the error message information in case someone felt an error was false-positive, to make it possible to review the query and/or report an issue more easily.

src/main.rs
```diff
@@ -37,6 +37,28 @@ fn main() -> anyhow::Result<()> {
             .info(CompileTimeInformation::default())
             .print::<Markdown>();
         std::process::exit(0);
+    } else if args.list {
+        let queries = query::SemverQuery::all_queries();
+        let mut rows = vec![["id", "type"], ["==", "===="]];
```
obi1kenobi, line 42: The `description` field in the query files is intended to be that kind of short description, though I'm not that skilled at making it nice :)
https://github.com/obi1kenobi/cargo-semver-check/blob/main/src/queries/auto_trait_impl_removed.ron#L4

Feel free to change the descriptions of existing queries if you'd like.

obi1kenobi (commented):
Is it worth adding a check that makes sure unexpected CLI use like `--list --explain foo` gets noticed and reported? Right now, it seems like `--list` dominates almost everything else and causes it to be silently ignored.

src/main.rs
epage: I didn't think about it but I also don't think it should go in here.  We should keep the scope of what is shown to be fairly narrow and as pertinent as possible.

src/main.rs
```diff
@@ -37,6 +37,44 @@ fn main() -> anyhow::Result<()> {
             .info(CompileTimeInformation::default())
             .print::<Markdown>();
         std::process::exit(0);
+    } else if args.list {
```
epage, line 40: > Is it worth adding a check that makes sure unexpected CLI use like --list --explain foo gets noticed and reported? Right now, it seems like --list dominates almost everything else and causes it to be silently ignored.

I can go either way.  I had originally written it treating the flags as if they were overwriting each other (which we can explicitly have clap do on our behalf so intent is clear and to make sure  the last one wins)

src/main.rs
epage: Some aren't so short: https://github.com/obi1kenobi/cargo-semver-check/blob/main/src/queries/struct_repr_transparent_removed.ron

obi1kenobi: Good catch. Would it make sense to make an optional "short description" field that we could include here with `short_description.unwrap_or(long_description)` style logic? Then `repr(transparent)` could get a short description and the actually-short ones would be fine omitting it.

Or would you recommend something else, or just skipping for now and coming back to it later? I have no strong feelings one way or the other.

src/main.rs
obi1kenobi: Since some things are flags and some are arguments (`check-release`), I'm worried about weird cases like `cargo semver-checks check-release --list` which might non-obviously exit 0 through the completely unintended outcome of not actually checking the release.

I'm inclined to say we should probably return errors for unexpected cases rather than doing last-one-wins.

src/main.rs
epage: I was wanting to get this in with what we have today and then iterate from there.

src/main.rs
epage: > `cargo semver-checks check-release --list`

I intentionally supported that with `global = true` as these seemed like universally applicable flags.

For mixing them together though, I've gone ahead and made them exclusive

obi1kenobi (approved):

Issue #126, opened 2022-09-09
More resiliency against rustdoc format breaks


obi1kenobi:
The rustdoc JSON format has moved from v18 to v21 in approximately a week, which is rather more frequent changes than I had hoped for. This harms the user experience.

We should explore some ideas around supporting more format versions. For example:
- split the Trustfall adapter into its own crate
- publish new major versions of that crate for every new JSON format, with each version number matching the format number (v18 supports rustdoc v18, etc.)
- make `cargo-semver-checks` depend on multiple of the most recent adapter versions (e.g. v18, v19, v20, v21)
- execute semver queries using the adapter for the appropriate JSON format version

Time permitting, I'll try to prototype this over the weekend and see how it might go. This might also help with #67.

PR #130, opened 2022-09-12
Prototype for simultaneous support of multiple rustdoc format versions.

obi1kenobi:
Prototype for supporting multiple rustdoc format versions in the same version of cargo-semver-checks (related to #126). Intended for discussion, and not planned to merge as-is.

It supports rustdoc JSON format v18 and v21, and will happily deserialize either file format. However, the baseline and current JSON files must have the same version -- no mixing and matching is allowed. This constraint is not a problem for the GitHub Action or the common ways of using cargo-semver-checks from the terminal.

I'm strongly considering splitting out the version-specific `rustdoc_vX` directories (containing the Trustfall adapters for each format version) into their own crate, which for ease of use could have major version X compatible with rustdoc JSON format version X. Another crate could perhaps include the "choose whichever version makes sense" functionality and provide a `Box<dyn Adapter>` for Trustfall querying. This would probably be useful outside of cargo-semver-checks since this is not the only use case for querying rustdoc JSON.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -199,7 +199,8 @@ dependencies = [
  "ignore",
  "log",
  "ron 0.7.1",
- "rustdoc-types",
+ "rustdoc-types 0.14.0",
+ "rustdoc-types 0.17.0",
  "semver",
  "serde",
  "serde_json",
@@ -980,6 +981,15 @@ name = "rustc-hash"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "08d43f7aa6b08d49f382cde6a7982047c3426db949b1424bc4b7ec9ae12c6ce2"
+
+[[package]]
+name = "rustdoc-types"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "eb8ae6df9a4006b7968781abb9a2ea2983b5f3cad2ca9294d195781618194c28"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "rustdoc-types"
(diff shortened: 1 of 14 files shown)


obi1kenobi:
I broke and/or deleted most of the tests but I tested this by hand with v18 and v21 files, and it works as designed:
- v18 current and v18 baseline works
- v21 current and v21 baseline works
- mix-and-match files produce an error, the expectation being that the files are generated by the same-ish rustdoc version and therefore have the same format number

obi1kenobi:
@epage if you have a second, I'd love your thoughts on whether this is worth pursuing and whether this approach makes sense or if there's something better we should do instead.

epage:
So long as we aren't distributed with rustup, I think this is worth pursuing.

I also think its worth reconsidering using stable rust with RUSTC_BOOTSTRAP as that will produce less churn, not require users to have an extra toolchain, and give us more time to support a rustdoc version.

obi1kenobi:
> I also think its worth reconsidering using stable rust with RUSTC_BOOTSTRAP as that will produce less churn, not require users to have an extra toolchain, and give us more time to support a rustdoc version.

`RUSTC_BOOTSTRAP` is an interesting idea. If I'm understanding correctly, it would allow us to stay on stable with whatever rustdoc JSON format version that stable produces, and we'd have time to prepare for the next rustdoc JSON format version since we'd know which version will be in the next stable?

Are there any downsides we should consider? Because to me, that seems very worth doing — perhaps *in addition* to multi-version capability so we can support more than one stable at a time.

epage:
> Are there any downsides we should consider?

The main thing to remember is its just as unstable as nightly.

From what I understand, the Linux kernel is using the RUSTC_BOOTSTRAP approach.

> Because to me, that seems very worth doing — perhaps in addition to multi-version capability so we can support more than one stable at a time.

Yes, I see both routes being worth while.

obi1kenobi:
The draft served its purpose and this idea was merged as #133.

PR #133, opened 2022-09-28
Support multiple rustdoc versions: v16 (stable), v21 (beta), v22 (nightly).

### LINKED ISSUE #126: More resiliency against rustdoc format breaks

obi1kenobi:
Moved the adapter and indexed crate code to `trustfall-rustdoc-adapter`, whose major versions match the rustdoc format version number they can parse and query.

Added `trustfall_rustdoc`, a general-purpose rustdoc querying crate that can support any combination of versions we choose to add, and where versions can be enabled/disabled using crate features. It uses `trustfall-rustdoc-adapter`. Yes, I made a mess of dashes and underscores again :( No, there's not much that can be done about it now.

Now `cargo-semver-checks` contains only the CLI app + rustdoc generation facilities + semver queries and associated checking logic.

Tested by generating both stable and beta versions of the test rustdocs and running `cargo test`. Merging #131 is a prerequisite for actually testing on stable and beta (because we need `RUSTC_BOOTSTRAP`), so I'll add the test matrix across versions after this and #131 both merge.

Resolves #126.


obi1kenobi:
New nightly is out and uses rustdoc v22. Let's see how small of a change I can make adding that be.

obi1kenobi:
> New nightly is out and uses rustdoc v22. Let's see how small of a change I can make adding that be.

Answer: a dependency-only change, to update to a newer `trustfall_rustdoc` and add the new version's feature.

epage (approved):

PR #139, opened 2022-09-28
Add note on supported Rust versions for rustdoc JSON format.

obi1kenobi:
Also update a few broken links and references in documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,23 +29,36 @@ or other reference pages, as appropriate. It also includes the item name
 and file location that are the cause of the problem, as well as a link
 to the implementation of that query in the current version of the tool.
 
-Notes:
-- **No false positives**: Currently, all queries report constructive proof of semver violations:
-  there are no false positives. They always list a file name and line number for the baseline item
-  that could not be found in the new code.
-- **There are false negatives**: This tool is a work-in-progress, and cannot check all kinds of
-  semver violations yet. Just because it doesn't find any semver issues doesn't mean
-  they don't exist.
-- **Support for most recent stable and beta Rust**: Different Rust compiler versions
-  produce different rustdoc format versions. The most recent `cargo-semver-checks` always
-  supports the most recent Rust stable and beta versions. It *may* also support some nightly
-  versions, without guarantees and on a *best effort* basis.
-- If running on a massive codebase (hundreds of thousands of lines of Rust),
-  the queries may be a bit slow: there is some `O(n^2)` scaling for `n` items in a few places that
-  I haven't had time to optimize down to `O(n)` yet. Apologies! I have temporarily prioritized
-  features over speed, and the runtime will improve significantly with a small amount of extra work.
-
 ## FAQ
+
+### What Rust versions does `cargo-semver-checks` support?
+
+`cargo-semver-checks` uses the rustdoc tool to analyze the crate's API.
+Rustdoc's JSON output format isn't stable, and can have breaking changes in new Rust versions.
+
+When each `cargo-semver-checks` version is released, it will at minimum include support
+for the then-current stable and beta Rust versions. It may, but is not guaranteed to,
+additionally support some nightly Rust versions.
+
+[The GitHub Action](https://github.com/obi1kenobi/cargo-semver-checks-action) uses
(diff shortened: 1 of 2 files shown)


obi1kenobi:
Concise writing isn't my strongest side. Feel free to edit as you see fit, especially around which Rust/rustdoc versions are supported.

README.md
```diff
@@ -30,16 +30,20 @@ and file location that are the cause of the problem, as well as a link
...
-  I haven't had time to optimize down to `O(n)` yet. Apologies! I have temporarily prioritized
-  features over speed, and the runtime will improve significantly with a small amount of extra work.
 - **No false positives**: Currently, all queries report constructive proof of semver violations:
   there are no false positives. They always list a file name and line number for the baseline item
   that could not be found in the new code.
 - **There are false negatives**: This tool is a work-in-progress, and cannot check all kinds of
   semver violations yet. Just because it doesn't find any semver issues doesn't mean
   they don't exist.
+- **Support for most recent stable and beta Rust**: Different Rust compiler versions
+  produce different rustdoc format versions. The most recent `cargo-semver-checks` always
```
aDotInTheVoid, line 40: > Different Rust compiler versions produce different rustdoc format versions.

Is it reasonable to assume that a user will understand (or needs to understand) what a rustdoc json format version is?

README.md
```diff
@@ -30,16 +30,20 @@ and file location that are the cause of the problem, as well as a link
...
+  supports the most recent Rust stable and beta versions. It *may* also support some nightly
```
aDotInTheVoid, line 41: > The most recent `cargo-semver-checks` always supports the most recent Rust stable and beta versions.

I'd more carefully qualify this, as this says that you will maintain the adapter until rustdoc-json is stable. I'd think long and hard before publicly committing to that.

aDotInTheVoid (commented):
I don't think this fully explains what/why the situation is. Here's another draft that's probably too wordy

> `cargo-semver-checks` relies on rustdoc-json, a currently unstable rustdoc feature that gives us programmatic access to a crates API. Until that feature is stabilized, `cargo-semver-checks` cannot guarantee suport for all future versions of rust.
>
> However, we aim to support the *current* stable/beta. This is done by using `RUSTC_BOOTSTRAP`, a hack to allow stable/beta compilers to use nightly features.  Despite this, this doesn't imply support for *future* stable releases, so when you update rustc/rustdoc, `cargo-semver-checks` will also need to be updated
>
> We may also support the latest nightly, but that is harder, as the format changes much more frequently (and with less notice) that faking it with stable/beta
>
> The currently supported versions are listed \[here\]\(LINK\)

README.md
epage: Agreed

README.md
epage: Could you expand on how you are reading that into it because I'm not.

obi1kenobi:
Since there's a fair bit of nuance we're trying to convey here, is it worth taking that bullet point and giving it is own section with a paragraph or two?

README.md
aDotInTheVoid: I think it's worth clarifying to the user that keeping up with stable requires ongoing maintenance from the maintainers of `cargo-semver-checks`. This is in contrast all rust crates, where once they work on stable will work on stable without ongoing support.

The way it's written makes it sound like the MSRV is the latest rust, and that you will drop support for old stables immediately, which is much more common in the ecosystem.

obi1kenobi:
I'm going to split out the changes to `CONTRIBUTING.md` and merge them separately since the current links there are just flat out broken, and the unresolved discussion here isn't related to that file at all.

obi1kenobi:
This PR is imperfect, but I think it's better than the status-quo. For example, it much more clearly states that false-positives are bugs and false-negatives are missing features rather than bugs. Having that section in the FAQ would probably have saved at least one user a bunch of work and confusion: https://github.com/obi1kenobi/cargo-semver-check/issues/148

@aDotInTheVoid and @epage, I'm not sure if my rewrite fully addressed your concerns, though I did my best. If you get a chance to take a look post-merge and you still have concerns, please add comments and I'd be happy to address them in a follow-up PR.

PR #151, opened 2022-10-15
Rename skipped checks to unnecessary checks for clarity.

obi1kenobi:
Users reported some confusion around why checks were being skipped: https://github.com/libp2p/rust-libp2p/pull/2647#issuecomment-1256852275

This PR updates the "skipped" message terminology to use "unnecessary" instead. Here's an example before/after.

Before:
```
Completed [   0.551s] 18 checks; 17 passed, 1 failed, 0 skipped
```

After:
```
Completed [   0.551s] 18 checks; 17 passed, 1 failed, 0 unnecessary
```

### DIFF
--- a/src/check_release.rs
+++ b/src/check_release.rs
@@ -129,7 +129,7 @@ pub(super) fn run_check_release(
             config.shell_status(
                 "Starting",
                 format_args!(
-                    "{} checks, {} skipped (allowed by version change)",
+                    "{} checks, {} unnecessary",
                     queries_to_run.len(),
                     skipped_queries
                 ),
@@ -226,7 +226,7 @@ pub(super) fn run_check_release(
             .shell_print(
                 "Completed",
                 format_args!(
-                    "[{:>8.3}s] {} checks; {} passed, {} failed, {} skipped",
+                    "[{:>8.3}s] {} checks; {} passed, {} failed, {} unnecessary",
                     total_duration.as_secs_f32(),
                     queries_to_run.len(),
                     queries_to_run.len() - queries_with_errors.len(),
@@ -355,7 +355,7 @@ pub(super) fn run_check_release(
             .shell_print(
                 "Completed",
                 format_args!(
-                    "[{:>8.3}s] {} checks; {} passed, {} skipped",
+                    "[{:>8.3}s] {} checks; {} passed, {} unnecessary",
                     total_duration.as_secs_f32(),
                     queries_to_run.len(),
                     queries_to_run.len(),
--- a/src/check_release.rs
+++ b/src/check_release.rs
@@ -129,7 +129,7 @@ pub(super) fn run_check_release(
             config.shell_status(
                 "Starting",
                 format_args!(
-                    "{} checks, {} skipped",
+                    "{} checks, {} skipped (allowed by version change)",
                     queries_to_run.len(),
                     skipped_queries
                 ),


thomaseizinger:
"Skip" to me implies that the check did not run. If a version change allows for a certain to change to be made, doesn't that mean the check "passed"?

thomaseizinger:
An analogy here would be the `assert_ne!` macro. We are testing a negative condition that - if successful - will make the test pass.

obi1kenobi:
> "Skip" to me implies that the check did not run. If a version change allows for a certain to change to be made, doesn't that mean the check "passed"?

The wording situation is tricky here. The check *did not* run: `cargo-semver-checks` saw that anything the check found would have been allowed anyway, so it skipped running it.

> An analogy here would be the `assert_ne!` macro. We are testing a negative condition that - if successful - will make the test pass.

Applying the analogy here: the compiler proved that the equality case (panic branch of the `assert_ne!()`) is unreachable, and completely omitted the assert.

What do you think about replacing `skipped` with `unnecessary`? Example:
```
Completed [   0.551s] 18 checks; 17 passed, 1 failed, 0 unnecessary (allowed by version change)
```

thomaseizinger:
> > "Skip" to me implies that the check did not run. If a version change allows for a certain to change to be made, doesn't that mean the check "passed"?
> 
> The wording situation is tricky here. The check _did not_ run: `cargo-semver-checks` saw that anything the check found would have been allowed anyway, so it skipped running it.

That sounds like an implementation detail to me. Why should I care about the difference between the check running and succeeding vs the optimisation of not running it at all? :)

Would it make sense to reduce the output to:

```
X semver violations found by checking Y rules.
```

As a user I primarily care about the number of violations (want it to be 0). The latter element is useful to know that anything has been checked at all. One could argue that with enough trust in the tool, that might actually be unnecessary too or only something that is shown with debug output.

src/check_release.rs
```diff
@@ -129,7 +129,7 @@ pub(super) fn run_check_release(
             config.shell_status(
                 "Starting",
                 format_args!(
-                    "{} checks, {} skipped",
+                    "{} checks, {} skipped (allowed by version change)",
```
epage, line 132: If we are skipping checks because they are allowed, I feel like "ignored" would be a better word than "skipped".

obi1kenobi: Sounds good. To be clear, are you advocating for just `{} ignored` or the more elaborate `{} ignored (allowed by version change)` option?

obi1kenobi:
> > > "Skip" to me implies that the check did not run. If a version change allows for a certain to change to be made, doesn't that mean the check "passed"?
> > 
> > 
> > The wording situation is tricky here. The check _did not_ run: `cargo-semver-checks` saw that anything the check found would have been allowed anyway, so it skipped running it.
> 
> That sounds like an implementation detail to me. Why should I care about the difference between the check running and succeeding vs the optimisation of not running it at all? :)
> 
> Would it make sense to reduce the output to:
> 
> ```
> X semver violations found by checking Y rules.
> ```

I think that would raise more questions — I could see a user like me asking "Why do different runs do different numbers of checks?"

I'd love it if that info could be inline in the terminal output rather than in docs somewhere. The challenge is obviously how to give maximum info in minimum word count.

> As a user I primarily care about the number of violations (want it to be 0). The latter element is useful to know that anything has been checked at all. One could argue that with enough trust in the tool, that might actually be unnecessary too or only something that is shown with debug output.

I agree. But I don't think linters in general, or `cargo-semver-checks` in particular, have earned that level of trust quite yet. There have been too many bugs in too many tools, and many users (understandably) are still wary.

There's also a familiarity benefit to having similar output to other tools commonly used by the community. `cargo-semver-checks` has its output modeled after `cargo-nextest`: https://nexte.st/

src/check_release.rs
epage: I feel like the parenthetical statement makes it overly long and harder to scan.  Ideally there would be a better way to communicate that

obi1kenobi: A few alternatives:
- `ignored by {} version bump` inserting `major/minor/patch` into the new formatting slot
- `ignored for {} version bump` but otherwise same as above
- `unnecessary` instead of `ignored` in the above two options

Any others you'd suggest?

I agree that ideally we'd have a super-clear one-word way to communicate what's going on. But the current phrasing is unclear and I'm inclined to go for a "better than the status quo" option. We can always polish more in the future if we find a better way to phrase it.

epage: Alternatively, we list the reason alongside the check

No strong opinions on what direction is taken atm though

thomaseizinger:
> > > > "Skip" to me implies that the check did not run. If a version change allows for a certain to change to be made, doesn't that mean the check "passed"?
> > > 
> > > 
> > > The wording situation is tricky here. The check _did not_ run: `cargo-semver-checks` saw that anything the check found would have been allowed anyway, so it skipped running it.
> > 
> > That sounds like an implementation detail to me. Why should I care about the difference between the check running and succeeding vs the optimisation of not running it at all? :)
> > 
> > Would it make sense to reduce the output to:
> > 
> > ```
> > X semver violations found by checking Y rules.
> > ```
> 
> I think that would raise more questions — I could see a user like me asking "Why do different runs do different numbers of checks?"

The idea I was trying to get across is that "checking Y rules" will always say the same number (unless you add more rules). The rule is still there but it didn't report a violation because the semver change makes whatever API change you are making okay.

Does that make sense?

obi1kenobi:
> The idea I was trying to get across is that "checking Y rules" will always say the same number (unless you add more rules). The rule is still there but it didn't report a violation because the semver change makes whatever API change you are making okay.

Ah I see. That's an interesting way to look at it for sure.

I'm not the best at CLI user experience design, so I mostly mirrored `cargo-nextest`'s approach and terminology and otherwise let @epage handle most of it since he's much better at it than I am.

@epage, what do you think about this idea of reporting `X semver violations found by checking Y rules.` and not reporting which/how many rules were ignored?

src/check_release.rs
obi1kenobi: Oh that's great feedback — thanks. How about this: I'll swap in `unnecessary` since it seems like a clear and immediate improvement over the status quo, and we can continue finessing the presentation as we do more user research and gather more feedback on what's confusing.

thomaseizinger: > Oh that's great feedback — thanks. How about this: I'll swap in `unnecessary` since it seems like a clear and immediate improvement over the status quo, and we can continue finessing the presentation as we do more user research and gather more feedback on what's confusing.

Sounds great!

PR #156, opened 2022-10-19
New lint - detecting field removals from a struct variant in enum

### LINKED ISSUE #153: Lint: Removal of a field in a enum struct variant
For example:
```
pub enum Foo {
    Bar {
        a: usize,
        b: u64,
    }
}
```
Removing e.g. the `b` field here is a breaking change.

The fields of struct variants always have "default" visibility, just like variants themselves: they inherit the visibility of the enum itself.

This lint is similar to lints checking for removal of pub fields in structs and for checks over enum variants.

It may require a small amount of new schema and adapter code: currently the [schema](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/main/src/rustdoc_schema.graphql#L249-L261) for `StructVariant` doesn't have a `field` edge like `Struct` [does](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/main/src/rustdoc_schema.graphql#L109), and probably should have one. I believe the type of struct variant fields should still be `StructField`.

tonowak:
Resolves #153.


obi1kenobi:
Will re-run tests after https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/3 is merged and published.

src/queries/enum_structvariant_field_missing.ron
```diff
@@ -0,0 +1,66 @@
+SemverQuery(
+    id: "enum_structvariant_field_missing",
+    human_readable_name: "pub enum structvariant's field removed or renamed" ,
+    description: "An enum variant that is a struct has a field that is no longer available under its prior name.",
```
obi1kenobi, line 4: These descriptions are really hard to write for me too. Here's a suggestion for you to consider -- this seems a bit more readable to me.
```suggestion
    description: "An enum's struct variant has a field that is no longer available under its prior name.",
```

src/queries/enum_structvariant_field_missing.ron
```diff
@@ -0,0 +1,66 @@
...
+    required_update: Major,
+    reference_link: None,
```
obi1kenobi, line 6: Struct variant fields are still items the same way that struct fields are items, so the `item-removed` link that's in a few of the other lints is also appropriate here.

src/queries/enum_structvariant_field_missing.ron
```diff
@@ -0,0 +1,66 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "zero": 0,
+    },
+    error_message: "A publicly-visible enum has at least one variant that has a field that is no longer available under its prior name. It may have been renamed or removed entirely.",
```
obi1kenobi, line 64: Another suggested clarification:
```suggestion
    error_message: "A publicly-visible enum has a struct variant whose field is no longer available under its prior name. It may have been renamed or removed entirely.",
```

obi1kenobi (commented):
A few nitpicks, otherwise this looks great and passes tests with the new schema!

In addition to the inline comments, let's also add an underscore to `enum_structvariant_field_missing` throughout the PR: let's make it `enum_struct_variant_field_missing`.

PR #164, opened 2022-10-30
Load rustdoc JSON using target name, not crate name.

### LINKED ISSUE #162: Crate with custom bin target causes cargo-semver-check to not find the rustdoc JSON output

To reproduce this setup:
```
$ cargo new testing --lib
$ cd testing
$ cargo new test2 --bin
```
To `testing/Cargo.toml`, append:
```toml
[workspace]
members = ["test2"]
```
To `test2/Cargo.toml`, append:
```toml
[[bin]]
name = "test3-actually"
path = "src/main.rs"
```
As far as I know cargo-semver-checks requires the crate to be published to crates.io with no way to disable looking for the crate in the index, so this won't be reproducible with a minimal setup like this; however I have run across this issue in Mica - feel free to test based on [this commit](https://github.com/mica-lang/mica/commit/2eb22f354f5abd521514edc72717baba91ae0e0b).

### Actual Behaviour

What will happen is `cargo semver-checks check-release --workspace` will fail, saying it could not find the expected rustdoc output. I haven't published the minimal setup above for obvious reasons, but on [mica](https://github.com/mica-lang/mica) this can be easily reproduced with the following output:
```
    Updating index
     Parsing mica v0.5.0 (current)
     Parsing mica v0.5.0 (baseline)
     Parsing mica-cli v0.5.0 (current)
Error: Could not find expected rustdoc output for `mica-cli`: /home/daknus/Coding/Rust/mica/target/semver-checks/target/doc/mica_cli.json
```
because `mica-cli` renames its bin target to `mica`.

### Expected Behaviour

cargo-semver-check should not fail and instead find the rustdoc JSON successfully.

### Additional Context

What happens is that the custom named `bin` target causes rustdoc to use that target's name when outputting the JSON file. I have not tested what happens with multiple `bin` targets.

### Debug Output

_No response_

obi1kenobi:
Confirmed to fix #162. Before this, any crates that renamed the default target would fail to check due to a failure to find the generated rustdoc JSON.

The four test crates in `rustdoc_tests` are just default `cargo init --lib` and `cargo init --bin` crates, helping test rustdoc JSON loading for both implicit target names and explicitly renamed target names.


src/manifest.rs
```diff
@@ -1,21 +1,97 @@
...
+    pub(crate) doc: toml_edit::Document,
+}
+
+impl<'a> Manifest<'a> {
+    pub(crate) fn parse(path: &'a std::path::Path) -> anyhow::Result<Self> {
+        let manifest_text = std::fs::read_to_string(path)
+            .map_err(|e| anyhow::format_err!("Failed when reading {}: {}", path.display(), e))?;
+        let doc = manifest_text
+            .parse()
+            .map_err(|e| anyhow::format_err!("Failed to parse {}: {}", path.display(), e))?;
```
epage, line 13: Shouldn't most of this be done for us with `cargo_metadata`?

obi1kenobi: These lines are just moved and extracted into their own method, see above.

src/manifest.rs
```diff
@@ -1,21 +1,97 @@
-pub fn get_package_name(manifest_path: &std::path::Path) -> anyhow::Result<String> {
-    let manifest = std::fs::read_to_string(manifest_path).map_err(|e| {
-        anyhow::format_err!("Failed when reading {}: {}", manifest_path.display(), e)
-    })?;
-    let manifest: toml_edit::Document = manifest
-        .parse()
-        .map_err(|e| anyhow::format_err!("Failed to parse {}: {}", manifest_path.display(), e))?;
-    let package = manifest.get("package").ok_or_else(|| {
```
obi1kenobi, line 8: @epage here's the source for the lines you commented on. I'm not that familiar with the `cargo_metadata` crate and I assumed you had looked into it originally so I didn't dig in further. Happy to take a second look.

src/manifest.rs
obi1kenobi: After looking into it a bit, it seems that `metadata.workspace_packages()` can give us a `Vec<&Package>` and the `Package` type has the info on targets we need.

However, I'm not sure how to pick out the correct `&Package` from that vector based on the `pkg_spec: Option<&str>` function parameter. In general I'm also not familiar enough with cargo's data model (e.g. specs work) so I'm not that comfortable or effective with this code.

If you're able to take a look at it (and you feel it's worth looking into), that would be awesome.

PR #170, opened 2022-11-04
Add function parameters

tonowak:


src/queries/function_parameter_count_changed.ron
```diff
@@ -0,0 +1,50 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+    },
+    error_message: "Parameter count of a publicly-visible function has changed.",
+    per_result_error_template: Some("function {{join \"::\" path}}, previously in file {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 49: Whenever the same item exists in both the baseline and the current version, let's try to report span information for the current version whenever possible — that's a bit more convenient for the user of the tool since looking up the relevant code doesn't require checking out an older version of the code first.

Also, what are your thoughts on mentioning how the parameter count has changed in the per-result error message? Would that be useful for the user to see, or just noise?

src/queries/method_parameter_count_changed.ron
```diff
@@ -0,0 +1,68 @@
...
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "public_or_default": ["public", "default"],
+    },
+    error_message: "Parameter count of a publicly-visible method has changed.",
+    per_result_error_template: Some("method {{join \"::\" path}}, previously in file {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 67: Same questions here as in the other check.

src/test_data/function_parameter_count_changed.output.ron
```diff
@@ -0,0 +1,28 @@
...
+            String("semver_tests"),
+            String("test_cases"),
+            String("parameter_count_changed"),
+            String("function_with_parameters_removed")
+        ]),
+        "span_begin_line": Uint64(8),
+        "span_filename": String("src/test_cases/parameter_count_changed.rs"),
+        "visibility_limit": String("public")
+    }
+]
```
obi1kenobi, line 28:
```suggestion
]

```
It's nice to end files in a newline, since that makes it more likely that there won't be any interoperation issues with other programs (more info here: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline ).

It's very likely that your editor can be configured to do this automatically when you save the file.

src/test_data/method_parameter_count_changed.output.ron
```diff
@@ -0,0 +1,32 @@
...
+            String("semver_tests"),
+            String("test_cases"),
+            String("parameter_count_changed"),
+            String("StructWithMethods"
+        ]),
+        "span_begin_line": Uint64(23),
+        "span_filename": String("src/test_cases/parameter_count_changed.rs"),
+        "visibility_limit": String("public")
+    }
+]
```
obi1kenobi, line 32: another newline here, as above

src/queries/method_parameter_count_changed.ron
```diff
@@ -0,0 +1,68 @@
...
+                        name @filter(op: "=", value: ["%name"])
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                        }
+
+                        # We use "impl" instead of "inherent_impl" here because moving
+                        # an inherently-implemented method to a trait is not necessarily
+                        # a breaking change, so we don't want to report it.
+                        impl {
```
obi1kenobi, line 49: For bonus points and if isn't too much work, consider adding a test case for this to guard against a possible future regression. If we accidentally were to change this `impl` to `inherent_impl`, this could result in a false-positive.

False-positives are particularly dangerous to the `cargo-semver-checks` project as a whole, because they dilute the confidence users have in the tool, and may lead to users to stop using it or not adopt it in the first place.

obi1kenobi (commented):
Lots to like here, and I'm excited to get this merged and published 🎉 Just a few small nitpicks and suggestions for further improvements.

src/queries/method_parameter_count_changed.ron
staniewzki: Actually, the comment was misleading: I copied it from inherent_method_missing query, but now have I realised that's not the case here. If there is no matching function on the current crate, the error won't get reported. Nevertheless, I will add some more test cases.

obi1kenobi: Don't we still want to catch cases where a method got moved from an inherent impl to a trait impl *and also* had its parameter count change? If this is `inherent_impl` instead of `impl` then that case sounds like it wouldn't get reported.

staniewzki: Yes, I was thinking of a scenario, where a method is moved from an inherent impl to a trait impl and the parameter count doesn't change. Then, changing `impl` to `inherent_impl` wouldn't generate an error, as there is no match in the current crate for the baseline version of the method.

obi1kenobi: Right. But leaving `inherent_impl` would also not generate an error because the parameter count wouldn't have changed, right? So it seems like `inherent_impl` is the better option, and the comment is not out of place?

obi1kenobi:
Yeah, that's a good question. Unfortunately here we are limited by how rustdoc structures the data and its terminology. It currently differentiates between functions and methods, and treats associated functions as methods as well, so we must do so as well. But there was some discussion that the distinction between functions and methods may be going away (which I would consider a good thing) and if/when that happens then we'll be able to streamline our checks accordingly.

src/queries/function_parameter_count_changed.ron
staniewzki: I think that in most cases it is just a bloat, as after receiving such an error a manual inspection of the parameters is required anyways. The only scenario where I think that could be helpful is when somebody is new to the programming and the rust language, and could maybe remove the self parameter without realising that it is indeed a parameter, but that is a pretty specific case :)

staniewzki: Okay, now that you put it like that, I do like it. I would go with just printing the baseline vs current parameters count though.

src/test_data/method_parameter_count_changed.output.ron
```diff
@@ -0,0 +1,77 @@
...
+        "old_parameter_count": Uint64(0),
+        "path": List([
+            String("semver_tests"),
+            String("test_cases"),
+            String("parameter_count_changed"),
+            String("StructWithMethods")
+        ]),
+        "span_begin_line": Uint64(26),
+        "span_filename": String("src/test_cases/parameter_count_changed.rs"),
+        "visibility_limit": String("public")
```
obi1kenobi, line 15: These outputs don't seem to have the `current_parameter_count` key-value which the `per_result_error_template` requires. I think you probably forgot an `@output` for it in the query.

Is there a way we could have caught this bug automatically, rather than catching it in code review? In a way, catching bugs in code review is like a trapeze artist falling into the safety net during a performance: it's great that the safety net was there, but the actual objective was to not slip and fall at all.

src/queries/method_parameter_count_changed.ron
```diff
@@ -0,0 +1,68 @@
...
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "public_or_default": ["public", "default"],
+    },
+    error_message: "Parameter count of a publicly-visible method has changed.",
+    per_result_error_template: Some("parameter count of method {{join \"::\" path}} changed from {{old_parameter_count}} to {{current_parameter_count}}, in {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 67: Consider a slight tweak to the wording here for brevity and clarity. It's helpful to look at example printed output when deciding how to phrase user-facing messages like this.

Here's a comparison between the current phrasing and the new phrasing I'd suggest:
```
parameter count of method foo::bar::baz changed from 3 to 2, in foo/bar/baz/mod.rs:1234
foo:bar::baz now takes 2 parameters instead of 3, in foo/bar/baz/mod.rs:1234
```

src/queries/method_parameter_count_changed.ron
```diff
@@ -0,0 +1,68 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "public_or_default": ["public", "default"],
+    },
+    error_message: "Parameter count of a publicly-visible method has changed.",
```
obi1kenobi, line 66: Consider this as an alternative. I'm a bit worried that "parameter count" might be confusing to users.
```suggestion
    error_message: "A publicly-visible method now takes a different number of parameters.",
```

src/queries/function_parameter_count_changed.ron
```diff
@@ -0,0 +1,50 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+    },
+    error_message: "Parameter count of a publicly-visible function has changed.",
+    per_result_error_template: Some("parameter count of function {{join \"::\" path}} changed from {{old_parameter_count}} to {{current_parameter_count}}, in {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 49: Slight wording tweak suggestions for these user-facing messages — see the other query below, they are analogous.

obi1kenobi (commented):
A small bug and a couple of user-facing phrasing suggestions, otherwise this is good to go 🚀

src/test_data/method_parameter_count_changed.output.ron
staniewzki: Wouldn't it be easy to for the error message be tested as well? In the test, after receiving the query results. It should be easy to check if the error message template can be instantiated, and there could also be some precise expected output as it is currently for the queries in the src/test_data

obi1kenobi: That could certainly work — ideally combined with an assertion that the expected output for each query file is non-empty i.e. that an example row exists with which to test instantiating the error message. Testing just the instantiation is probably a good start — I don't think we have any particularly complex templating going on right now for which we might need more precise output testing.

src/queries/method_parameter_count_changed.ron
```diff
@@ -0,0 +1,68 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "public_or_default": ["public", "default"],
+    },
+    error_message: "A publicly-visible method now takes a ddifferent number of parameters.",
```
obi1kenobi, line 66:
```suggestion
    error_message: "A publicly-visible method now takes a different number of parameters.",
```

obi1kenobi (commented):
Just one typo, otherwise it looks great.

obi1kenobi (approved):

PR #171, opened 2022-11-05
New schema for attributes

SmolSir:
`#[repr(...)]` related queries are now using the new `Attribute` schema from the updated adapter (https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/5). This allows to solve problems such as https://github.com/obi1kenobi/cargo-semver-check/issues/104. For now, only the `#[repr(...)]` related queries and their tests have been updated as they were the main reason for this issue.

Queries require the new adapter https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/5 to work correctly.


semver_tests/src/test_cases/enum_repr_int_removed.rs
```diff
@@ -49,3 +49,47 @@ pub enum UsizeEnum {
...
+    Bar,
+    Baz,
+}
+
+#[cfg(feature = "enum_repr_int_removed")]
+#[repr(u8, C)]
+pub enum SeparateU8ToU8CEnum {
+    Bar,
+    Baz,
+}
```
obi1kenobi, line 95:
```suggestion
}

```
It's nice to end files in a newline, since that makes it more likely that there won't be any interoperation issues with other programs (more info here: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline ).

It's very likely that your editor can be configured to do this automatically when you save the file.

semver_tests/src/test_cases/enum_repr_int_removed.rs
```diff
@@ -49,3 +49,47 @@ pub enum UsizeEnum {
     Bar,
     Baz,
 }
+
+#[cfg(not(feature = "enum_repr_int_removed"))]
+#[repr(u8, C)]
+pub enum U8CEnum {
```
obi1kenobi, line 55: Consider adding the analogous test cases as in my suggestion on the other file.

semver_tests/src/test_cases/struct_repr_c_removed.rs
```diff
@@ -1,10 +1,35 @@
...
+#[repr(align(16))]
+pub struct SeparateCAlign16 {
+    pub bar: usize,
+}
+
+#[cfg(feature = "struct_repr_c_removed")]
+#[repr(C, align(16))]
+pub struct SeparateCAlign16 {
+    pub bar: usize,
+}
```
obi1kenobi, line 35:
```suggestion
}

```

semver_tests/src/test_cases/struct_repr_transparent_removed.rs
```diff
@@ -114,6 +114,18 @@ pub struct WithGeneric {
...
 }
 
+#[cfg(not(feature = "struct_repr_transparent_removed"))]
+#[repr(transparent)]
+pub struct BothTransparent {
+    pub bar: usize,
+}
+
+#[cfg(feature = "struct_repr_transparent_removed")]
+#[repr(transparent, )]
```
obi1kenobi, line 124: Oh, fascinating, I didn't realize this was legal! I'm assuming the semantics here are "trailing commas are ignored" and not something more devious?

Would it be worth adding such test cases to the other test files as well?

src/queries/enum_repr_int_changed.ron
```diff
@@ -18,8 +18,14 @@ SemverQuery(
                         name @tag @output
 
                         attribute {
-                            attr_value: value @filter(op: "regex", value: ["$repr_regex"])
-                                                @tag @output(name: "old_attr")
+                            as_string @output(name: "old_attr")
```
obi1kenobi, line 21: (no need for change, just informational) an equivalent way to write this is `old_attr: as_string @output`. I have no preference between the two, whatever works best for you.

src/queries/enum_repr_int_changed.ron
```diff
@@ -18,8 +18,14 @@ SemverQuery(
...
+                            content {
+                                base @filter(op: "=", value: ["$repr"])
+                                argument {
+                                    old_attr_value: base @filter(op: "regex", value: ["$repr_regex"])
+                                                            @tag
```
obi1kenobi, line 26:
```suggestion
                                    old_attr_value: base @filter(op: "regex", value: ["$repr_regex"])
                                                         @tag
```
Just an alignment fix to make the `@tag` align under `@filter`.

src/queries/enum_repr_int_changed.ron
```diff
@@ -43,9 +49,14 @@ SemverQuery(
...
                         attribute @fold @transform(op: "count") @filter(op: ">", value: ["$zero"]) {
-                            value @filter(op: "regex", value: ["$repr_regex"])
-                                    @filter(op: "!=", value: ["%attr_value"])
-                                    @output(name: "new_attr")
+                            as_string @output(name: "new_attr")
+                            content {
+                                base @filter(op: "=", value: ["$repr"])
+                                argument {
+                                    base @filter(op: "regex", value: ["$repr_regex"])
+                                            @filter(op: "!=", value: ["%old_attr_value"])
```
obi1kenobi, line 57:
```suggestion
                                    base @filter(op: "regex", value: ["$repr_regex"])
                                         @filter(op: "!=", value: ["%old_attr_value"])
```
Another alignment fix. It's possible your editor is doing something funny with tabs and spaces, might be worth a look.

Another thing I frequently do and highly recommend: using the GitHub PR review interface to self-review your PR after opening it. The "change of scenery" in how you read the code makes many kinds of bugs pop out in a way that reading the same code in your editor doesn't.

src/queries/enum_repr_int_changed.ron
```diff
@@ -59,7 +70,8 @@ SemverQuery(
     }"#,
     arguments: {
         "public": "public",
-        "repr_regex": "#\\[repr\\([ui]\\d+|size\\)\\]",
+        "repr": "repr",
+        "repr_regex": "[ui]\\d+|size",
```
obi1kenobi, line 74:
```suggestion
        "repr_regex": "[ui](\\d+|size)",
```
I believe that in regex, the alternation (`|`) operator has the lowest precedence. If I'm correct on that then this regex was being interpreted as "either 'size', or 'u/i' followed by digits" which is not correct.

Consider adding test cases for `usize/isize` to verify whether this bug was actually present, and protect against such a regression in the future. Regex is tricky and non-obvious so even if we avoid this bug right now, it could happen again in the future. Better safe than sorry so we should have a test regardless.

src/queries/enum_repr_int_removed.ron
```diff
@@ -52,7 +62,8 @@ SemverQuery(
     }"#,
     arguments: {
         "public": "public",
-        "repr_regex": "#\\[repr\\([ui]\\d+|size\\)\\]",
+        "repr": "repr",
+        "repr_regex": "[ui]\\d+|size",
```
obi1kenobi, line 66: Same regex concern as in the other file, we probably want tests for this as well.

obi1kenobi (commented):
Lots of stuff to like here! This will be a significant improvement to `cargo-semver-checks`'s attribute-checking ability, and I'm excited to get this merged 🎉

I have a few suggestions for additional tests to ward off potential future bugs, and I believe I may have spotted a regex-related bug in a couple of the queries. But it all should be fixable in fairly straightforward fashion — no major concerns.

src/lints/enum_repr_int_changed.ron
```diff
@@ -18,8 +18,14 @@ SemverQuery(
...
 
                         attribute {
-                            attr_value: value @filter(op: "regex", value: ["$repr_regex"])
-                                                @tag @output(name: "old_attr")
+                            as_string @output(name: "old_attr")
+                            content {
+                                base @filter(op: "=", value: ["$repr"])
+                                argument {
+                                    old_attr_value: base @filter(op: "regex", value: ["$repr_regex"])
+                                                            @tag
```
obi1kenobi, line 26:
```suggestion
                                    old_attr_value: base @filter(op: "regex", value: ["$repr_regex"])
                                                         @tag
```
Just fixing alignment.

src/lints/enum_repr_int_changed.ron
```diff
@@ -43,9 +49,14 @@ SemverQuery(
...
                         attribute @fold @transform(op: "count") @filter(op: ">", value: ["$zero"]) {
-                            value @filter(op: "regex", value: ["$repr_regex"])
-                                    @filter(op: "!=", value: ["%attr_value"])
-                                    @output(name: "new_attr")
+                            as_string @output(name: "new_attr")
+                            content {
+                                base @filter(op: "=", value: ["$repr"])
+                                argument {
+                                    base @filter(op: "regex", value: ["$repr_regex"])
+                                            @filter(op: "!=", value: ["%old_attr_value"])
```
obi1kenobi, line 57:
```suggestion
                                    base @filter(op: "regex", value: ["$repr_regex"])
                                         @filter(op: "!=", value: ["%old_attr_value"])
```
Another minor alignment fix. Ideally one day we'll have an auto-formatter for this.

src/lints/enum_repr_int_removed.ron
```diff
@@ -18,7 +18,12 @@ SemverQuery(
                         name @tag @output
 
                         attribute {
-                            value @filter(op: "regex", value: ["$repr_regex"])
```
obi1kenobi, line 21: Can we output the `as_string` value here? It can really help with debugging, even though it isn't in the templated error message, because the new `-vv` "extra verbose" CLI option outputs all the values returned by a semver lint.

src/lints/struct_repr_c_removed.ron
```diff
@@ -16,7 +16,15 @@ SemverQuery(
                     ... on Struct {
                         visibility_limit @filter(op: "=", value: ["$public"]) @output
                         name @tag @output
-                        attrs @filter(op: "contains", value: ["$repr_c"])
+                        
+                        attribute {
```
obi1kenobi, line 20: Also output `as_string` here if possible, same reason as earlier.

test_crates/src/test_cases/struct_repr_c_removed.rs
```diff
@@ -1,10 +1,35 @@
...
+#[repr(align(16))]
+pub struct SeparateCAlign16 {
+    pub bar: usize,
+}
+
+#[cfg(feature = "struct_repr_c_removed")]
+#[repr(C, align(16))]
+pub struct SeparateCAlign16 {
+    pub bar: usize,
+}
```
obi1kenobi, line 35:
```suggestion
}

```
You should be able to configure your editor to always automatically add this terminating newline when it saves.

test_crates/src/test_cases/struct_repr_transparent_removed.rs
```diff
@@ -114,6 +114,18 @@ pub struct WithGeneric {
...
 }
 
+#[cfg(not(feature = "struct_repr_transparent_removed"))]
+#[repr(transparent)]
+pub struct BothTransparent {
+    pub bar: usize,
+}
+
+#[cfg(feature = "struct_repr_transparent_removed")]
+#[repr(transparent, )]
```
obi1kenobi, line 124: Nice! Great idea.

obi1kenobi (commented):
This is great! Just a few tiny improvement suggestions and it's ready to merge as soon as the adapter provides the new schema.

test_crates/src/test_cases/struct_repr_c_removed.rs
SmolSir: It is indeed possible, and very easy to do - [here](https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files). Thanks for suggesting that such features exist!

src/queries/enum_repr_int_changed.ron
SmolSir: That's just me not paying attention to the alignment 😅 I will try doing the self-review thing next time to avoid these small mistakes.

src/queries/enum_repr_int_changed.ron
SmolSir: I changed it as you have shown here - in my opinion it is more clear to have a notation reminiscent to `variable := value`.

test_crates/src/test_cases/struct_repr_transparent_removed.rs
SmolSir: Actually, both `cargo fmt` and `cargo clippy` allow for this and do not enforce a change, what is a bit surprising.

I'm not sure if this could be ignored or if we should do something about it, so I will create an issue about this.

semver_tests/src/test_cases/struct_repr_transparent_removed.rs
SmolSir: This may be a problem, although I'm not sure yet, so I will create an issue about this.

src/queries/enum_repr_int_changed.ron
SmolSir: Actually we already had tests for these cases, but I agree that we add the brackets around `|` operator arguments for easier human-parsing and future-proofing.

src/queries/enum_repr_int_removed.ron
SmolSir: Did so, just like in the other file.

SmolSir:
I added tests according to the above comments and did some small refactoring. Knowing that the recent PRs, especially #203 changed a lot in the project structure / how the tests work, I now have the following questions/ideas:

- Could we separate adding the `as_string` output for some lints into a dedicated issue, as well as expanding it with general lint refactoring / adjusting as the issue's target?
- I don't know if it would be easier to first merge the mentioned PR, then solve the merge conflict and merge this, or to create a separate branch for rewriting the tests to avoid an avalanche of merge conflicts? What works better in your opinion?

I feel guilty for leaving this PR inactive for too long, hopefully we can merge it (or whatever it's fate is) as soon as the new adapter is released!

obi1kenobi:
Happy to save the lint query changes for a future PR, good suggestion. I merged #203 since it was good to go and was huge and difficult to keep reviewing. Sorry to make you have to resolve a complex merge.

SmolSir:
After all, the merge wasn't really that complex - turned out to be more of a refactoring one. As soon as the new `trustfall-rustdoc` release comes out, including the [new schema for attributes](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/5), this PR should instantly pass the checks.

For now, I have 2 side notes / questions:
- in the `enum_repr_int_changed` test crate, there also were cases against it mistakenly reporting a removal of an attribute as a change (and vice-versa for the `enum_repr_int_removed` test crate). Since we now run every lint on all test cases, this resulted not only in the two aforementioned files being almost identical, but also their test outputs. I thus decided to remove the duplicate code, but if the testing strategy was due to change in the future, I can bring the files back to their original form.
- the two files mentioned above also share a portion of cases that just rearrange the attributes rather than changing or removing them. With the new crate test system, these could be separated to form an error-free crate (potentially representing a wise user that gracefully avoided semver violations), potentially introducing a new test scenario (more of these could be added in the future as well). Once again, depending on the future testing strategy, should I bring this idea to life or leave things as-is?

test_crates/enum_repr_int_changed/old/src/lib.rs
```diff
@@ -22,30 +22,85 @@ pub enum IsizeToUsizeEnum {
...
+pub enum U8CToSeparateU16CEnum {
+    Bar,
+    Baz,
+}
+
+
+// The following enums have *rearrangements* of repr(*), potentially
+// splitting singular repr(*) into multiple, smaller repr(*) or merging
+// repr(*) into larger repr(*).
+// They should not be reported by this rule, because they are legal.
```
obi1kenobi, line 47: Can you add this matching comment to the `new` version of this crate?

In general, we want the `old` and `new` crates' source code to be structurally as close as possible to each other, so that one can read them side-by-side and always know which items correspond to each other. Comment blocks like this help break up the long sequences of nothing-but-code, and act as visual anchors when doing that kind of side-by-side reading of the two crates.

test_crates/enum_repr_int_changed/old/src/lib.rs
```diff
@@ -22,30 +22,85 @@ pub enum IsizeToUsizeEnum {
     Baz,
 }
 
-// The following enums have *removals* of repr(i*) and repr(u*),
-// not changes to another repr(i*) or repr(u*).
-// They should not be reported by this rule, because they have their own rule.
+#[repr(u8, C)]
```
obi1kenobi, line 25: Can you throw in test cases for `u64 -> usize` and `i64 -> isize`, just to be safe? `usize/isize` on 64bit x86 are representationally identical to `u64/i64`, but this is still a breaking change (because there's no guarantee that 64bit x86 is the only architecture where a given crate will run) and it would be great to be 100% sure that we catch and report it as such.

obi1kenobi (commented):
A few minor comments, otherwise this is ready to go!

obi1kenobi:
> * in the `enum_repr_int_changed` test crate, there also were cases against it mistakenly reporting a removal of an attribute as a change (and vice-versa for the `enum_repr_int_removed` test crate). Since we now run every lint on all test cases, this resulted not only in the two aforementioned files being almost identical, but also their test outputs. I thus decided to remove the duplicate code, but if the testing strategy was due to change in the future, I can bring the files back to their original form.

I'm generally not that worried about duplicate test cases in different crates: there just isn't much damage that this can do in practice, as far as I can tell. Would you agree?

If so, let's not worry about duplicate test cases. But also if you are _sure_ you didn't accidentally deduplicate code that _looked_ similar but wasn't identical, there's no need to revert, either.

> * the two files mentioned above also share a portion of cases that just rearrange the attributes rather than changing or removing them. With the new crate test system, these could be separated to form an error-free crate (potentially representing a wise user that gracefully avoided semver violations), potentially introducing a new test scenario (more of these could be added in the future as well). Once again, depending on the future testing strategy, should I bring this idea to life or leave things as-is?

Either way is fine. I'm not opposed to the idea, I just don't think these kinds of changes are particularly high-impact: our users won't notice them, and it doesn't seem like it would make us particularly more effective either. So I would personally pick something else.

> * if the testing strategy won't change, should I also look through all of the tests to remove duplicate cases / separate cases that don't break semver into their own crates?

Same as above. Not opposed to it, just doesn't seem to me like it makes that big of an impact so I would advise picking something else.

SmolSir:
What's changed:
- `enum_repr_int_changed` received additional tests for all permutations of `old_attr -> new_attr` with `u64`, `i64`, `usize`, `isize` as one of either attributes. The missing mirror comment from `old` version was also copied to `new` version,
- all of the `_removed` lints affected by this PR (`enum_repr_c_removed`, `enum_repr_int_removed`, `struct_repr_c_removed`, `struct_repr_transparent_removed`) have received an addition of `old_attr: raw_attribute @output` line to help with potential debugging in the future (templated error messages have not been changed). I decided to add this output for all of the lints mentioned above to keep the output somewhat uniform using the new schema for attributes. Hopefully it's ok this way!

test_crates/enum_repr_int_changed/old/src/lib.rs
SmolSir: These are great life-based test cases, I added them!

test_crates/enum_repr_int_changed/old/src/lib.rs
SmolSir: I added the missing comment, indeed it's a lot easier to navigate in the two test files having matching comments in both of them.

obi1kenobi:
Looks great! I'll publish the underlying changes this needs in the adapter, and then merge this right after.

Should be able to release a new cargo-semver-checks tonight 🤞

PR #179, opened 2022-11-12
Removed Cargo.lock

tonowak:
When developing the adapter, I change the `Cargo.toml` to reference a local copy of the adapter, instead of the one from `crates.io`. That changes `Cargo.lock` and it is a tiny tiny bit problematic when `Cargo.lock` from the main branch changes (for example, by changing the adapter version). Then, while pulling, there is a merge conflict that annoys a bit.


epage (changes requested):
Lock files for binaries are important.  See https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

Cargo offers tools to be able to help with this case that I recommend looking into (sorry, don't have time to look  them up at the moment)

tonowak:
I see, thanks for informing me about the lock file.

I've tried searching for the tools you've mentioned, but I can't find them (I've searched mainly in crates.io in section Cargo plugins). @epage Could you maybe hint a bit where I can find the tool, or could you say where I can ask someone about the tool (a zulip channel or something)?

epage:
Sorry, was on mobile and couldn't look it up before.  You want the `[patch]` section of the [config](https://doc.rust-lang.org/cargo/reference/config.html).

Issue #193, opened 2022-11-23
semver-checks check-release detects non-existant changes in core-graphics crate


jrmuizel:
### Steps to reproduce the bug with the above code

Check out https://github.com/servo/core-foundation-rs and run `cargo semver-checks check-release --verbose` in the `core-graphics` directory.

### Actual Behaviour

```
    Updating index
     Parsing core-graphics v0.22.3 (current)
warning: use of deprecated type alias `libc::int32_t`: Use i32 instead.
  --> core-graphics-types/src/base.rs:29:26
   |
29 | pub type CGError = libc::int32_t;
   |                          ^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default

warning: `core-graphics-types` (lib) generated 1 warning
 Documenting core-graphics v0.22.3 (/Users/jrmuizel/src/core-foundation-rs/core-graphics)
    Finished dev [unoptimized + debuginfo] target(s) in 0.66s
     Parsing core-graphics v0.22.3 (baseline)
    Updating crates.io index
 Documenting core-graphics v0.22.3
    Finished dev [unoptimized + debuginfo] target(s) in 0.93s
    Checking core-graphics v0.22.3 -> v0.22.3 (no change)
    Starting 22 checks, 0 unnecessary
        PASS [   0.167s]       major        auto_trait_impl_removed
        PASS [   0.049s]       major        derive_trait_impl_removed
        PASS [   0.001s]       major        enum_marked_non_exhaustive
        PASS [   0.020s]       major        enum_missing
        PASS [   0.008s]       major        enum_repr_c_removed
        PASS [   0.004s]       major        enum_repr_int_changed
        PASS [   0.004s]       major        enum_repr_int_removed
        PASS [   0.001s]       major        enum_struct_variant_field_missing
        PASS [   0.063s]       major        enum_variant_added
        PASS [   0.064s]       major        enum_variant_missing
        PASS [   0.048s]       major        function_missing
        PASS [   0.105s]       major        function_parameter_count_changed
        PASS [   0.156s]       major        inherent_method_missing
        FAIL [   0.004s]       major        method_parameter_count_changed
        PASS [   0.048s]       major        sized_impl_removed
        PASS [   0.001s]       major        struct_marked_non_exhaustive
        PASS [   0.038s]       major        struct_missing
        PASS [   0.004s]       major        struct_pub_field_missing
        PASS [   0.004s]       major        struct_repr_c_removed
[...]

PR #198, opened 2022-11-26
Check that the core-graphics crate does not trigger false-positives.

obi1kenobi:
Adding an integration-style test for `core-graphics`, similar to the one we already have for `rust-libp2p`. Aimed at preventing a regression of #193.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,9 +104,13 @@ jobs:
       - name: test
         run: cargo test
 
-  # Run cargo-semver-checks on a crate with no semver violations,
-  # to make sure there are no false-positives.
-  run-on-crate:
+  run-on-rust-libp2p:
+    # Run cargo-semver-checks on a crate with no semver violations,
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -129,7 +132,6 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
-          components: rustfmt, clippy
```
obi1kenobi, line 132: We didn't actually use rustfmt or clippy here, so we shouldn't install them.

.github/workflows/ci.yml
```diff
@@ -115,8 +120,6 @@ jobs:
         with:
           persist-credentials: false
 
-      # In this particular version, a feature adds an enum variant.
-      # Reference: https://github.com/obi1kenobi/cargo-semver-check/issues/147
```
obi1kenobi, line 119: Just moved all the comments describing what's going on to the same place.

Issue #144, opened 2022-10-03
Convert tests to pairs of Rust projects with their own manifests


obi1kenobi:
Currently, test data is generated from a single project which has a (complicated) way to trigger semver violations by enabling or disabling its features.

In addition to being complicated and non-obvious, this also prevents us from semver-checking manifest information (#48) since there's currently only a single manifest.

It would be an improvement to switch to a test structure where each query is associated with a pair of Rust projects, complete with `Cargo.toml` files and their own source files:
```
/semver_tests/<query_name>/old/
/semver_tests/<query_name>/new/
``` 
Each of these two projects should have a crate name matching the query name, and the same version (e.g. 1.0.0).

To form a test case, one could create the "old" project, then `cp -R old new` and delete / alter code as necessary to cause the desired changes for the test. In this way, the `new` project is pretending to be the "next version" of the `old` project.

Running the tests would involve generating all the rustdoc JSON files as normal, and then asserting that `cargo-semver-checks` finds only the expected semver issues, with no false positives, in each pair of projects.

A crate like `insta` might be helpful here as well, since this is essentially a collection of snapshot tests.

tonowak:
> though I'm on the fence as to whether right now is the time for that.

Well, to solve this issue I need to change a bit the testing environment and it's easier/faster to implement other structural changes at the same time. So it's a good idea to think about it now, because I'll have the full context and the will to make the changes, and I'm not sure whether there'll be a better time.

> I agree that lint files don't have any real dependency on the Rust code in src/, I think that the inverse relationship is definitely there: Rust code in src/ will likely include_str!(...) the lint files. In my mind, that makes the lint files source code as well, so I think it would be good to leave them in src/.

But then, it is possible to move out a file to the `lints/` that does `include_str!` on all lints. Then no changes will be done to `src/`, because it would just include this file. The code is written to run _some_ tests, deliberately without knowing what are those specific tests. Of course, it is not possible to completely remove the relation between the code and the tests, but it is as thin as it can be.

> I agree with the renaming of queries -> lints

What about renaming the `semver_tests`?

> What happens if a lint flags code in more than one test project? How would we represent that?

Good point. But I would expect the lints to be disjoint in the projects they detect. Aren't they?

> then that PR will need to modify the test output files in the existing lint directory. This makes it look like the PR is adding and modifying multiple lints at once, and doesn't make it clear that it's actually only one lint + some new test data (with new test outputs) that are added.

[...]

PR #203, opened 2022-12-02
Split each source code in test_crates/ into pair of crates

tonowak:
Closes https://github.com/obi1kenobi/cargo-semver-checks/issues/144

Unfortunately this PR is a bit tedious to review, sorry.


scripts/regenerate_test_rustdocs.sh
```diff
@@ -3,34 +3,28 @@
...
+	for crate_version in "new" "old"; do
+		crate="$crate_pair/$crate_version"
+		echo "Generating: $crate"
+
+		cd "$TOPLEVEL/test_crates/$crate"
+		RUSTC_BOOTSTRAP=1 $RUSTDOC_CMD -- -Zunstable-options --output-format json
+		mkdir -p "$TARGET_DIR/$crate"
+		mv "$RUSTDOC_OUTPUT_DIR/$crate_pair.json" "$TARGET_DIR/$crate/rustdoc.json"
+	done
+done
```
obi1kenobi, line 28: Can you configure your editor to use spaces instead of tabs here?

scripts/regenerate_test_rustdocs.sh
```diff
@@ -3,34 +3,28 @@
...
-    RUSTC_BOOTSTRAP=1 $RUSTDOC_CMD --features "$feat" -- -Zunstable-options --output-format json
-    mv "$RUSTDOC_OUTPUT" "$TARGET_DIR/$feat.json"
-done <<< "$features"
+# Run rustdoc on test_crates/*/{new,old}/
+for crate_pair in $(ls "$TOPLEVEL/test_crates"); do
+	for crate_version in "new" "old"; do
+		crate="$crate_pair/$crate_version"
+		echo "Generating: $crate"
+
+		cd "$TOPLEVEL/test_crates/$crate"
```
obi1kenobi, line 23: It's a bit more pleasant if the script saves and restores the working directory as it goes. Otherwise e.g. the working directory at the end of the loop (say if we add more code after it) is undefined and unknowable.
```suggestion
		pushd "$TOPLEVEL/test_crates/$crate"
```

scripts/regenerate_test_rustdocs.sh
```diff
@@ -3,34 +3,28 @@
...
+		RUSTC_BOOTSTRAP=1 $RUSTDOC_CMD -- -Zunstable-options --output-format json
+		mkdir -p "$TARGET_DIR/$crate"
+		mv "$RUSTDOC_OUTPUT_DIR/$crate_pair.json" "$TARGET_DIR/$crate/rustdoc.json"
```
obi1kenobi, line 26: To restore the previous working directory, as above.
```suggestion
		mv "$RUSTDOC_OUTPUT_DIR/$crate_pair.json" "$TARGET_DIR/$crate/rustdoc.json"
		popd
```

src/query.rs
```diff
@@ -202,32 +201,26 @@ mod tests {
...
-            .with_context(|| format!("Could not load localdata/test_data/{}.json file, did you forget to run ./scripts/regenerate_test_rustdocs.sh ?", query_name))
-            .expect("failed to load rustdoc under test");
-
-        let baseline = VersionedIndexedCrate::new(&baseline_crate);
-        let current = VersionedIndexedCrate::new(&current_crate);
+    fn get_test_crate_names<'a>() -> Vec<String> {
+        std::fs::read_dir("./test_crates/")
+            .unwrap()
+            .map(|dir_entry| dir_entry.unwrap())
+            .filter(|dir_entry| dir_entry.metadata().unwrap().is_dir())
```
obi1kenobi, line 212: Can you use `.expect()` instead of `.unwrap()` here with a suitable error message, to make it a bit more obvious which unwrapping failed if we get an error from one of them?

src/query.rs
```diff
@@ -238,16 +231,34 @@ mod tests {
...
+        let mut actual_results: Vec<BTreeMap<_, _>> = get_test_crate_names()
+            .into_iter()
+            .map(|crate_pair| {
+                let load_generated_rustdoc = |crate_version: &str| {
+                    let path = format!(
+                        "./localdata/test_data/{}/{}/rustdoc.json",
+                        &crate_pair, crate_version
+                    );
+                    load_pregenerated_rustdoc(&path)
+                };
```
obi1kenobi, line 245: Why not lift this out into a top-level function of its own?

src/query.rs
```diff
@@ -202,32 +203,31 @@ mod tests {
...
-            vec!["test_crates", "CheckPubUseHandling"],
-            vec!["test_crates", "import_handling", "CheckPubUseHandling"],
-            vec![
-                "test_crates",
-                "import_handling",
-                "inner",
-                "CheckPubUseHandling",
-            ],
-        ];
+        let expected_paths = vec![vec!["pub_use_handling", "CheckPubUseHandling"]];
```
obi1kenobi, line 206: This doesn't seem right. Why did some of the results disappear?

This is a high-risk change, because while the refactor you're doing will change the import paths, I wouldn't expect it to cause paths to disappear entirely.

src/query.rs
```diff
@@ -238,16 +238,26 @@ mod tests {
...
-
-        let adapter = VersionedRustdocAdapter::new(&current, Some(&baseline))
-            .expect("could not create adapter");
-        let results_iter = adapter
-            .run_query(&semver_query.query, semver_query.arguments.clone())
-            .unwrap();
-
-        let mut actual_results: Vec<BTreeMap<_, _>> = results_iter
-            .map(|res| res.into_iter().map(|(k, v)| (k.to_string(), v)).collect())
+                .expect("Could not parse expected outputs as ron format.");
```
obi1kenobi, line 241:
```suggestion
                .expect("could not parse expected outputs as ron format");
```

obi1kenobi: Probably best not to change the style of `.expect()` calls, and instead match the existing one.

src/query.rs
```diff
@@ -238,16 +238,26 @@ mod tests {
...
+            .into_iter()
+            .flat_map(|crate_pair| {
+                let crate_new = load_pregenerated_rustdoc(&crate_pair, "new");
+                let crate_old = load_pregenerated_rustdoc(&crate_pair, "old");
+                let indexed_crate_new = VersionedIndexedCrate::new(&crate_new);
+                let indexed_crate_old = VersionedIndexedCrate::new(&crate_old);
+
+                let adapter =
+                    VersionedRustdocAdapter::new(&indexed_crate_new, Some(&indexed_crate_old))
+                        .expect("Could not create adapter.");
```
obi1kenobi, line 253:
```suggestion
                        .expect("could not create adapter");
```
Same as above.

test_crates/enum_repr_int_changed/new/src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+pub enum IsizeToUsizeEnum {
+    Bar,
+    Baz,
+}
+
+// The following enums have *removals* of repr(i*) and repr(u*),
+// not changes to another repr(i*) or repr(u*).
+// They should not be reported by this rule, because they have their own rule.
+
+#[cfg(feature = "enum_repr_int_changed")]
```
obi1kenobi, line 29: This feature isn't defined in the corresponding `Cargo.toml`, perhaps it shouldn't be used here?

test_crates/enum_struct_variant_field_missing/new/src/main.rs
```diff
@@ -0,0 +1,10 @@
+pub enum Enum {
+    FieldWillBeMissing {
+        foo: usize,
+    }
+}
+
+/// This struct variant should not be reported by the `enum_struct_variant_field_missing` rule:
+/// it will be removed altogether, so the correct rule to catch it is
+/// the `enum_variant_missing` rule and not the rule for missing fields.
```
obi1kenobi, line 9: This comment is unclear in this incarnation, since it's referencing code that has already been deleted: there's no `this struct variant` here.

In the original context, it was fine because the struct variant was feature-flagged and present in the source file near this comment. Consider rewording it to be more appropriate to this context.

test_crates/item_missing/new/src/main.rs
```diff
@@ -0,0 +1,30 @@
...
+    pub struct PubUseRemovedStruct;
+
+    pub enum PubUseRemovedEnum {}
+
+    pub fn pub_use_removed_fn() {}
+}
+
+// Moving an inherent method to an implemented trait should not be a breaking change,
+// both when the method is defined inside the trait and when it's implemented externally.
+#[cfg(feature = "inherent_method_missing")]
```
obi1kenobi, line 20: More feature uses that aren't defined in the `Cargo.toml` file.

test_crates/item_missing/new/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "item_missing"
```
obi1kenobi, line 2: `item_missing` tests for multiple different kinds of items going missing. Would it be better to have separate crates for different kinds of items?

obi1kenobi (commented):
I wanted to get *some* comments to you before it's too late in your timezone, and only managed to go through about half of this PR so far.

I'd highly recommend reading your own PR in the PR review interface. There are a bunch of things that stand out there that you probably wouldn't notice in your editor, and I bet you'll catch a number of issues and areas in need of polish and improvement by doing that. This is a high-risk PR since it changes so many things at once, and having more eyes reviewing it will also make it less likely that we end up with bugs as a result.

test_outputs/enum_missing.output.ron
```diff
@@ -2,25 +2,56 @@
...
+        "span_filename": String("src/main.rs"), 
+        "visibility_limit": String("public")
+    }, 
+    {
+        "name": String("UsizeEnum"), 
+        "path": List([String("enum_repr_int_changed"), String("UsizeEnum")]), 
+        "span_begin_line": Uint64(48), 
+        "span_filename": String("src/main.rs"), 
+        "visibility_limit": String("public")
     }
```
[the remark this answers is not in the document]
tonowak: Nevermind, the output was wrongly generated.

test_outputs/inherent_method_missing.output.ron
```diff
@@ -3,28 +3,42 @@
...
+    },
+    {
+        "method_name": String("moved_method"), 
+        "method_visibility": String("public"), 
+        "name": String("Foo"), 
+        "path": List([String("item_missing"), String("Foo")]), 
+        "span_begin_line": Uint64(21), 
+        "span_filename": String("src/main.rs"), 
+        "visibility_limit": String("public")
     }
```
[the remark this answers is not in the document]
tonowak: Nevermind, the output was wrongly generated.

test_outputs/enum_repr_int_removed.output.ron
```diff
@@ -38,5 +38,33 @@
...
+        "span_begin_line": Uint64(39), 
+        "span_filename": String("src/main.rs"), 
+        "visibility_limit": String("public")
+    }, 
+    {
+        "name": String("UsizeEnum"), 
+        "path": List([String("enum_repr_int_changed"), String("UsizeEnum")]), 
+        "span_begin_line": Uint64(44), 
+        "span_filename": String("src/main.rs"), 
+        "visibility_limit": String("public")
```
tonowak, line 68: This doesn't make any sense whatsoever. The `span_begin_line` in those outputs don't exist in the `new/src/main.rs` (the file has only 19 lines). Do you have any idea what happened here @obi1kenobi ?

test_outputs/method_parameter_count_changed.output.ron
```diff
@@ -4,89 +4,79 @@
         "method_name": String("associated_function_with_a_parameter_added"),
```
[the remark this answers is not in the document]
tonowak: (we've moved the discussion to Zulip)

test_outputs/enum_repr_int_removed.output.ron
obi1kenobi: I'm guessing the span begin line numbers are the old numbers from before you refactored the test structure. Note that this test is currently failing due to a mismatch between the actual and expected test data. You may need to just update the line numbers to match the new file (which doesn't have feature-flagged code so of course the line numbers don't match up).

test_outputs/struct_missing.output.ron
```diff
@@ -3,26 +3,46 @@
...
+        "visibility_limit": String("public")
+    },
+    {
+        "name": String("NonExhaustiveUnitStructToPlain"), 
+        "path": List([String("unit_struct_changed_kind"), String("NonExhaustiveUnitStructToPlain")]), 
+        "span_begin_line": Uint64(4), 
+        "span_filename": String("src/main.rs"), 
+        "struct_type": String("unit"), 
+        "visibility_limit": String("public")
+    },
```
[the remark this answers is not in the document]
obi1kenobi: What do you see here that doesn't seem right? I'm missing context.

test_outputs/method_parameter_count_changed.output.ron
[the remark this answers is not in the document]
obi1kenobi: Summary from Zulip for folks following along here: still triaging, unclear what is happening. Assertion inside Trustfall is getting triggered.

test_outputs/enum_repr_int_removed.output.ron
tonowak: Still, something weird happened here. This test output I've generated just now by running the script and then `cargo test`. I'm sure that this test output is not a mistake, because it has been generated just like the other two test outputs that have been failing before the latest commit. And the CI have just been completed for this commit and it is now passing.

test_outputs/struct_missing.output.ron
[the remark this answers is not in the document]
tonowak: It's more about me not being sure that this is the correct output. From what I remember, those structs changed in a particular way, but they haven't been removed.

src/query.rs
```diff
@@ -240,14 +240,24 @@ mod tests {
...
-            .expect("could not create adapter");
-        let results_iter = adapter
-            .run_query(&semver_query.query, semver_query.arguments.clone())
-            .unwrap();
-
-        let mut actual_results: Vec<BTreeMap<_, _>> = results_iter
-            .map(|res| res.into_iter().map(|(k, v)| (k.to_string(), v)).collect())
+        let mut actual_results: Vec<BTreeMap<_, _>> = get_test_crate_names()
+            .into_iter()
+            .flat_map(|crate_pair| {
```
tonowak, line 245: A better idea might be to not flatten it, but leave the information about which crate those outputs are.

test_outputs/method_parameter_count_changed.output.ron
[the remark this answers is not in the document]
obi1kenobi: Resolved in `trustfall_core v0.1.1`.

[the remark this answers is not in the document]
tonowak: Should I create a PR in `trustfall-rustdoc` and a commit here to use the new version, or is it something that you do?

obi1kenobi:
Merged `main` into this branch to pull in the update to `trustfall v0.1.1` which will resolve at least one of the issues this PR ran into.

tonowak:
The `CONTRIBUTING.md` needs to be updated.

test_crates/README.md
```diff
@@ -0,0 +1,10 @@
+# How the testing crate pairs work
+
+In the tests, each lint is ran on _all_ crate pairs from this directory.
+That means that the output of a lint (saved in the `../test_outputs` directory)
+is the concatenation of the outputs on each crate pair.
+
+The output of a lint on a crate pair (a pair is a `new` and `old` crate) 
+is the raw output of a query, assuming that the current crate is the `new` one,
+and the baseline crate is the `old` one.
+
```
tonowak, line 10: It might be good to add that the `template` crate pair is meant to be copy-pasted.

tonowak:
> The CONTRIBUTING.md needs to be updated (after merging https://github.com/obi1kenobi/cargo-semver-checks/pull/216 )

Maybe it's better to move `test_crates/README.md` to `CONTRIBUTING.md` @obi1kenobi ?

tonowak:
It might be beneficial to add the witness crates in a separate PR.

test_outputs/enum_repr_int_removed.output.ron
tonowak: Ah, now that the code uses `map` instead of `flatten`, I immediately see that they are from another crate, not from this same crate. The output seems to be ok.

test_crates/README.md
tonowak: Done in ae210f22ed7c4ccd893d6e6c5f966d4cff9611cd.

src/query.rs
```diff
@@ -236,38 +236,73 @@ mod tests {
...
+        if expected_results != actual_results {
+            let results_to_string = |name, results| {
+                format!(
+                    "{}:\n{}\n",
+                    name,
+                    ron::ser::to_string_pretty(&results, ron::ser::PrettyConfig::default())
+                        .unwrap()
+                )
+            };
+            panic!("Query {} produced incorrect output (./src/lints/{}).\n{}\n{}\nNote that the individual outputs might have been deliberately reordered.", &query_name, &query_name,
```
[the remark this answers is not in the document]
tonowak: Done in 1ea9a6421bbe0c125919fbd0e8c3ea1948a87fb7.

test_crates/pub_use_handling/old/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub struct CheckPubUseHandling;
```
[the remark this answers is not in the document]
tonowak: @obi1kenobi

test_crates/struct_missing/old/new/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "template"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
```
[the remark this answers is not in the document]
tonowak: Done in 1ea9a6421bbe0c125919fbd0e8c3ea1948a87fb7.

obi1kenobi:
> > The CONTRIBUTING.md needs to be updated (after merging #216 )
> 
> Maybe it's better to move `test_crates/README.md` to `CONTRIBUTING.md` @obi1kenobi ?

I'd say move the contributing-specific information to `CONTRIBUTING.md` but also have a `README.md` because that's what GitHub shows when you open that directory, and because not all the info that belongs in that file is contributing-related.

What do you think?

test_outputs/struct_marked_non_exhaustive.output.ron
```diff
@@ -1,41 +1,59 @@
...
+            "name": String("NonExternallyConstructibleStruct"),
+            "path": List([
+                String("non_exhaustive"),
+                String("NonExternallyConstructibleStruct"),
+            ]),
+            "span_begin_line": Uint64(28),
+            "span_filename": String("src/lib.rs"),
+            "struct_type": String("plain"),
+            "visibility_limit": String("public"),
+        },
```
obi1kenobi, line 57: These two results are unexpected. Adding `#[non_exhaustive]` on structs that were already not externally constructible is not breaking. I'm not sure why they are being incorrectly flagged now.

test_outputs/method_parameter_count_changed.output.ron
```diff
@@ -1,92 +1,148 @@
...
+            "old_parameter_count": Uint64(1),
+            "path": List([
+                String("inherent_method_missing"),
+                String("Foo"),
+            ]),
+            "span_begin_line": Uint64(9),
+            "span_filename": String("src/lib.rs"),
+            "visibility_limit": String("public"),
+        },
+    ],
```
obi1kenobi, line 35: Interesting! These are false-positive on account of being duplicate lints: these methods were removed, but they are being reported as having changed parameter counts as well.

To move this PR along, let's just file an issue for this and resolve it in a future PR.

test_outputs/struct_missing.output.ron
```diff
@@ -1,28 +1,63 @@
...
+            "path": List([
+                String("unit_struct_changed_kind"),
+                String("NonExhaustiveUnitStructToPlain"),
+            ]),
+            "span_begin_line": Uint64(4),
+            "span_filename": String("src/lib.rs"),
+            "struct_type": String("unit"),
+            "visibility_limit": String("public"),
+        },
+    ],
```
obi1kenobi, line 62: These are also false-positive: changing the kind of struct (from plain to tuple or unit) is not deleting the struct.

Also best to open a new false-positive issue and sort out in a future PR rather than trying to resolve here.

obi1kenobi (commented):
Overall, it's great: it has already helped flag multiple false-positive issues, and the test crate code is much more readable on account of not having to mess with features and conditional compilation.

I flagged a few minor things that should be easy to fix, and we need to remember to open a few GitHub issues so we don't forget to resolve the lint issues we've flagged. If you get a chance to open them — great! If not, I'll open them when it's not 1am over here 😴

scripts/regenerate_test_rustdocs.sh
```diff
@@ -3,34 +3,32 @@
...
+        crate="$crate_pair/$crate_version"
+        echo "Generating: $crate"
+
+        pushd "$TOPLEVEL/test_crates/$crate"
+        RUSTC_BOOTSTRAP=1 $RUSTDOC_CMD -- -Zunstable-options --output-format json
+        mkdir -p "$TARGET_DIR/$crate"
+        mv "$RUSTDOC_OUTPUT_DIR/$crate_pair.json" "$TARGET_DIR/$crate/rustdoc.json"
+        popd
+    done
+done
```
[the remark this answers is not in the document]
tonowak: Whaat, I was convinced that I already had done this.

scripts/regenerate_test_rustdocs.sh
```diff
@@ -3,34 +3,32 @@
...
-    jq --exit-status -r '.packages[] | select(.name = "test_crates") | .features | keys[]')"
-while IFS= read -r feat; do
-    echo "Generating: $feat"
-    RUSTC_BOOTSTRAP=1 $RUSTDOC_CMD --features "$feat" -- -Zunstable-options --output-format json
-    mv "$RUSTDOC_OUTPUT" "$TARGET_DIR/$feat.json"
-done <<< "$features"
+# Run rustdoc on test_crates/*/{new,old}/
+for crate_pair in $(find "$TOPLEVEL/test_crates/" -maxdepth 1 -mindepth 1 -type d); do
+	# Removing path prefix, leaving only the directory name without forward slashes
+	crate_pair=${crate_pair#"$TOPLEVEL/test_crates/"}
```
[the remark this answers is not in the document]
tonowak: Why shouldn't this be quoted? Maybe the `crate_pair` would have spaces in it, for whatever reasons. Though I hope that the quotation marks match with each other correctly in the old version.

src/query.rs
```diff
@@ -211,38 +211,77 @@ mod tests {
...
-                elem["span_begin_line"].as_usize().unwrap(),
-            )
+        let sort_individual_outputs = |results: &mut BTreeMap<String, Vec<_>>| {
+            let key_func = |elem: &BTreeMap<String, FieldValue>| {
+                (
+                    elem["span_filename"].as_str().unwrap().to_owned(),
+                    elem["span_begin_line"].as_usize().unwrap(),
+                )
+            };
+            for (_key, value) in results.iter_mut() {
```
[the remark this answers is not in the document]
tonowak: Thanks! I was trying to search for something like that in the docs.

src/query.rs
```diff
@@ -211,38 +211,77 @@ mod tests {
...
+                        .unwrap()
+                )
+            };
+            let messages = vec![
+                format!("Query {} produced incorrect output (./src/lints/{}.ron).", &query_name, &query_name),
+                results_to_string(format!("Expected output (./test_outputs/{}.output.ron)", &query_name), &expected_results),
+                results_to_string("Actual output".to_string(), &actual_results),
+                "Note that the individual outputs might have been deliberately reordered. Also, remember about running ./scripts/regenerate_test_rustdocs.sh when needed.".to_string(),
+            ];
+            panic!("\n{}\n", messages.join("\n\n"));
```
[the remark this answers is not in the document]
tonowak: My reasoning is that this whole message is huge and without the endlines it wraps with other lines from the `panic!` macro. With endlines it is more clear where the custom message starts and ends. Though I'm not sure whether it is a good solution, do you think I should remove the endlines? (to decide, it might be good to deliberately launch this `panic!` to see the output in the terminal)

scripts/regenerate_test_rustdocs.sh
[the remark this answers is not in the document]
tonowak: Done in 1e553605724f717ff9ee475cd45f762e15275fb1.

src/query.rs
```diff
@@ -177,32 +178,31 @@ mod tests {
...
+                dir_entry
+                    .metadata()
+                    .expect("failed to retrieve test_crates/* metadata")
+                    .is_dir()
+            })
+            .map(|dir_entry| {
+                String::from(
+                    String::from(dir_entry.path().to_str().unwrap())
+                        .strip_prefix("./test_crates/")
+                        .unwrap(),
```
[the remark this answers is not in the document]
tonowak: Fixed in d2fb843c1d8fa4571aa89398ec3b5cfb3f730d03.

tonowak:
> I'd say move the contributing-specific information to CONTRIBUTING.md but also have a README.md because that's what GitHub shows when you open that directory, and because not all the info that belongs in that file is contributing-related.

Done.

> I flagged a few minor things that should be easy to fix, and we need to remember to open a few GitHub issues so we don't forget to resolve the lint issues we've flagged. If you get a chance to open them — great! If not, I'll open them when it's not 1am over here sleeping

I'll create the issues after merging this PR, to reference the exact lines of codes in the `main` branch.
I plan to create the following issues:
- witness crates,
- `method_parameter_count_changed` lint false-positive,
- `struct_missing` lint false-posivite,
- `struct_marked_non_exhaustive` lint false-positive.

Is there anything I've missed?

scripts/regenerate_test_rustdocs.sh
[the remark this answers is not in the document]
obi1kenobi: I think it should be quoted -- my suggested edit just wraps the whole thing in one extra set of quotes. That's just something I do obsessively in bash scripts since I've debugged several too many scripts where missing quotes either broke something (less bad option) or enabled script injection (much worse option). I just was wondering if adding the extra quotes might break something that I didn't notice.

src/query.rs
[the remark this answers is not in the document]
obi1kenobi: We can leave it as-is for now and focus on getting this merged. We can always change it later if we decide to.

obi1kenobi (approved):

scripts/regenerate_test_rustdocs.sh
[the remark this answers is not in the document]
tonowak: Oh sorry, I've misunderstood. I thought you wanted to delete the quotes. I'll change it in a separate PR.

test_outputs/struct_missing.output.ron
obi1kenobi: Resolved in #218.

test_outputs/method_parameter_count_changed.output.ron
obi1kenobi: Fixed in #220.

obi1kenobi:
Another idea I had for automating false-positive detection in the future:

Some crates intentionally include items that are crafted as "bait" for false-positives, e.g. to act as regression tests for past false-positive issues. If we always place such "bait" in a dedicated file / module (e.g. `false_positive.rs`), then any time a rule reports an item from its own crate's `false_positive` module, we know that's a bug.

Other crates are allowed to report on `false_positive` modules' items, since those items may trigger other lints. For example, the false-positive where deleted methods get reported as changing their number of parameters: if that test case existed in `false_positive.rs` test crates for the lint for changing parameter count, it would have correctly been reported as a missing method by the lint that looks for those instead.

test_outputs/struct_marked_non_exhaustive.output.ron
obi1kenobi: Resolved in #222.

obi1kenobi:
> I'll create the issues after merging this PR, to reference the exact lines of codes in the `main` branch. I plan to create the following issues:
> 
> * witness crates,
> * `method_parameter_count_changed` lint false-positive,
> * `struct_missing` lint false-posivite,
> * `struct_marked_non_exhaustive` lint false-positive.

I hit a burst of productivity today, so I debugged and fixed all three false-positives, and opened the witness crates issue 🚀

PR #204, opened 2022-12-03
One macro to include lints

tonowak:
This PR removes a duplicated list of lints in the source code of the project.

Previously, there was a list of paths to lints in `src/query.rs:SemverQuery:all_queries()` and a list of lint names in `src/query.rs:tests:query_execution_tests!()`. This PR replaces the `query_execution_tests` macro with a macro `add_lints` to generate the tests (just like the previous macro) and to generate a function that returns the list of queries' contents.

The reasoning behind doing this change is that it will be a bit easier to add a new lint (and less room for a mistake). The ideal solution (in my opinion) would be to have a file listing all the lint names in the `src/lints/` directory and then have a macro that reads the file and generates the tests etc, but I don't know how to do it with my limited Rust knowledge.

@obi1kenobi please comment whether you like the idea behind this PR and whether you have any ideas about how to do the above mentioned macro.


src/query.rs
```diff
@@ -280,40 +255,52 @@ mod tests {
...
     }
+}
+
+macro_rules! add_lints {
+    ($($name:ident,)*) => {
+        $(
+            #[test]
+            fn $name() {
+                tests::check_query_execution(stringify!($name))
+            }
```
obi1kenobi, line 266: Doesn't this put this test function outside of `mod tests` and into the regular module that always gets compiled, even in non-test builds?

I like the general idea of having one place to add lints, but I don't think we want to add test-only code that needs to be compiled in non-test builds too.

tonowak: Oh ok, I didn't know what the `#[cfg(test)]` does.

But then, I don't know how to write this code in the `#[cfg(test)] mod tests` module. Rust doesn't allow redefining/extending a `mod`.

obi1kenobi: I'm not sure either. It might be worth looking at other bigger and more established crates (e.g. clippy) to see if they have any tricks that might help here as well.

src/query.rs
```diff
@@ -215,7 +190,7 @@ mod tests {
         assert_eq!(expected_paths, actual_paths);
     }
 
-    fn check_query_execution(query_name: &str) {
+    pub fn check_query_execution(query_name: &str) {
```
obi1kenobi, line 193: Can this be `pub(in query)` or something narrower like that, rather than `pub`?

obi1kenobi (commented):
Looks good! One question for you to consider, and otherwise I think this is fine to merge.

Oh, it might be a good idea to update `CONTRIBUTING.md` as well, since I think some of its instructions will be outdated after this PR merges.

src/query.rs
tonowak: Oh, good idea. I didn't think it's possible to narrow the scope like that in Rust. I'll change it. I think it's not possible to narrow it further, because I'm calling this function from another (sibling) module, so it's not possible to narrow it to a module and I guess the next wider scope is per-file. But idk, I still have to read about the possible scope narrowing in Rust.

obi1kenobi: I think that function is in `mod query -> mod tests` and is used from `mod query` and from `mod query -> mod tests_lints`. So if my mental model of this is correct, narrowing to `pub(in query)` should work since all child modules of `query` should be able to access it too.

I don't think there's real "per file" scoping in Rust, just per module with files as implicit modules.

tonowak:
> Oh, it might be a good idea to update CONTRIBUTING.md as well, since I think some of its instructions will be outdated after this PR merges.

Ups, I didn't know about the existence of this file. My recent PR already made some changes that made this markdown file not up-to-date. I'll create a PR to solve that.

src/query.rs
obi1kenobi: Perfect!

obi1kenobi:
If you can resolve the merge conflict, this is good to go 🚀

tonowak:
Here you go: https://github.com/obi1kenobi/cargo-semver-checks/pull/204/commits/345ea946d1ba9630b63b568d308f964f5dc62763
I tried to make the changes so that a merge conflict wouldn't happen, but I failed, sorry.

CONTRIBUTING.md
```diff
@@ -129,12 +129,12 @@ Checklist:
   specifically reports the removal of the struct rather than all its fields separately.
 - Add the outputs you expect your query to produce over your test case in
   a new file: `test_outputs/<query_name>.output.run`.
-- Add `<query_name>` to the list of queries tested by the `query_execution_tests!()`
-  macro near the bottom of `src/query.rs`.
+- Add `<query_name>` to the list of queries used by the `add_lints!()`
+  macro near the bottom of `src/query.rs`. It includes the query content and also creates
+  a new test named `<query_name>` that compares the output of running this query on `test_crates/`
```
obi1kenobi, line 134: Add the word "function" ("... creates a new test function named ...") and reflow text to keep lines short.
```suggestion
- Add `<query_name>` to the list of queries used by the `add_lints!()` macro near the bottom
  of `src/query.rs`. It includes the query content and also creates a new test function
  named `<query_name>` that compares the output of running this query on `test_crates/`
```

PR #205, opened 2022-12-03
Move rustdoc_tests to test_rustdoc

tonowak:
In one of my previous PRs I've missed the existence of the `rustdoc_tests` directory, in which I've introduced the `test_*` directory pattern and this PR just renames the `rustdoc_tests` to this pattern.

Honestly, I'm not sure whether it is a good idea - currently `test_crates` and `test_outputs` are closely linked with one another, whereas `rustdoc_tests` doesn't have anything to do with them.

I'm hoping you'll comment your opinion @obi1kenobi


obi1kenobi (approved):

obi1kenobi:
I don't have strong feelings one way or the other. If you think this is worth doing, go for it. I think you should have the ability to merge now, is that right? If so, feel free to go ahead.

PR #206, opened 2022-12-03
Run lints on nonchanged crates

tonowak:
This PR is a follow-up to https://github.com/obi1kenobi/cargo-semver-checks/pull/203. In the tests, it runs each lint on all testing crates, comparing the difference of a crate with the same crate (and it expects that there is no difference). Thus, it is an additional check to be sure that there are no false-positives.

I'm not sure how GitHub behaves when I've created a PR that is built on another, not yet merged PR. I hope that what I did makes sense, but if not, I can close the PR and reopen it after the previous PR will get merged.


tonowak:
This `git merge` produced quite a lot of commits in the Conversation page here. Was there a way to avoid it?

obi1kenobi:
> This `git merge` produced quite a lot of commits in the Conversation page here. Was there a way to avoid it?

Not that I know of. Also there are now a lot of merge conflicts.

The workflow of having PRs that depend on other unmerged PRs is called "stacked commits" and it's worth reading about, if you're interested in learning how to perfect it.

tonowak:
One possible issue: probably `cargo test` is now even slower.

src/query.rs
```diff
@@ -233,18 +233,40 @@ mod tests {
...
+                    crate_pair
+                );
+                assert!(
+                    run_query_on(&indexed_crate_old, &indexed_crate_old)
+                        .1
+                        .is_empty(),
+                    "{}",
+                    crate_pair
+                );
+                run_query_on(&indexed_crate_new, &indexed_crate_old)
```
obi1kenobi, line 269: Would you mind adding some comments about what's going on here and why? Otherwise this code seems pretty unusual and confusing...

tonowak: Isn't it self-explanatory? In short, I'm asserting that running the crate compared to the same crate returns no errors.

tonowak: What do you think about the new version? @obi1kenobi

src/query.rs
```diff
@@ -252,28 +281,32 @@ mod tests {
...
+                     crate_version: &str| {
+                         let output_pair = run_query_on(crate_, crate_);
+                         if !output_pair.1.is_empty() {
+                             let output_difference = pretty_format_output_difference(
+                                 query_name,
+                                 "Expected output (empty output)".to_string(),
+                                 BTreeMap::new(),
+                                 format!("Actual output ({}/{})", crate_pair, crate_version),
+                                 BTreeMap::from([output_pair]));
+                             panic!("Running a query on a crate that didn't change should always produce an empty output.\n{}\n",
```
obi1kenobi, line 295: This message states an invariant but doesn't mention how the invariant was broken, which is confusing. Did the crate change unexpectedly? Did the query fail to run and instead return an error? Or was the output non-empty?

Also, the invariant here can be stated more strongly: it's not just that the crate didn't change, but the rustdoc JSON input was identical between the baseline and current.

src/query.rs
```diff
@@ -252,28 +281,32 @@ mod tests {
...
-                assert!(
-                    run_query_on(&indexed_crate_old, &indexed_crate_old)
-                        .1
-                        .is_empty(),
-                    "{}",
-                    crate_pair
-                );
+                let assert_no_false_positives_in_nonchanged_crate =
+                    |crate_: &VersionedIndexedCrate,
+                     crate_version: &str| {
```
obi1kenobi, line 286: This code is getting quite difficult to read. This is already a second closure inside a closure inside the test function. Try to move some of this functionality out into functions of their own?

src/query.rs
```diff
@@ -252,28 +281,32 @@ mod tests {
...
+                         let output_pair = run_query_on(crate_, crate_);
+                         if !output_pair.1.is_empty() {
```
obi1kenobi, line 288: nit: `output_pair.1` is opaque. You already have a `let` binding, and destructuring the tuple gives you the opportunity to give `output_pair.1` a name that will guide the reader's intuition.

Code is written ~once but read many, many times, by many different people, most of whom didn't write or review it originally. Optimize for readability, not for being easy to write.

src/query.rs
```diff
@@ -252,28 +281,32 @@ mod tests {
...
+                             let output_difference = pretty_format_output_difference(
+                                 query_name,
+                                 "Expected output (empty output)".to_string(),
+                                 BTreeMap::new(),
+                                 format!("Actual output ({}/{})", crate_pair, crate_version),
```
obi1kenobi, line 293: nit: I think the next `clippy` version is going to get lints that encourage inline use of the formatting variables whenever possible, like so:
```suggestion
                                 format!("Actual output ({crate_pair}/{crate_version})"),
```
I already had to fix a bunch of them earlier: https://github.com/obi1kenobi/cargo-semver-checks/pull/219/files#diff-f0c63cae3ed0328ebbe14ac22d86b43bc8a40146944b03c8326681e4f3802443L217-R221

obi1kenobi (commented):
This PR can still benefit from comments in the code. Yes, the `assert_no_false_positives_in_nonchanged_crate` function name is pretty good, but adding a comment that goes one step further and explains why anything found in an unchanged crate is a false-positive (and why we're even bothering to test this) won't hurt anyone. For example, it would be perfectly understandable if a new person reading the code assumed that testing a crate against itself is *guaranteed by construction* to return no lints — that getting lints in that situation would have caused a type error first. They have no way of knowing this is not true, and also no way of knowing that this sort of check has failed on real-world crates in the past because of obscure reasons. I certainly could not have imagined the ways in which this invariant failed to hold in the real world.

Here's the high-level view: we want `cargo-semver-checks` to be a successful project with a long life and a community of maintainers. People have limited time and energy to contribute, and many projects they could be contributing to. Why would they contribute to a project whose code is confusing, if they could contribute to a project that is easy to understand and has a welcoming codebase with code comments explaining possible sticky situations?

Thus far, the emphasis was on making `cargo-semver-checks` minimally viable as a tool: useful functionality that solves a real-world problem, users that like it, etc. We've succeeded, and it's already more than minimally viable: it runs in the repos of many established projects, and has prevented real-life semver violations. We now have to also start thinking about clean code that contributors can dive into, that lets us build and ship new features quickly and without bugs, etc.

src/query.rs
tonowak: Fixed in c2222fffe7f3daa27fb468c6f542b1ce1665fac1

src/query.rs
tonowak: Fixed in a22b9b5a1123262abdc190c3dc561ce2365e8084

src/query.rs
tonowak: Fixed in 9545e230ccfdd7260810d25f8999059d5fcbbe1b and 6898e90b3cef46952492dbd1882b71137bb68054

src/query.rs
tonowak: Fixed in b58de954ac788d4a4eab30f13eba324cb670d4ee

src/query.rs
tonowak: Fixed in b58de954ac788d4a4eab30f13eba324cb670d4ee

obi1kenobi (approved):

Issue #86, opened 2022-08-18
What should the CLI look like?


epage:
Current, everything is under `cargo semver-checks check-release`.  Deeply nested subcommands like that should be in rare cases and we should be cautious of having long command names like that.

Do we need that extra subcommand?  Is there another way of expressing that idea?

Let's first breakdown expected workfows (which will help with #47).

The overall aim is to help with API evolution,. including
- Report to reviewers that a breaking change was made when it shouldn't
- Report to reviewers that a breaking change was made to ensure it gets documented
- Check local changes for unexpected breaking changes
- Check local changes for breaking changes to document
- As a user of a crate, find out what changed in its API
- Get a suggested minimal version to upgrade to
- Check that we aren't upgrading to too small of a version

obi1kenobi:
Possible additional operations:
- Run user-specified queries to check additional invariants
  - Example courtesy of @rcoh: "ensure third-party crates' types (modulo explicit allow-list) aren't leaking through this crate's API"
- Get a suggested kind of version upgrade (patch / minor / major), as opposed to a specific version number
  - Useful if labelling a GitHub PR with the outcome, since the label would be something like `M-semver-minor` and not a specific version number.

obi1kenobi:
To me, at least the "About to publish, please lint all" and the "Checking a PR, advise on semver changes" modes seem different enough to be subcommands rather than flags.

That said, I admit that I have only minimal experience designing CLIs, so it's entirely possible I'm missing something obvious.

epage:
> Example courtesy of @rcoh: "ensure third-party crates' types (modulo explicit allow-list) aren't leaking through this crate's API"

I know this is just an example but Isn't this a lint against a a specific snapshot rather than in comparing snapshots?  There is also an [RFC for this](https://rust-lang.github.io/rfcs/1977-public-private-dependencies.html?highlight=publi#)

obi1kenobi:
It can be phrased and implemented as "no new third-party types leak through compared to the last version" as well.

In general, though, I'm not sure every lint must always be across snapshots. Some of the lints in #5 are only over the "current" version. I'm not sure how to best make that work in the CLI, but the query engine can run those queries just fine — in `adapter.rs` the baseline crate info is always optional.

epage:
[...]

PR #207, opened 2022-12-03
Script that runs the lints on most popular crates

tonowak:
It downloads data from crates.io about the most downloaded crates and its versions, and then runs cargo-semver-checks on those crates to find false-positives or bugs in the tool.


scripts/run_on_popular_crates/src/main.rs
```diff
@@ -0,0 +1,127 @@
+// use std::io;
+// use std::fs::File;
+// use core::error::Error;
+use serde::Deserialize;
+use std::collections::BTreeMap;
+use std::path::Path;
+
+fn download_crates_data() -> String {
+    println!("Downloading db-dump.tar.gz (it can take a while)");
+    let resp = reqwest::blocking::get("https://static.crates.io/db-dump.tar.gz").unwrap();
```
obi1kenobi, line 10: Do we really need the full dump? Can't we just use an API client to fetch the most-downloaded crates directly, like this: https://docs.rs/crates_io_api/0.8.1/crates_io_api/struct.CratesQuery.html#method.set_sort

tonowak: Oh, so this is where the API is. I couldn't find it on the crates.io.

scripts/run_on_popular_crates/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+fn save_manifest(manifest: cargo_toml::Manifest<()>, path: &str) -> anyhow::Result<()> {
+    std::fs::write(path, toml::to_string(&manifest)?)?;
+    Ok(())
+}
+
+fn run_checks_on_manifest() -> anyhow::Result<()> {
+    let output = Command::new("../../../target/debug/cargo-semver-checks")
```
obi1kenobi, line 56: This seems to assume that a `cargo build` has already happened in debug mode, and would fail with an opaque error message if that isn't the case.

Perhaps it would be better to run a `cargo run` instead?

scripts/run_on_popular_crates/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+        .output()?;
+    std::io::stdout().write_all(&output.stdout)?;
+    std::io::stdout().write_all(&output.stderr)?;
+    println!();
+    Ok(())
+}
+
+fn main() -> anyhow::Result<()> {
+    let client = crates_io_api::SyncClient::new(
+        "crates_io_api: cargo-semver-checks (obi1kenobi82@gmail.com)",
```
obi1kenobi, line 72: Please put a link to the project's GitHub instead of my personal email here.

scripts/run_on_popular_crates/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+        Duration::from_millis(10),
```
obi1kenobi, line 73: This is in violation of crates.io's API policy, which states clients should not make requests more than once per second.
```suggestion
        Duration::from_millis(1000),
```

scripts/run_on_popular_crates/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+                ..Product::default()
+            };
+            Some(product)
+        },
+        dependencies: {
+            let project_with_features = DependencyDetail {
+                version: Some(version.to_string()),
+                features: features
+                    .iter()
+                    .map(|(key, _values)| key.clone())
```
obi1kenobi, line 35: Is this not just `.keys().cloned()`?

scripts/run_on_popular_crates/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+        .build();
+    query.set_page(1);
+
+    let crate_page = client.crates(query)?;
+
+    for crate_info in crate_page.crates.into_iter() {
+        let versions = client.get_crate(&crate_info.name)?.versions;
+        for (i, version_i0) in versions.iter().enumerate() {
+            if i + 1 < versions.len() {
+                let version_i1 = &versions[i + 1];
```
obi1kenobi, line 88: You want `tuple_windows()` here instead of rolling your own:
https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.tuple_windows

scripts/run_on_popular_crates/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+                if version_i0.yanked && !version_i1.yanked { // it gives more interesting results for testing
```
obi1kenobi, line 89: This would certainly give interesting results, but aren't we primarily interested in checking the most recent version vs its baseline? I'd assume we want to make sure:
- new crates can adopt cargo-semver-checks without running into any errors
- we don't keep re-checking mostly-the-same code over and over again, and instead make better use of our limited CPU time

I think this is fine to have as a mode behind a flag, but I think the primary mode of this tool would just be to check the most recent "normal" version of each crate against its baseline.

scripts/run_on_popular_crates/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
```
obi1kenobi, line 1: I don't think we want to put an entire crate in `scripts`. Future maintainers of this project would find that very confusing, and it's not in line with how Rust projects tend to be structured. Consider moving the crate into its own top-level directory.

tonowak: Can you suggest a good name for the crate?

scripts/run_on_popular_crates/src/main.rs
tonowak: Done in 81b00a95ec7520843f9cc283b042f4e246335713.

scripts/run_on_popular_crates/src/main.rs
tonowak: Done in f5696ca6eb60b601761e845dae60c5e84dc8b3ef.

scripts/run_on_popular_crates/src/main.rs
tonowak: Done in 44511f74a70ce7e3527727f08fa7610de5d9fce0.

scripts/run_on_popular_crates/src/main.rs
tonowak: Done in 4f463225ca84523b8d04f140ce3ffec64a8524e8.

scripts/run_on_popular_crates/src/main.rs
tonowak: Removed yanked releases in 4f463225ca84523b8d04f140ce3ffec64a8524e8.
It nerveless didn't work, because cargo doesn't allow setting a yanked release as a dependency.

scripts/run_on_popular_crates/src/main.rs
tonowak: 8e252e4cbbdd6711e070f7a48c18995a3c407e39

scripts/run_on_popular_crates/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+                    println!("{} {} {} {}", crate_info.name, version_i0.num, version_i1.num, version_i0.yanked);
+                    save_manifest(create_rustdoc_manifest_for_crate_version(&crate_info.name, &version_i0.num, &version_i0.features), "crate_current/Cargo.toml")?;
+                    save_manifest(create_rustdoc_manifest_for_crate_version(&crate_info.name, &version_i1.num, &version_i1.features), "crate_baseline/Cargo.toml")?;
```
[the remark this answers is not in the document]
tonowak: Done in 4c04d2620f91e4a030dce6ce0dc6e5006d9c4750.
It seems I found another bug because of it (sigh).

scripts/run_on_popular_crates/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+                    run_checks_on_manifest()?;
```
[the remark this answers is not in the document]
tonowak: Right. When the script will work I'll rewrite this part to log an appropriate information and skip this particular crate.

scripts/run_on_popular_crates/src/main.rs
tonowak: I plan to add `--all-releases` and `--only-latest-release` as commands when the script will work.

scripts/run_on_popular_crates/Cargo.toml
tonowak: Maybe `check-semver-of-registry-crates`?

obi1kenobi: Perhaps `registry-semver-scanner`? Or, a different direction, `semver-crater` as a riff on `crater`, the tool that tests new Rust releases against all registry crates?

`check-semver-of-registry-crates` is a bit long, especially for something we may need to type in.

Also, depending on the discussion in #86, you may be able to move this to `src/bin/` and configure the `Cargo.toml` to gate this code behind a new `unstable` feature.

tonowak:
This PR is now stacked on https://github.com/obi1kenobi/cargo-semver-checks/pull/314.

src/bin/semver-crater.rs
```diff
@@ -0,0 +1,114 @@
...
+                    false => None,
+                }
+            }));
+            filtered.push(
+                versions
+                    .last()
+                    .expect("received empty list of versions")
+                    .clone(),
+            );
+            filtered
```
tonowak, line 45: I won't deny -- this code is ugly as hell. But it works. And I don't really have any ideas on how to make it nicer.

tonowak:
The CLI needs some polishing, but I'm not experienced in stuff like that -- @obi1kenobi can you review it?

I'm thinking that `semver-crater` might be useful not only for data gathering, but also for convincing people that the tool is worthy. The `semver-crater` allows to easily check whether there has been any issues in the user's crate (just run a single `cargo` command that install `cargo-semver-checks` and runs it on all releases of the user's crate) and to show people "look, some popular crates (like clap) had issues, here is how in a single command you can check that it is indeed true and the tool even detects it".

Assuming `semver-crater` would be used in this manner, we have to think about how to make its CLI convenient to use.

PR #208, opened 2022-12-06
Upload binaries to github release in actions

staniewzki:
Added new release workflow, and made the ci workflow to trigger it when the publishing should happen.

There has been problems with openssl-sys crate. When cross compiling, openssl built for target is needed. Thus, I enabled vendored-openssl feature in crates-index, to make that simpler.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,6 +123,8 @@ jobs:
       - name: Install protobuf-compiler
         run: sudo apt install protobuf-compiler
 
+      - uses: Swatinem/rust-cache@v2
+
       - name: Run semver-checks
         run: cargo run semver-checks check-release --manifest-path="crate/core/Cargo.toml"
 
@@ -135,7 +137,7 @@ jobs:
   run-on-core-graphics:
     # Run cargo-semver-checks on a crate with no semver violations,
     # to make sure there are no false-positives.
-
+    #
     # cargo-semver-checks previously reported a false-positive here,
     # due to multiple methods by the same name being defined on the same type.
     # https://github.com/obi1kenobi/cargo-semver-checks/issues/193
@@ -161,6 +163,8 @@ jobs:
           toolchain: stable
           profile: minimal
           override: true
+
+      - uses: Swatinem/rust-cache@v2
 
       - name: Run semver-checks
         run: cargo run semver-checks check-release --manifest-path="crate/core-graphics/Cargo.toml"
@@ -213,11 +217,12 @@ jobs:
             # Unexpected outcome, indicates a bug.
             exit "$EXIT_CODE"
           fi
-      - name: Check the version
+      - name: Determine the tag name
         id: tag
         run: |
           export TAG_NAME="v$(./scripts/get_current_version.sh cargo-semver-checks)"
           echo "tag=$TAG_NAME" >> $GITHUB_OUTPUT
+
   pre-publish-checks:
     name: Check for semver compliance
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,8 +123,6 @@ jobs:
       - name: Install protobuf-compiler
         run: sudo apt install protobuf-compiler
 
-      - uses: Swatinem/rust-cache@v2
-
       - name: Run semver-checks
         run: cargo run semver-checks check-release --manifest-path="crate/core/Cargo.toml"
 
@@ -137,7 +135,7 @@ jobs:
   run-on-core-graphics:
     # Run cargo-semver-checks on a crate with no semver violations,
     # to make sure there are no false-positives.
-    #
+
     # cargo-semver-checks previously reported a false-positive here,
     # due to multiple methods by the same name being defined on the same type.
     # https://github.com/obi1kenobi/cargo-semver-checks/issues/193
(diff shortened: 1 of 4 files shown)


staniewzki:
Not sure how to test the workflow. The simplified version is working on my fork.
https://github.com/staniewzki/cargo-semver-check/actions/runs/3630752868
https://github.com/staniewzki/cargo-semver-check/actions/runs/3630772995

.github/workflows/ci.yml
```diff
@@ -186,10 +186,10 @@ jobs:
           git tag "v$CURRENT_VERSION"
           git push origin "v$CURRENT_VERSION"
```
obi1kenobi, line 187: If moving publishing to a workflow, I think it's better to move the tagging into that workflow too. For bonus points, maybe tag the version after `cargo publish` rather than before.

.github/workflows/release.yml
```diff
@@ -0,0 +1,76 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - v[0-9]+.*
+  workflow_dispatch:
```
obi1kenobi, line 7: Use `workflow_call` only here: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_call

We don't want to run on tags because then we'll get duplicate runs.

.github/workflows/release.yml
```diff
@@ -0,0 +1,76 @@
...
+    name: ${{ matrix.target }}
+    needs:
+      - create-release
+    strategy:
+      matrix:
+        include:
+          - target: aarch64-unknown-linux-gnu
+          - target: aarch64-unknown-linux-musl
+          - target: aarch64-apple-darwin
+            os: macos-11
```
obi1kenobi, line 45: Not `macos-latest`?

.github/workflows/release.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: true
+
+      - name: Publish
+        run: cargo publish
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+
```
obi1kenobi, line 27: I'd put the `git tag` step here, between the `cargo publish` and the GitHub release.

.github/workflows/release.yml
```diff
@@ -0,0 +1,76 @@
...
+          - target: aarch64-unknown-linux-gnu
+          - target: aarch64-unknown-linux-musl
+          - target: aarch64-apple-darwin
+            os: macos-11
+          - target: x86_64-unknown-linux-gnu
+          - target: x86_64-unknown-linux-musl
+          - target: x86_64-apple-darwin
+            os: macos-10.15
+          - target: x86_64-pc-windows-msvc
+            os: windows-2019
```
obi1kenobi, line 51: Not `windows-latest`?

Cargo.toml
```diff
@@ -31,7 +31,7 @@ ignore = "0.4.18"
 clap-verbosity-flag = "2.0.0"
 log = "0.4.17"
 git2 = { version = "0.15.0", default-features = false }
-crates-index = "0.18.9"
+crates-index = { version = "0.18.9", features = ["vendored-openssl"] }
```
obi1kenobi, line 34: @epage any concerns with adding this feature? It seems to make the build simpler and more reliable, but I wanted to double-check with you in case there were any downsides I'm not familiar with.

.github/workflows/release.yml
```diff
@@ -0,0 +1,76 @@
...
+  workflow_dispatch:
+
+env:
+  RUST_BACKTRACE: 1
+  CARGO_TERM_COLOR: always
+
+jobs:
+  create-release:
+    name: Create release
+    runs-on: ubuntu-latest
```
obi1kenobi, line 16: I'd add another safety check here to make sure this is all running on the `main` branch and wasn't accidentally called from somewhere else:
```suggestion
    if: github.ref == 'refs/heads/main'
    runs-on: ubuntu-latest
```

obi1kenobi (commented):
Excellent! Thanks for putting this together, excited to ship it 🚀

.github/workflows/release.yml
staniewzki: Workflows don't trigger off events happening in github actions, so there wouldn't be any duplicates

.github/workflows/release.yml
staniewzki: It is copied from https://github.com/taiki-e/cargo-hack/blob/202e6e59d491c9202ce148c9ef423853267226db/.github/workflows/release.yml#L47-L84, but I think `macos-latest` makes more sense

.github/workflows/release.yml
staniewzki: Well, the ref here is not main, so that's a bit problematic

.github/workflows/release.yml
obi1kenobi: Neat, TIL! Even so, if we use `workflow_call` we still don't need this to run on tags.

.github/workflows/release.yml
obi1kenobi: If using a workflow-to-workflow call as I suggested, it would be main since the workflow wouldn't be triggered on a tag push.

.github/workflows/release.yml
thomaseizinger: > Workflows don't trigger off events happening in github actions, so there wouldn't be any duplicates

Just FYI: They can if you use a Personal Access Token.

.github/workflows/ci.yml
staniewzki: Both `taiki-e/create-gh-release-action@v1` and `taiki-e/upload-rust-binary-action@v1` must run on the tag ref.

I tried to set a enviormental variable and use it later on, but this somehow failed on Windows. 

Actually, the pair programming sounds like a cool idea, as I have spent some time already on trial and error with moderate success.

obi1kenobi: It seems like they would accept a `ref` argument though:
https://github.com/taiki-e/create-gh-release-action/blob/main/action.yml#L45-L49
https://github.com/taiki-e/upload-rust-binary-action/blob/main/action.yml#L61-L66

If we `git tag` and push the tag before running those actions, and set their `ref` to the tag name, I think that should work.

Let me know over Zulip what times are best for pair-programming for you in the next few days, and we can make something work.

.github/workflows/ci.yml
```diff
@@ -123,8 +123,6 @@ jobs:
       - name: Install protobuf-compiler
         run: sudo apt install protobuf-compiler
 
-      - uses: Swatinem/rust-cache@v2
-
```
obi1kenobi, line 127: Why remove the caching here?

.github/workflows/ci.yml
```diff
@@ -137,7 +135,7 @@ jobs:
   run-on-core-graphics:
     # Run cargo-semver-checks on a crate with no semver violations,
     # to make sure there are no false-positives.
-    #
+
```
obi1kenobi, line 138: This looks like an unintentional change, consider putting it back.

.github/workflows/ci.yml
```diff
@@ -164,48 +162,41 @@ jobs:
           profile: minimal
           override: true
 
-      - uses: Swatinem/rust-cache@v2
-
```
obi1kenobi, line 168: Why remove the caching here as well?

.github/workflows/ci.yml
```diff
@@ -222,7 +213,11 @@ jobs:
             # Unexpected outcome, indicates a bug.
             exit "$EXIT_CODE"
           fi
-
+      - name: Check the version
```
obi1kenobi, line 216:
```suggestion
      - name: Determine the tag name
```

obi1kenobi (commented):
Broadly looks good! I'm just not sure why some of the caching is removed.

I also highly recommend doing a self-review of the PR in the GitHub PR interface. The change of perspective will usually let you find most of the kinds of things reviewers will also flag, and the fewer comments a PR has, the faster it'll merge.

staniewzki:
Ah, sorry for that, no idea how those changes made it to the PR. Thanks for the GitHub PR interface tip!

obi1kenobi (approved):

obi1kenobi:
Planning to make a new release either later today or tomorrow morning to test this out. Great job putting it together!

PR #234, opened 2022-12-19
Ensure the lint files are part of the binary file itself.

obi1kenobi:
As part of the `add_lints!()` macro, we had accidentally made the lints no longer be part of the `cargo-semver-checks` binary itself. This means that everything worked fine as long as it was run from the `cargo-semver-checks` repo itself, and failed spectacularly as soon as that was not the case (e.g. after `cargo install cargo-semver-checks`).

This PR re-adds the lints into the binary using `include_str!()`.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -75,7 +75,7 @@ impl SemverQuery {
     pub(crate) fn all_queries() -> BTreeMap<String, SemverQuery> {
         let mut queries = BTreeMap::default();
         for query_text in get_query_text_contents() {
-            let query: SemverQuery = ron::from_str(&query_text).unwrap_or_else(|e| {
+            let query: SemverQuery = ron::from_str(query_text).unwrap_or_else(|e| {
                 panic!(
                     "\
 Failed to parse a query: {}
@@ -318,13 +318,13 @@ macro_rules! add_lints {
             )*
         }
 
-        fn get_query_text_contents() -> Vec<String> {
+        // No way to avoid this lint -- the push() calls are macro-generated.
+        #[allow(clippy::vec_init_then_push)]
+        fn get_query_text_contents() -> Vec<&'static str> {
             let mut temp_vec = Vec::new();
             $(
                 temp_vec.push(
-                    std::fs::read_to_string(
-                        format!("./src/lints/{}.ron", stringify!($name))
-                    ).unwrap()
+                    include_str!(concat!("lints/", stringify!($name), ".ron"))
                 );
             )*
             temp_vec


tonowak:
What :O

tonowak:
I see. Have you already thought about possible tests that could catch this?

obi1kenobi:
> What :O

I know, right? At compile-time, `include_str!()` has a stable `.` directory, so in my code review I completely missed that we had switched to a runtime-evaluated `std::fs::read_to_string()` with a relative path, which could point anywhere since the program itself doesn't attempt to set it to anything. And since we also don't bundle the lint files as part of the build, this was never going to work.

I found it this by accident while testing something else with the locally-built binary for the newest version, so we got lucky. Otherwise, I think this would have made it through our entire release process and would have broken our users.

It's probably a good idea to add a release-time check that does `cargo build --release`, then does a `cd` into the release directory and runs the built binary directly on some rustdoc JSON files (e.g. from the test suite).

obi1kenobi:
Would you be interested in adding that CI check, or should I go for it?

src/query.rs
```diff
@@ -318,13 +318,13 @@ macro_rules! add_lints {
             )*
         }
 
-        fn get_query_text_contents() -> Vec<String> {
+        // No way to avoid this lint -- the push() calls are macro-generated.
+        #[allow(clippy::vec_init_then_push)]
```
tonowak, line 322: Is that a new clippy lint? Should I update my clippy?

src/query.rs
```diff
@@ -75,7 +75,7 @@ impl SemverQuery {
     pub(crate) fn all_queries() -> BTreeMap<String, SemverQuery> {
         let mut queries = BTreeMap::default();
         for query_text in get_query_text_contents() {
-            let query: SemverQuery = ron::from_str(&query_text).unwrap_or_else(|e| {
+            let query: SemverQuery = ron::from_str(query_text).unwrap_or_else(|e| {
```
tonowak, line 78: Shouldn't that have been caught by clippy in my PR?

src/query.rs
```diff
@@ -318,13 +318,13 @@ macro_rules! add_lints {
...
+        // No way to avoid this lint -- the push() calls are macro-generated.
+        #[allow(clippy::vec_init_then_push)]
+        fn get_query_text_contents() -> Vec<&'static str> {
             let mut temp_vec = Vec::new();
             $(
                 temp_vec.push(
-                    std::fs::read_to_string(
-                        format!("./src/lints/{}.ron", stringify!($name))
-                    ).unwrap()
+                    include_str!(concat!("lints/", stringify!($name), ".ron"))
```
tonowak, line 327: Why concat and not the format that I wrote?

tonowak:
> Would you be interested in adding that CI check, or should I go for it?

I can try doing that this week. Will you create an appropriate issue?

src/query.rs
obi1kenobi: `query_text` changed from `String` to `&'static str` here. The original code linted clean, but with the type change this was a needless borrow.

src/query.rs
obi1kenobi: I don't think it's new, it's been there since 1.51 per this page: https://rust-lang.github.io/rust-clippy/master/index.html#vec_init_then_push

The old code didn't trigger that lint for me, though. Not sure why only the new code gets it.

src/query.rs
obi1kenobi: `include_str!()` demands a string *literal*, not just a string-like value:
image

PR #240, opened 2022-12-20
Describe the new more-automated process for adding new lints.

obi1kenobi:


staniewzki (approved):
I have read through these instructions while working on https://github.com/obi1kenobi/cargo-semver-checks/pull/249.

The only issue I encountered was that when adding the new lint, the scenario I that was meant to be checked was already present in two test crates, and there's no solution in the instructions. That's not that much of a problem, but it's probably not optimal to have exact same scenario in multiple test crates in the long run. On the other hand, changing the existing test crates could be error-prone.

Another nitpick: I had to `chmod +x scripts/make_new_lint.sh` the script to make it executable (as it is more convenient this way, personally), updated the permissions in the mentioned PR.

obi1kenobi:
Thanks for the review!

> The only issue I encountered was that when adding the new lint, the scenario I that was meant to be checked was already present in two test crates, and there's no solution in the instructions. That's not that much of a problem, but it's probably not optimal to have exact same scenario in multiple test crates in the long run.

I think it's okay to have the same semver issue in multiple crates. Absolute deduplication doesn't buy us much, and it would be confusing and difficult to accomplish, since sometimes we're testing that a lint catches an issue, and sometimes we're testing that a lint _doesn't_ catch semver issues that are the domain of another lint.

> Another nitpick: I had to `chmod +x scripts/make_new_lint.sh` the script to make it executable (as it is more convenient this way, personally), updated the permissions in the mentioned PR.

Good catch! I'll open a PR for that. I really wish WSL2's Windows filesystem had proper "rwx" flags.

Issue #243, opened 2022-12-21
Tests fail for unknown reasons


tonowak:
### Steps to reproduce the bug with the above code

```sh
git clone git@github.com:obi1kenobi/cargo-semver-checks.git
```
```
Cloning into 'cargo-semver-checks'...
remote: Enumerating objects: 1886, done.
remote: Counting objects: 100% (1886/1886), done.
remote: Compressing objects: 100% (648/648), done.
remote: Total 1886 (delta 1143), reused 1826 (delta 1122), pack-reused 0
Receiving objects: 100% (1886/1886), 388.04 KiB | 671.00 KiB/s, done.
Resolving deltas: 100% (1143/1143), done.
```
```sh
cd cargo-semver-checks
./scripts/regenerate_test_rustdocs.sh
```
```
Generating: auto_trait_impl_removed/new
~/zpp/cargo-semver-checks/test_crates/auto_trait_impl_removed/new ~/zpp/cargo-semver-checks
 Documenting auto_trait_impl_removed v0.1.0 (/home/tonowak/zpp/cargo-semver-checks/test_crates/auto_trait_impl_removed/new)
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
~/zpp/cargo-semver-checks
Generating: auto_trait_impl_removed/old
~/zpp/cargo-semver-checks/test_crates/auto_trait_impl_removed/old ~/zpp/cargo-semver-checks
 Documenting auto_trait_impl_removed v0.1.0 (/home/tonowak/zpp/cargo-semver-checks/test_crates/auto_trait_impl_removed/old)
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
~/zpp/cargo-semver-checks
Generating: constructible_struct_adds_field/new
~/zpp/cargo-semver-checks/test_crates/constructible_struct_adds_field/new ~/zpp/cargo-semver-checks
 Documenting constructible_struct_adds_field v0.1.0 (/home/tonowak/zpp/cargo-semver-checks/test_crates/constructible_struct_adds_field/new)
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
~/zpp/cargo-semver-checks
Generating: constructible_struct_adds_field/old
~/zpp/cargo-semver-checks/test_crates/constructible_struct_adds_field/old ~/zpp/cargo-semver-checks
[...]

PR #245, opened 2022-12-21
Generate CI test rustdocs with the toolchain being tested.

obi1kenobi:
Explicitly make sure that the rustdoc generated for testing is created using the toolchain being tested.


obi1kenobi:
Revealed by #243, it seems like our beta and nightly tests were running `cargo-semver-checks` on beta/nightly but on rustdocs generated by stable.

@tonowak and I are on it.

Issue #235, opened 2022-12-19
Run `cargo build` and test built binary to ensure it works


obi1kenobi:
> I found it this by accident while testing something else with the locally-built binary for the newest version, so we got lucky. Otherwise, I think this would have made it through our entire release process and would have broken our users.
>
> It's probably a good idea to add a CI check that does `cargo build`, then does a `cd` into the release directory and runs the built binary directly on some rustdoc JSON files (e.g. from the test suite). If we put it after the `cargo test` step then it probably will have minimal cost, since `cargo test` will already have done almost the entire build.

_Originally posted by @obi1kenobi in https://github.com/obi1kenobi/cargo-semver-checks/issues/234#issuecomment-1358418554_

#234 got lucky, and caught and fixed a critical issue that would otherwise have made it through our entire release process without getting caught. Let's amend our test setup to prevent it going forward.

PR #246, opened 2022-12-23
Run cargo build and test built binary to ensure it works

tonowak:
Closes https://github.com/obi1kenobi/cargo-semver-checks/issues/235


epage (commented):
imo if you have this complex of a step, it should probably broken out.

What I'd recommend doing is writing this as a test. Cargo builds binaries before running tests, so you can write end-to-end cli tests in Rust. I used to write mine with assert_cmd/assert_fs but have switched to snapbox

tonowak:
Thanks for the information! I wrote the test using `assert_cmd`.

Cargo.toml
```diff
@@ -37,3 +37,4 @@ bugreport = "0.5.0"
 itertools = "0.10.5"
 cargo_toml = "0.13.0"
 toml = "0.5.9"
+assert_cmd = "2.0"
```
epage, line 40: This should be added as a `dev-dependency`

src/main.rs
```diff
@@ -282,8 +282,43 @@ struct CheckRelease {
     verbosity: clap_verbosity_flag::Verbosity<clap_verbosity_flag::InfoLevel>,
 }
 
-#[test]
-fn verify_cli() {
-    use clap::CommandFactory;
-    Cargo::command().debug_assert()
+#[cfg(test)]
+mod tests {
```
epage, line 286: > thread 'tests::verify_binary_contains_lints' panicked at 'called `Result::unwrap()` on an `Err` value: CargoError { cause: Some(NotFoundError { path: "/home/runner/work/cargo-semver-checks/cargo-semver-checks/target/debug/cargo-semver-checks" }) }', src/main.rs:298:69

I think the guarantee that the binary is built first is only available for "integration" / "black box" tests (in the `tests/` directory)

src/main.rs
```diff
@@ -282,43 +282,8 @@ struct CheckRelease {
...
-            "function_const_removed",
-            "function_unsafe_added",
-        ] {
-            assert_on_crate_pair(crate_pair).failure();
-        }
-    }
+#[test]
+fn verify_cli() {
+    use clap::CommandFactory;
+    Cargo::command().debug_assert()
```
obi1kenobi, line 288: Did you mean to change this and pull it out of `mod tests`?

tonowak: I decided to keep `src/main.rs` how it was originally (by simply reverting the `src/main.rs` after my changes to the state before my changes). Should I nevertheless move it to `mod tests`?

obi1kenobi (approved):

src/main.rs
obi1kenobi: Sorry, I must have been looking at a stale version of the PR.

Issue #242, opened 2022-12-20
New lint: exhaustive tuple struct changed to plain struct


obi1kenobi:
This is breaking even if the fields have the same names, because tuple struct patterns won't work.

Source: Rust for Rustaceans, Chapter 3, "Type Modifications", page 51

PR #249, opened 2022-12-27
New lint:  exhaustive tuple struct changed to plain struct

staniewzki:
resolves https://github.com/obi1kenobi/cargo-semver-checks/issues/242

Signed-off-by: Michał Staniewski <m.staniewzki@gmail.com>


src/lints/tuple_struct_to_plain_struct.ron
```diff
@@ -0,0 +1,62 @@
...
+
+                        importable_path {
+                            path @tag @output
+                        }
+                    }
+                }
+            }
+            current {
+                item {
+                    ... on Struct {
```
obi1kenobi, line 35: If the struct didn't use to be `#[non_exhaustive]` but now became `#[non_exhaustive]`, that's a different breaking change that we perhaps shouldn't report here: non-exhaustive structs cannot be constructed with a literal nor pattern-matched, and the author of that code has explicitly marked the struct as such.

Instead, we'd report that under the "struct used to be exhaustive but now is non-exhaustive" lint.

We probably want an `attr @filter(...)` here and a test case to check for this.

src/lints/tuple_struct_to_plain_struct.ron
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "tuple": "tuple",
+        "plain": "plain",
+        "non_exhaustive": "#[non_exhaustive]",
+        "zero": 0,
+    },
+    error_message: "A publicly-visible, exhaustive tuple struct with all fields public changed to normal (curly-braces) struct, which cannot be constructed using the same struct literal syntax.",
```
obi1kenobi, line 60:
```suggestion
    error_message: "A publicly-visible, exhaustive tuple struct with pub fields changed to normal (curly-braces) struct, which cannot be constructed using the same struct literal syntax.",
```
This message is a bit long. This is my attempt at making it a bit more concise.

obi1kenobi (commented):
Looks good overall, just a minor tweak or two. I didn't see you were changing the flags on `scripts/make_new_lint.sh` so I merged a PR changing that already.

tonowak:
I think this PR is waiting for your approval @obi1kenobi?

obi1kenobi (approved):

PR #252, opened 2022-12-30
New lints: Trait becomes/stops being unsafe

staniewzki:
note: depends on https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/19


Cargo.toml
```diff
@@ -14,7 +14,7 @@ categories = ["command-line-utilities", "development-tools::cargo-plugins"]
 
 [dependencies]
 trustfall_core = "0.1.1"
-trustfall_rustdoc = { version = "0.6.0", features = ["v16", "v21", "v22", "v23"] }
+trustfall_rustdoc = { path = "../trustfall-rustdoc", version = "0.6.0", features = ["v16", "v21", "v22", "v23"] }
```
tonowak, line 17: :)

tonowak (dismissed):

Cargo.toml
staniewzki: Yes, I know, but thanks anyways.

src/lints/trait_unsafe_removed.ron
```diff
@@ -0,0 +1,49 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+    },
+    error_message: "A `pub unsafe` trait became safe, so `unsafe impl` block can no longer be used.",
```
obi1kenobi, line 47: Slight wording change for you to consider:
```suggestion
    error_message: "A formerly-unsafe public trait became safe, so it is no longer allowed to be implemented with an `unsafe impl` block.",
```

test_crates/trait_unsafe_added/old/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+// Public trait becomes unsafe, should get reported.
+pub trait TraitBecomesUnsafe {}
+
+// Trait that was not publicly-visible becomes unsafe, shouldn't get reported.
+trait TraitBecomesPublicAndUnsafe {}
+
+// Private trait becomes unsafe, shouldn't get reported.
+trait PrivateTraitBecomesUnsafe {}
```
obi1kenobi, line 8: Consider adding the reverse test case too: a `pub trait` becomes both private and unsafe. It shouldn't get reported for being unsafe because the bigger problem is that it's now private.

test_crates/trait_unsafe_removed/old/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+// Public unsafe trait becomes safe, should get reported.
+pub unsafe trait TraitBecomesSafe {}
+
+// Trait that was not publicly-visible becomes safe, shouldn't get reported.
+unsafe trait TraitBecomesPublicAndSafe {}
```
obi1kenobi, line 5: Consider the reverse test case here as well: `pub unsafe trait` becomes safe and private.

obi1kenobi (approved):

PR #255, opened 2023-01-01
Exclude yanked releases when choosing baseline version

### LINKED ISSUE #254: Yanked release selected as baseline
The latest matching release from crates.io is selected by `check-release` even if it has been yanked, which results in `error: failed to select a version for the requirement (...)`.

### How to reproduce

- Clone `cargo-semver-checks`.
- Change version in `Cargo.toml` to `0.4.5`.
- Run `cargo run semver-checks check-release`.

mgr0dzicki:
Closes #254

Extracts the selection of baseline version to a separate function, adds tests for it.

### DIFF
--- a/src/baseline.rs
+++ b/src/baseline.rs
@@ -364,6 +364,8 @@ mod tests {
     use super::choose_baseline_version;
 
     fn new_mock_version(version_name: &str, yanked: bool) -> Version {
+        // `crates_index::Version` cannot be created explicitly, as all its fields
+        // are private, so we use the fact that it can be deserialized.
         serde_json::from_value(serde_json::json!({
             "name": "test-crate",
             "vers": version_name,
@@ -376,6 +378,8 @@ mod tests {
     }
 
     fn new_crate(versions: &[Version]) -> Crate {
+        // `crates_index::Crate` cannot be created explicitly, as its field
+        // is private, so we use the fact that it can be deserialized.
         serde_json::from_value(serde_json::json!({ "versions": versions })).unwrap()
     }
 
--- a/src/baseline.rs
+++ b/src/baseline.rs
@@ -403,7 +403,8 @@ mod tests {
                 .collect(),
         );
         let current_version = current_version_name.map(|version_name| {
-            semver::Version::parse(version_name).expect("Failed to parse hand-written string as a valid version name.")
+            semver::Version::parse(version_name)
+                .expect("Failed to parse hand-written string as a valid version name.")
         });
         let chosen_baseline = choose_baseline_version(&crate_, current_version.as_ref())
             .expect("choose_baseline_version should not return any error");
--- a/src/baseline.rs
+++ b/src/baseline.rs
@@ -391,4 +391,43 @@ mod tests {
             "1.2.0".to_string()
         );
     }
+
+    #[test]
+    fn choose_baseline_version_not_latest() {
+        let crate_ = new_crate(&[
+            new_mock_version("1.2.0", false),
+            new_mock_version("1.2.1", false),
+        ]);
+        let current_version = semver::Version::new(1, 2, 0);
+        assert_eq!(
+            choose_baseline_version(&crate_, Some(&current_version)).unwrap(),
+            "1.2.0".to_string()
+        );
+    }
+
+    #[test]
+    fn choose_baseline_version_pre_release() {
+        let crate_ = new_crate(&[
+            new_mock_version("1.2.0", false),
+            new_mock_version("1.2.1-rc1", false),
+        ]);
+        let current_version = semver::Version::parse("1.2.1-rc2").unwrap();
+        assert_eq!(
+            choose_baseline_version(&crate_, Some(&current_version)).unwrap(),
+            "1.2.0".to_string()
+        );
+    }
+
+    #[test]
+    fn choose_baseline_version_no_current() {
+        let crate_ = new_crate(&[
+            new_mock_version("1.2.0", false),
+            new_mock_version("1.2.1-rc1", false),
+            new_mock_version("1.3.1", true),
+        ]);
+        assert_eq!(
+            choose_baseline_version(&crate_, None).unwrap(),
+            "1.2.0".to_string()
+        );
+    }
 }
--- a/src/baseline.rs
+++ b/src/baseline.rs
@@ -403,11 +403,11 @@ mod tests {
                 .collect(),
         );
         let current_version = current_version_name.map(|version_name| {
-            semver::Version::parse(version_name).expect("current_version_name should be valid")
+            semver::Version::parse(version_name).expect("Failed to parse hand-written string as a valid version name.")
         });
-        let choosen_baseline = choose_baseline_version(&crate_, current_version.as_ref())
+        let chosen_baseline = choose_baseline_version(&crate_, current_version.as_ref())
             .expect("choose_baseline_version should not return any error");
-        assert_eq!(choosen_baseline, expected.to_owned());
+        assert_eq!(chosen_baseline, expected.to_owned());
     }
 
     #[test]
--- a/src/baseline.rs
+++ b/src/baseline.rs
@@ -251,6 +251,7 @@ fn choose_baseline_version(
         let mut instances = crate_
             .versions()
             .iter()
+            .filter(|i| !i.is_yanked())
             .filter_map(|i| semver::Version::parse(i.version()).ok())
             // For unpublished changes when the user doesn't increment the version
             // post-release, allow using the current version as a baseline.
@@ -353,5 +354,41 @@ fn need_retry(res: Result<(), crates_index::Error>) -> anyhow::Result<bool> {
             }
         }
         Err(err) => Err(err.into()),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crates_index::{Crate, Version};
+
+    use super::choose_baseline_version;
+
+    fn new_mock_version(version_name: &str, yanked: bool) -> Version {
+        serde_json::from_value(serde_json::json!({
+            "name": "test-crate",
+            "vers": version_name,
+            "deps": [],
+            "features": {},
+            "yanked": yanked,
+            "cksum": "00".repeat(32),
+        }))
+        .unwrap()
+    }
+
+    fn new_crate(versions: &[Version]) -> Crate {
+        serde_json::from_value(serde_json::json!({ "versions": versions })).unwrap()
+    }
+
+    #[test]
+    fn choose_baseline_version_yanked() {
+        let crate_ = new_crate(&[
+            new_mock_version("1.2.0", false),
+            new_mock_version("1.2.1", true),
+        ]);
+        let current_version = semver::Version::new(1, 2, 2);
+        assert_eq!(
+            choose_baseline_version(&crate_, Some(&current_version)).unwrap(),
+            "1.2.0".to_string()
+        );
     }
 }
--- a/src/baseline.rs
+++ b/src/baseline.rs
@@ -278,12 +278,14 @@ fn choose_baseline_version(
                 )
             })
     } else {
-        // If there is no normal version (not yanked and not a pre-release)
-        // choosing the latest one anyway is more reasonable than throwing an
-        // error, as there is still a chance that it is what the user expects.
         let instance = crate_
             .highest_normal_version()
-            .unwrap_or_else(|| crate_.highest_version())
+            .unwrap_or_else(|| {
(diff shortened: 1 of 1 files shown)


src/baseline.rs
```diff
@@ -340,3 +356,82 @@ fn need_retry(res: Result<(), crates_index::Error>) -> anyhow::Result<bool> {
...
+            "yanked": yanked,
+            "cksum": "00".repeat(32),
+        }))
+        .unwrap()
+    }
+
+    fn new_crate(versions: &[Version]) -> Crate {
+        // `crates_index::Crate` cannot be created explicitly, as its field
+        // is private, so we use the fact that it can be deserialized.
+        serde_json::from_value(serde_json::json!({ "versions": versions })).unwrap()
```
tonowak, line 383: From what I remember, we're trying to use `expect` (or any alternatives with custom error messages) instead of `unwrap` for better debugging messages.

src/baseline.rs
```diff
@@ -340,3 +356,82 @@ fn need_retry(res: Result<(), crates_index::Error>) -> anyhow::Result<bool> {
...
+    }
+
+    #[test]
+    fn choose_baseline_version_yanked() {
+        let crate_ = new_crate(&[
+            new_mock_version("1.2.0", false),
```
tonowak, line 389: Each `new_crate` call is done by passing a list of `new_mock_version`. I wonder, wouldn't it be more concise to pass a list of tuples/structs (`version`, `is_yanked`), instead of a list of objects constructed by passing `version` and `is_yanked`?

This comment is not a request to change anything in the code -- I'm just curious whether this would be more readable/concise and I'm open for discussion.

src/baseline.rs
```diff
@@ -340,3 +356,82 @@ fn need_retry(res: Result<(), crates_index::Error>) -> anyhow::Result<bool> {
...
+    }
+
+    fn new_crate(versions: &[Version]) -> Crate {
+        // `crates_index::Crate` cannot be created explicitly, as its field
+        // is private, so we use the fact that it can be deserialized.
+        serde_json::from_value(serde_json::json!({ "versions": versions })).unwrap()
+    }
+
+    #[test]
+    fn choose_baseline_version_yanked() {
```
tonowak, line 387: This function name is misleading. It is not a yanked version that is being chosen.

src/baseline.rs
```diff
@@ -340,3 +356,82 @@ fn need_retry(res: Result<(), crates_index::Error>) -> anyhow::Result<bool> {
...
+        ]);
+        let current_version = semver::Version::parse("1.2.1-rc2").unwrap();
+        assert_eq!(
+            choose_baseline_version(&crate_, Some(&current_version)).unwrap(),
+            "1.2.0".to_string()
+        );
+    }
+
+    #[test]
+    fn choose_baseline_version_no_current() {
```
tonowak, line 426: This function name is possibly misleading. From the name, I would expect the test to have a `current` version and a list of crates that doesn't have this `current` version.

src/baseline.rs
```diff
@@ -242,6 +243,43 @@ fn create_rustdoc_manifest_for_crate_version(
...
+                anyhow::format_err!(
+                    "No available baseline versions for {}@{}",
+                    crate_.name(),
+                    current
+                )
+            })
+    } else {
+        let instance = crate_
+            .highest_normal_version()
+            .unwrap_or_else(|| crate_.highest_version())
```
tonowak, line 277: I know that this code is unchanged in this PR, but I don't really understand this line -- maybe it could be commented to explain what's going on here @obi1kenobi. What is considered a "normal" version and why it is OK to choose a "not normal" (whatever is "normal") version when there isn't a normal one? Assuming that a "normal" possibly means non-yanked and non-pre-release, what would happen when there are only "not normal" versions?

src/baseline.rs
```diff
@@ -259,30 +297,8 @@ impl BaselineLoader for RegistryBaseline {
         // - Most likely the user cares about the last official release
```
obi1kenobi, line 297: This comment block seems relevant to the code block you extracted into a function. Probably move it there too?

src/baseline.rs
obi1kenobi: Your guess was correct, it's "not yanked, not pre-release": https://docs.rs/crates-index/latest/crates_index/struct.Crate.html#method.highest_normal_version

If there's no "normal" version, then our options are:
- error
- choose the best option from what's available

The choice isn't necessarily obvious, and this isn't my code. But I think the decision to try to "guess what the user wants" in this case rather than making it a hard error is a reasonable one: in the worst case, we error anyway, and in the best case we guess correctly and delight the user.

Leaving a comment is reasonable. It might just restate what the docs for the two methods on `Crate` already say, but that's alright if it saves the next reader of the code a Google search.

obi1kenobi (commented):
I'm excited to have test coverage on this tricky logic! Thanks for putting it together.

It would be good to add two more test cases that check & document that when all versions are either yanked or pre-releases, cargo-semver-checks will pick the largest-numbered version regardless of whether it's been yanked or is a pre-release. (One test where it's yanked and one where it's a pre-release.)

src/baseline.rs
mgr0dzicki: I've done some refactoring to avoid this code duplication in tests, how do you find it now?

src/baseline.rs
mgr0dzicki: I'm not really good at naming thing, what better proposition do you have?

src/baseline.rs
mgr0dzicki: And same question here, do you have some idea how can we name it better?

src/baseline.rs
mgr0dzicki: Changed all `unwrap`s to `expect`s, please have a look at the messages.

mgr0dzicki:
> It would be good to add two more test cases that check & document that when all versions are either yanked or pre-releases, cargo-semver-checks will pick the largest-numbered version regardless of whether it's been yanked or is a pre-release. (One test where it's yanked and one where it's a pre-release.)

Added! They revealed that my logic of handling yanked versions was not behaving as it should (it was ignoring yanked releases when there weren't any normal one). I fixed this.

src/baseline.rs
```diff
@@ -340,3 +360,104 @@ fn need_retry(res: Result<(), crates_index::Error>) -> anyhow::Result<bool> {
...
+        // are private, so we use the fact that it can be deserialized.
+        serde_json::from_value(serde_json::json!({
+            "name": "test-crate",
+            "vers": version_name,
+            "deps": [],
+            "features": {},
+            "yanked": yanked,
+            "cksum": "00".repeat(32),
+        }))
+        .expect("Parsing JSON as crates_index::Version should not result in any errors.")
```
tonowak, line 381: Something I learned from @obi1kenobi -- this message describes an invariant, but it makes more sense in an error message to describe what failed. So maybe consider writing it as `Failed to create crates_index::Version from a hand-written JSON`.

src/baseline.rs
tonowak: Can you leave a comment here (in this code) in this PR @mgr0dzicki ?

src/baseline.rs
tonowak: What do you think about `baseline_choosing_logic_skips_yanked`?

src/baseline.rs
tonowak: What do you think about `baseline_choosing_logic_without_current_picks_latest_normal`? A bit long, but I don't see any reason to make it shorter.

src/baseline.rs
```diff
@@ -374,64 +374,68 @@ mod tests {
...
-    fn new_crate(versions: &[Version]) -> Crate {
+    fn new_crate(versions: Vec<Version>) -> Crate {
         // `crates_index::Crate` cannot be created explicitly, as its field
         // is private, so we use the fact that it can be deserialized.
-        serde_json::from_value(serde_json::json!({ "versions": versions })).unwrap()
+        serde_json::from_value(serde_json::json!({ "versions": versions }))
+            .expect("Parsing JSON as crates_index::Crate should not result in any errors.")
+    }
+
+    fn test_choose_baseline_version(
```
tonowak, line 391: This function name doesn't tell much.

src/baseline.rs
mgr0dzicki: Done! Please have a look at it @tonowak

src/baseline.rs
```diff
@@ -340,3 +363,113 @@ fn need_retry(res: Result<(), crates_index::Error>) -> anyhow::Result<bool> {
...
+        expected: &str,
+    ) {
+        let crate_ = new_crate(
+            versions
+                .into_iter()
+                .map(|(version, yanked)| new_mock_version(version, yanked))
+                .collect(),
+        );
+        let current_version = current_version_name.map(|version_name| {
+            semver::Version::parse(version_name).expect("current_version_name should be valid")
```
tonowak, line 406: Here you also have `expect`s as invariants.

src/baseline.rs
```diff
@@ -340,3 +363,113 @@ fn need_retry(res: Result<(), crates_index::Error>) -> anyhow::Result<bool> {
...
+        });
+        let choosen_baseline = choose_baseline_version(&crate_, current_version.as_ref())
+            .expect("choose_baseline_version should not return any error");
```
tonowak, line 409: I think it should be `chosen`, not `choosen`?

mgr0dzicki: definitely ;)

src/baseline.rs
```diff
@@ -340,3 +363,105 @@ fn need_retry(res: Result<(), crates_index::Error>) -> anyhow::Result<bool> {
...
+        // are private, so we use the fact that it can be deserialized.
+        serde_json::from_value(serde_json::json!({
+            "name": "test-crate",
+            "vers": version_name,
+            "deps": [],
+            "features": {},
+            "yanked": yanked,
+            "cksum": "00".repeat(32),
+        }))
+        .expect("Failed to create crates_index::Version from a hand-written JSON.")
```
tonowak, line 384: @obi1kenobi can you describe what exactly should be the format of `expect` messages? I remember that you said that they should start with a small letter and not end with a comma. But also, here it's explained that the string should describe an invariant, not what went wrong: https://stackoverflow.com/questions/66362625/why-is-rusts-expect-called-expect

src/baseline.rs
```diff
@@ -340,3 +363,105 @@ fn need_retry(res: Result<(), crates_index::Error>) -> anyhow::Result<bool> {
...
+                .into_iter()
+                .map(|(version, yanked)| new_mock_version(version, yanked))
+                .collect(),
+        );
+        let current_version = current_version_name.map(|version_name| {
+            semver::Version::parse(version_name)
+                .expect("Failed to parse hand-written string as a valid version name.")
+        });
+        let chosen_baseline = choose_baseline_version(&crate_, current_version.as_ref())
+            .expect("choose_baseline_version should not return any error");
```
tonowak, line 410: Inconsistent style with the rest of `expect`s.

src/baseline.rs
obi1kenobi: This is worth a read: https://doc.rust-lang.org/std/error/index.html#common-message-styles

Definitely lowercase and not ending with punctuation. For the rest, read the doc at the link above and then use your best judgment :)

mgr0dzicki: Changed to lowercase without ending punctuation. I used the “expect as precondition” style everywhere, hope it's finally good enough ;)

src/baseline.rs
```diff
@@ -242,6 +243,52 @@ fn create_rustdoc_manifest_for_crate_version(
...
+                anyhow::format_err!(
+                    "No available baseline versions for {}@{}",
+                    crate_.name(),
+                    current
+                )
+            })
+    } else {
+        // If there is no normal version (not yanked and not a pre-release)
+        // choosing the latest one anyway is more reasonable than throwing an
+        // error, as there is still a chance that it is what the user expects.
```
obi1kenobi, line 283: This comment isn't in the right place, and is confusing as a result. Its current placement seems to indicate that it's summarizing what's happening in the entire `else` block. But instead, the comment only applies to the `.unwrap_or_else(|| crate_.highest_version())` part.

I'd make that closure multi-line, and put the comment inside its `{` block instead. Then it's impossible to misinterpret.

mgr0dzicki: Definitely good idea, thank you!

obi1kenobi (approved):
This looks good to me! @tonowak if your concerns are also resolved, let's merge?

PR #257, opened 2023-01-05
Use trustfall_rustdoc v0.7.0 with support for rustdoc v24.

obi1kenobi:


obi1kenobi:
This needs the changes from #171, so I'll cherry-pick this commit over there instead.

Issue #269, opened 2023-01-06
Custom target directory causes generated rustdoc files to overwrite each other


tonowak:
### Steps to reproduce the bug with the above code

```
cd /tmp
git clone ssh://git@github.com:obi1kenobi/cargo-semver-checks
cd cargo-semver-checks
mkdir -p .cargo
echo '[build]\ntarget-dir = "/tmp/target"' > .cargo/config.toml
cargo run semver-checks check-release --manifest-path ./test_crates/enum_missing/new/Cargo.toml --baseline-root=./test_crates/enum_missing/old/Cargo.toml
```

### Actual Behaviour

It ends successfully without reporting any issues.

### Expected Behaviour

It should find a semver breaking issue.

#### Software version

cargo-semver-checks 0.15.2 (f6b41f4)

#### Operating system

Linux 6.1.1-arch1-1

#### Command-line

```bash
/tmp/target/debug/cargo-semver-checks semver-checks --bugreport
```

#### cargo version

```
> cargo -V
cargo 1.67.0-beta.2 (f6e737b1e 2022-12-02)
```

#### Compile time information

- Profile: debug
- Target triple: x86_64-unknown-linux-gnu
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2
- Host: x86_64-unknown-linux-gnu

### Additional Context

I think this bug happens when the binary is in any other place other than in the default path (not only by passing a custom path to the built binary).

PR #270, opened 2023-01-07
Fix rustdoc .json files overwriting each other

staniewzki:
Fixes https://github.com/obi1kenobi/cargo-semver-checks/issues/269 and #309.

When `target-dir` is specified in `.cargo/config.toml`, the path for generated rustdoc .json file is the same for the baseline and current crate version. When generating the .json files both in advance, and them processing them later, this is obviously a problem, as one of them gets overwritten.

I don't know if the solution I have came up with is the best one, as we might load too much data into memory when processing multiple crates at once. On the other hand, this way we are sure, overwritting is not an issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,7 +205,7 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
         with:
-          key: bevy-core
+          key: bevy--core
 
       - name: Run semver-checks
         run: cargo run semver-checks check-release --manifest-path="subject/crates/bevy_core/Cargo.toml"
@@ -243,6 +243,40 @@ jobs:
 
       - name: Run semver-checks
         run: cargo run semver-checks check-release --manifest-path="subject/crates/bevy_gltf/Cargo.toml" --baseline-version 0.9.0
+
+  run-on-clap:
+    # clap v3.2.0 added a semver violation
+    # check if cargo-semver-checks detects the issue
+    # https://github.com/clap-rs/clap/issues/3876
+    name: Run cargo-semver-checks on clap v3.2.0
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout cargo-semver-checks
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+
+      - name: Checkout clap
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+          repository: 'clap-rs/clap'
+          ref: 'v3.2.0'
+          path: 'subject'
+
+      - name: Install rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: clap
+
+      - name: Run semver-checks
+        run: cargo run semver-checks check-release --manifest-path="subject/Cargo.toml"
 
   init-release:
     name: Run the release workflow
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -246,7 +246,7 @@ jobs:
 
   run-on-clap:
     # clap v3.2.0 added a semver violation
-    # check if cargo-semver-checks detects the issue
+    # check whether cargo-semver-checks detects the issue
     # https://github.com/clap-rs/clap/issues/3876
     name: Run cargo-semver-checks on clap v3.2.0
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -284,7 +284,9 @@ jobs:
           key: clap
 
       - name: Run semver-checks
-        run: not cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-root="subject-baseline/Cargo.toml" > output
+        run: |
+          cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-root="subject-baseline/Cargo.toml" > output
+          if [[ $? == 0 ]]; then exit 1; else exit 0; fi
 
       - name: Check output
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,7 @@ jobs:
       - run-on-core-graphics
       - run-on-bevy-core
       - run-on-bevy-gltf
+      - run-on-clap
     steps:
       - run: exit 0
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -292,12 +292,12 @@ jobs:
 
       - name: Check return code
         run: |
-          if [[ $(cat return_code) == 0 ]]; then exit 1; else exit 0; fi
+          if [[ "$(cat return_code)" == "0" ]]; then exit 1; else exit 0; fi
 
       - name: Check output
         run: |
-          EXPECTED=$(echo -e "--- failure auto_trait_impl_removed: auto trait no longer implemented ---\n--- failure trait_missing: pub trait removed or renamed ---")
-          RESULT=$(cat output | grep failure)
+          EXPECTED="$(echo -e "--- failure auto_trait_impl_removed: auto trait no longer implemented ---\n--- failure trait_missing: pub trait removed or renamed ---")"
+          RESULT="$(cat output | grep failure)"
           diff <(echo "$RESULT") <(echo "$EXPECTED")
 
   init-release:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -284,9 +284,14 @@ jobs:
           key: clap
 
       - name: Run semver-checks
+        continue-on-error: true
         run: |
           cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-root="subject-baseline/Cargo.toml" > output
-          if [[ $? == 0 ]]; then exit 1; else exit 0; fi
+          echo "$?" > return_code
+
+      - name: Check return code
+        run: |
+          if [[ $(cat return_code) == 0 ]]; then exit 1; else exit 0; fi
 
       - name: Check output
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -256,13 +256,21 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Checkout clap
+      - name: Checkout clap v3.2.0
         uses: actions/checkout@v3
         with:
           persist-credentials: false
           repository: 'clap-rs/clap'
           ref: 'v3.2.0'
-          path: 'subject'
+          path: 'subject-current'
+
+      - name: Checkout clap v3.1.18
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+          repository: 'clap-rs/clap'
+          ref: 'v3.1.18'
+          path: 'subject-baseline'
 
       - name: Install rust
         uses: actions-rs/toolchain@v1
@@ -276,7 +284,14 @@ jobs:
           key: clap
 
       - name: Run semver-checks
-        run: cargo run semver-checks check-release --manifest-path="subject/Cargo.toml"
+        run: ! cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-root="subject-baseline/Cargo.toml" > output
+
+      - name: Check output
+        run: |
+          EXPECTED=$(echo -e "--- failure auto_trait_impl_removed: auto trait no longer implemented ---\n--- failure trait_missing: pub trait removed or renamed ---")
+          RESULT=$(cat output | grep failure)
+          diff <(echo "$RESULT") <(echo "$EXPECTED")
+          [[ $RESULT == $EXPECTED ]]
 
   init-release:
     name: Run the release workflow
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,7 +205,7 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
         with:
-          key: bevy--core
+          key: bevy-core
 
       - name: Run semver-checks
         run: cargo run semver-checks check-release --manifest-path="subject/crates/bevy_core/Cargo.toml"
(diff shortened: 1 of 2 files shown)


obi1kenobi:
Nice catch!

Would it be possible to reorganize the code to do this instead:
- for each crate that needs to be checked:
  - generate the current rustdoc, then load it
  - generate the baseline, then load it
  - check that crate
  - move on to the next one

This way, we both don't have to worry about the files colliding or disappearing, and also we give the user feedback sooner on whether their crate passed or failed. I've recently been running cargo-semver-checks on some pretty large workspaces (dozens of crates), and I didn't love that I had to wait for all the crates' rustdocs to be generated (half an hour or more!) before any crate got checked.

tonowak:
> for each crate that needs to be checked: generate the current rustdoc, then load it, generate the baseline, then load it, check that crate, move on to the next one

Done.

src/main.rs
```diff
@@ -122,70 +122,67 @@ fn main() -> anyhow::Result<()> {
...
-                        let baseline_crate = load_rustdoc(&baseline_path)?;
-
-                        prepared_crates.push((crate_name.clone(), baseline_crate, current_crate));
-                    }
-                    prepared_crates
-                };
             let mut success = true;
-            for (crate_name, baseline_crate, current_crate) in prepared_crates {
-                if !run_check_release(&mut config, &crate_name, current_crate, baseline_crate)? {
+            let mut run_check_release = |config: &mut GlobalConfig, crate_name: &str, current_crate: VersionedCrate, baseline_crate: VersionedCrate| -> anyhow::Result<()> {
```
obi1kenobi: I mentioned this in #273, but I'm really not a fan of the "nontrivial inline closure that mutates external state" approach here.

Please reconsider the control flow here and try to make this code easier to read and understand.

src/main.rs
```diff
@@ -155,31 +172,26 @@ fn main() -> anyhow::Result<()> {
                         "Parsing",
                         format_args!("{crate_name} v{version} (current)"),
                     )?;
+
                     let rustdoc_path = rustdoc_cmd.dump(manifest_path, None, true)?;
+                    let current_crate = load_rustdoc(&rustdoc_path)?;
+
+                    // Can overwrite rustdoc of current crate.
+                    // For example, this happens when target-dir is specified in `.cargo/config.toml`.
+                    // That's the reason why we're immediately loading the rustdocs to memory.
```
obi1kenobi, line 181:
```suggestion
                    // The process of generating baseline rustdoc can overwrite 
                    // the already-generated rustdoc of the current crate.
                    // For example, this happens when target-dir is specified in `.cargo/config.toml`.
                    // That's the reason why we're immediately loading the rustdocs into memory.
                    // See: https://github.com/obi1kenobi/cargo-semver-checks/issues/269
```
More detail: what exactly causes the overwrite + link to the issue in question.

src/main.rs
```diff
@@ -155,31 +172,26 @@ fn main() -> anyhow::Result<()> {
...
+                    run_check_release(&mut config, crate_name, current_crate, baseline_crate)?;
                 }
             }
-            if success {
-                std::process::exit(0);
-            } else {
-                std::process::exit(1);
-            }
+
+            std::process::exit(i32::from(!success));
```
obi1kenobi, line 194: The new code is more clever, but much less readable and much harder to understand. The cleverness is hurting more than helping here. This is not a good tradeoff.

obi1kenobi (commented):
Thanks to both of you for putting this together! Just some stylistic comments to help readability.

src/main.rs
tonowak: I agree. But it's not really my fault.

The refactor I had in mind is `std::process::exit(if success { 0 } else { 1 });` which, in my opinion, is a bit more readable than the original code -- it makes it more clear that what is important is exiting with some error code, not branching through the possible states of `success`.

But `clippy` has suggested changing `if success { 0 } else { 1 }` to `i32::from(!success)` and I'm also not convinced it is more readable. Assuming we don't want to ignore the clippy lints, the only options are the old code and the code I proposed in this PR.

What do you think: should I somehow ignore the clippy lint and do the inlined `if`, or should I revert to the old code?

src/main.rs
tonowak: I'll move this conversation to zulip.

src/main.rs
obi1kenobi: I'd probably leave the old code as-is. Duplicating the `std::process::exit()` call feels like "the lesser evil" here, from a risk and readability perspective.

tonowak: Done in 5e811c3cf1c4076c49994b806421561739ecd9cf.

src/main.rs
tonowak: Done in 5e811c3cf1c4076c49994b806421561739ecd9cf.

src/main.rs
```diff
@@ -122,59 +123,74 @@ fn main() -> anyhow::Result<()> {
...
                 let name = "<unknown>";
                 let version = None;
-                vec![(
-                    name.to_owned(),
-                    loader.load_rustdoc(&mut config, &rustdoc_cmd, name, version)?,
-                    current_rustdoc_path.to_owned(),
-                )]
+                let baseline_path =
+                    loader.load_rustdoc(&mut config, &rustdoc_cmd, name, version)?;
+                let baseline_crate = load_rustdoc(&baseline_path)?;
```
obi1kenobi, line 135: Would it be possible to extract a function that generates the `(current_crate, baseline_crate)` tuple?

Asking because this code here can also be affected by the same JSON file overwriting bug as the code below near L167-178. That other code block has the big comment, which is great, but this part here doesn't. Instead of duplicating the comment, perhaps we can extract the logic into a function and put the big comment block on that function?

obi1kenobi (commented):
The fix looks good overall. Just a small refactoring idea to look into to make it even better.

Is there a test case we could reasonably add for this? Perhaps an integration-style test on the `clap` release with the `UnwindSafe` semver issue, with a custom target dir, requiring that the semver issue be found and reported? (I.e. not just "exit 1" but also a `grep` in the output for the failing lint and the name of the affected type.)

tonowak:
> The fix looks good overall. Just a small refactoring idea to look into to make it even better.
> 
> Is there a test case we could reasonably add for this? Perhaps an integration-style test on the `clap` release with the `UnwindSafe` semver issue, with a custom target dir, requiring that the semver issue be found and reported? (I.e. not just "exit 1" but also a `grep` in the output for the failing lint and the name of the affected type.)

Sounds good. From what I remember, the semver break is exactly between clap's v3.1.18 and v3.2.0.

src/main.rs
```diff
@@ -187,6 +201,40 @@ fn main() -> anyhow::Result<()> {
     }
 }
 
+// Argument to the generate_versioned_crates function.
+// It can be either a path to an existing rustdoc, or a manifest path of a crate.
```
tonowak, line 205: I'm not sure whether the comments here are helpful nor needed -- sure, it is useful in knowing that this enum is only used in this function, but the name and values of this enum are more than enough to understand what it contains and to understand the code (similarly, the second line in the comment is just a humanly phrased code that is short enough to understand without any problems).

tonowak (commented):
Nice, I like those changes -- now it would be hard to make a similar bug.

.github/workflows/ci.yml
```diff
@@ -205,7 +205,7 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
         with:
-          key: bevy-core
+          key: bevy--core
```
tonowak, line 208: Is this change intentional?

.github/workflows/ci.yml
```diff
@@ -244,6 +244,40 @@ jobs:
       - name: Run semver-checks
         run: cargo run semver-checks check-release --manifest-path="subject/crates/bevy_gltf/Cargo.toml" --baseline-version 0.9.0
 
+  run-on-clap:
+    # clap v3.2.0 added a semver violation
+    # check if cargo-semver-checks detects the issue
```
tonowak, line 249: Maybe "check whether" instead of "check if"?

.github/workflows/ci.yml
```diff
@@ -244,6 +244,40 @@ jobs:
...
+          toolchain: stable
+          profile: minimal
+          override: true
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: clap
+
+      - name: Run semver-checks
+        run: cargo run semver-checks check-release --manifest-path="subject/Cargo.toml"
```
tonowak, line 279: > (I.e. not just "exit 1" but also a grep in the output for the failing lint and the name of the affected type.)

I just want to be sure -- you still plan to add the following logic mentioned by @obi1kenobi, yes?

.github/workflows/ci.yml
```diff
@@ -244,6 +244,55 @@ jobs:
...
+        with:
+          key: clap
+
+      - name: Run semver-checks
+        run: ! cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-root="subject-baseline/Cargo.toml" > output
+
+      - name: Check output
+        run: |
+          EXPECTED=$(echo -e "--- failure auto_trait_impl_removed: auto trait no longer implemented ---\n--- failure trait_missing: pub trait removed or renamed ---")
+          RESULT=$(cat output | grep failure)
```
tonowak, line 292: Maybe let's also check that no other failures have been found?

.github/workflows/ci.yml
tonowak: Nevermind, I see that's now implemented in 82a0bb3b661b4aab1255206b84b53e2bb41dda08.

.github/workflows/ci.yml
staniewzki: What do you mean? I check if the failures are the exact ones that are expected.

tonowak: Oh, I see. Thanks for explaining.

.github/workflows/ci.yml
```diff
@@ -244,6 +244,62 @@ jobs:
...
+      - name: Check return code
+        run: |
+          if [[ $(cat return_code) == 0 ]]; then exit 1; else exit 0; fi
+
+      - name: Check output
+        run: |
+          EXPECTED=$(echo -e "--- failure auto_trait_impl_removed: auto trait no longer implemented ---\n--- failure trait_missing: pub trait removed or renamed ---")
+          RESULT=$(cat output | grep failure)
+          diff <(echo "$RESULT") <(echo "$EXPECTED")
+          [[ $RESULT == $EXPECTED ]]
```
tonowak, line 301: I think `diff` returns 1 when files differ, so maybe it would be enough to just rely on the `diff`, without doing the `[[ $RESULT == $EXPECTED ]]`?

src/main.rs
```diff
@@ -187,6 +201,39 @@ fn main() -> anyhow::Result<()> {
     }
 }
 
+// Argument to the generate_versioned_crates function.
+enum CurrentCratePath<'a> {
+    CurrentRustdocPath(&'a Path), // If rustdoc is passed, it is just loaded into the memory.
+    ManifestPath(&'a Path),       // Otherwise, the function generates the rustdoc.
```
tonowak, line 207: `the function` -- which function? Reading the file from top to bottom, it's not clear that it mentions the `generate_versioned_crates` function.

staniewzki: How so? It is mentioned in the comment above

tonowak: Oh, I missed this comment. Thanks.

.github/workflows/ci.yml
staniewzki: Uh, I used the diff for the debug, but yeah, it should work

tonowak (approved):

.github/workflows/ci.yml
```diff
@@ -244,6 +244,61 @@ jobs:
...
+
+      - name: Run semver-checks
+        continue-on-error: true
+        run: |
+          cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-root="subject-baseline/Cargo.toml" > output
+          echo "$?" > return_code
+
+      - name: Check return code
+        run: |
+          if [[ $(cat return_code) == 0 ]]; then exit 1; else exit 0; fi
```
obi1kenobi, line 294: Unlike Rust macros, bash expansions are not hygienic and can expand in unpredictable ways. In the interest of avoiding painful future debugging (and possible security vulnerabilities! imagine what happens if the contents of that block are attacker-controllable!), it's good to get in the habit of explicitly wrapping any `$( <stuff> )` in quotes, even when not strictly necessary. The slight annoyance of converting a number into a string is much, much preferable to a few painful hours of debugging, or the days or weeks that triaging and containing a security problem might cause.
```suggestion
          if [[ "$(cat return_code)" == "0" ]]; then exit 1; else exit 0; fi
```

.github/workflows/ci.yml
```diff
@@ -244,6 +244,61 @@ jobs:
...
+
+      - name: Check output
+        run: |
+          EXPECTED=$(echo -e "--- failure auto_trait_impl_removed: auto trait no longer implemented ---\n--- failure trait_missing: pub trait removed or renamed ---")
+          RESULT=$(cat output | grep failure)
```
obi1kenobi, line 299: Same as above.
```suggestion
          EXPECTED="$(echo -e "--- failure auto_trait_impl_removed: auto trait no longer implemented ---\n--- failure trait_missing: pub trait removed or renamed ---")"
          RESULT="$(cat output | grep failure)"
```

obi1kenobi (approved):
This looks great, well done @staniewzki and @tonowak!

obi1kenobi (approved):

.github/workflows/ci.yml
```diff
@@ -292,12 +292,12 @@ jobs:
...
       - name: Check return code
         run: |
-          if [[ $(cat return_code) == 0 ]]; then exit 1; else exit 0; fi
+          if [[ "$(cat return_code)" == "0" ]]; then exit 1; else exit 0; fi
 
       - name: Check output
         run: |
-          EXPECTED=$(echo -e "--- failure auto_trait_impl_removed: auto trait no longer implemented ---\n--- failure trait_missing: pub trait removed or renamed ---")
-          RESULT=$(cat output | grep failure)
+          EXPECTED="$(echo -e "--- failure auto_trait_impl_removed: auto trait no longer implemented ---\n--- failure trait_missing: pub trait removed or renamed ---")"
```
tonowak, line 299: Do the nested quotes work in bash?

obi1kenobi: They do! I know I asked the same question when my own mentor taught me to quote everything like this :)

PR #271, opened 2023-01-08
Add a field to the bug report form for config.toml information.

obi1kenobi:
This would have made #269 easier to find.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,6 +1,6 @@
 name: "Bug report"
 description: "Things didn't meet expectations"
-labels: 'C-bug'
+labels: ['C-bug']
 body:
   - type: textarea
     attributes:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -26,6 +26,10 @@ body:
       description: "Run `cargo semver-checks --bugreport` and copy-paste the output here."
     validations:
       required: true
+  - type: textarea
+    attributes:
+      label: "Build Configuration"
+      description: "Any local Rust build configuration in config.toml files? Paste the file's content here."
```
tonowak, line 32: Maybe also add a link explaining what that means? https://doc.rust-lang.org/cargo/reference/config.html

tonowak, line 32: Maybe "Paste the file's content here if you have one", to make it clear that not having one is normal (and possibly expected).

tonowak (approved):

PR #273, opened 2023-01-08
Add flag to get current rustdoc from registry

tonowak:
Todo list:
- [x] generalize `src/baseline.rs` to allow generating the rustdoc with different ways for both current and baseline. Do that by renaming the file `src/baseline.rs` and variables in the code (the filename and variables suggest that it is a code only for the baseline, but this PR will use this code also for the current).
- [ ] add the functionality in `src/main.rs` to generate current rustdoc from registry. A possible blocker is the `--workspace` flag.
- [ ] add the flag (what should be its name?).

This PR is stacked on top of https://github.com/obi1kenobi/cargo-semver-checks/pull/270.
It allows https://github.com/obi1kenobi/cargo-semver-checks/pull/207 to be shorter.

### DIFF
--- a/src/rustdoc_generation.rs
+++ b/src/rustdoc_generation.rs
@@ -7,44 +7,50 @@ use crate::dump::RustDocCommand;
 use crate::util::slugify;
 use crate::GlobalConfig;
 
-pub(crate) trait BaselineLoader {
-    fn load_rustdoc(
+pub(crate) trait RustdocGenerator {
+    fn generate_rustdoc(
         &self,
         config: &mut GlobalConfig,
-        rustdoc: &RustDocCommand,
+        rustdoc_command: &RustDocCommand,
         name: &str,
-        version_current: Option<&semver::Version>,
+        highest_considered_registry_version: Option<&semver::Version>,
+        crate_type: &str,
     ) -> anyhow::Result<std::path::PathBuf>;
 }
 
-pub(crate) struct RustdocBaseline {
+pub(crate) struct FromRustdoc {
     path: std::path::PathBuf,
 }
 
-impl RustdocBaseline {
+impl FromRustdoc {
     pub(crate) fn new(path: std::path::PathBuf) -> Self {
         Self { path }
     }
 }
 
-impl BaselineLoader for RustdocBaseline {
-    fn load_rustdoc(
+impl RustdocGenerator for FromRustdoc {
+    fn generate_rustdoc(
         &self,
-        _config: &mut GlobalConfig,
-        _rustdoc: &RustDocCommand,
-        _name: &str,
-        _version_current: Option<&semver::Version>,
+        config: &mut GlobalConfig,
+        _rustdoc_command: &RustDocCommand,
+        name: &str,
+        _highest_considered_registry_version: Option<&semver::Version>,
+        crate_type: &str,
     ) -> anyhow::Result<std::path::PathBuf> {
+        config.shell_status(
+            "Using",
+            format_args!("{name}: {} ({crate_type})", self.path.display()),
+        )?;
         Ok(self.path.clone())
     }
 }
 
-pub(crate) struct PathBaseline {
+pub(crate) struct FromPath {
     root: std::path::PathBuf,
     lookup: std::collections::HashMap<String, (String, std::path::PathBuf)>,
 }
 
-impl PathBaseline {
+impl FromPath {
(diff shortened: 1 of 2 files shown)


tonowak:
I'm not sure I like all the renames in 708c01626af4d27c5077558592aac28b2510f707. For example, the `rustdoc_generation.rs` name makes sense, but in truth, the rustdoc is generated in `dump.rs` (which, by the way, isn't a descriptive name). Nevertheless, I think the new names make more sense than previously and they help in understanding what exactly the code does.

src/rustdoc_generation.rs
```diff
@@ -7,44 +7,50 @@ use crate::dump::RustDocCommand;
...
+pub(crate) trait RustdocGenerator {
+    fn generate_rustdoc(
         &self,
         config: &mut GlobalConfig,
-        rustdoc: &RustDocCommand,
+        rustdoc_command: &RustDocCommand,
         name: &str,
-        version_current: Option<&semver::Version>,
+        highest_considered_registry_version: Option<&semver::Version>,
+        crate_type: &str,
```
obi1kenobi, line 17: At this point, I feel like this trait has gotten complex enough that it needs documentation on what all its arguments mean. It's not obvious to me what `crate_type` is supposed to be, and `highest_considered_registry_version` has a specificity to it that makes me afraid that it might be more complex than it might seem at first glance.

src/rustdoc_generation.rs
```diff
@@ -7,44 +7,50 @@ use crate::dump::RustDocCommand;
...
     ) -> anyhow::Result<std::path::PathBuf>;
 }
 
-pub(crate) struct RustdocBaseline {
+pub(crate) struct FromRustdoc {
```
obi1kenobi, line 21: I'm overall not a fan of the `From` prefixes in all these names. While `rustdoc_generation::FromRustdoc` etc. seems pretty obvious, there's zero guarantee that it will always be used like that and it seems brittle from an understandability perspective to assume it. For example, `Debug` would only print the name of the type itself.

Consider the `RustdocFrom*` prefix? Then this struct could be `RustdocFromFile` etc.

On a separate note, I'm astonished to see that these structs aren't `#[derive(Debug)]`. Please add that throughout?

src/rustdoc_generation.rs
```diff
@@ -271,14 +284,14 @@ fn create_rustdoc_manifest_for_crate_version(
     }
 }
 
-fn choose_baseline_version(
+fn choose_version_from_registry(
     crate_: &Crate,
-    version_current: Option<&semver::Version>,
+    highest_considered_registry_version: Option<&semver::Version>,
```
obi1kenobi, line 289: This strongly makes me feel like we're trying to force the wrong abstraction here. `highest_considered_registry_version` is only relevant for baselines, and specifically for registry-based baselines, and yet we're carrying it throughout the entire API including in the trait etc.

Perhaps take a step back from this and consider another approach? Maybe one where the baseline generation portion is aware it's generating a baseline (and can show appropriate status and error messages for it!), and a lower-level abstraction that just generates a rustdoc JSON file based on the information it's given?

obi1kenobi (commented):
I feel like this is trying to force the wrong abstraction through. This new API feels overly complex and not particularly elegant. I suggested an alternative approach in one of the comments.

tonowak:
I'm thinking about closing this PR and adding the functionality I had in mind through this PR to the https://github.com/obi1kenobi/cargo-semver-checks/pull/207. That'll create a clear line between `cargo-semver-checks` --a tool to analyze own's code, and https://github.com/obi1kenobi/cargo-semver-checks/pull/207 -- a tool to analyze results for chosen crates and versions (which will be used e.g. to test for false-positives of cargo-semver-checks in the CI).

What do you think about it?

Assuming you agree with closing this PR, should I nevertheless refactor `src/baseline.rs` in any way, or should I disregard all changes?

obi1kenobi:
Assuming @epage is okay with my suggestion in https://github.com/obi1kenobi/cargo-semver-checks/issues/86#issuecomment-1376035640, I think we'll still need to extract the code for generating rustdoc based on a registry crate's name and version into something the new tool can also call.

It's likely that those changes would look a bit different than the proposed refactor in this PR.

I'd recommend closing this draft PR, opening a new PR for those changes by themselves, and keeping them separate from #207 just in the interest of keeping the components small and easy to review. Ideally, #207 shouldn't have any effect on the `semver-checks` tool's portion of the code — that would make it very easy to read and merge.

PR #278, opened 2023-01-10
New lint: Enum `#[must_use]` added

SmolSir:
This is a part of solving #159, as well as splitting #268 into more manageable, smaller PRs.

Implements the check against adding `#[must_use]` attribute to a public Enum.

### DIFF
--- a/src/lints/enum_must_use_added.ron
+++ b/src/lints/enum_must_use_added.ron
@@ -59,6 +59,6 @@ SemverQuery(
         "must_use": "must_use",
         "zero": 0,
     },
-    error_message: "An enum has been marked with #[must_use]. This can cause downstream crates that did not use this enum's value to get a compiler lint.",
+    error_message: "An enum is now #[must_use]. Downstream crates that did not use its value will get a compiler lint.",
     per_result_error_template: Some("enum {{name}} in {{span_filename}}:{{span_begin_line}}"),
 )
--- a/src/lints/enum_must_use_added.ron
+++ b/src/lints/enum_must_use_added.ron
@@ -1,0 +1,64 @@
+SemverQuery(
+    id: "enum_must_use_added",
+    human_readable_name: "enum #[must_use] added",
+    description: "An enum has been marked with #[must_use].",
+    required_update: Minor,
+
+    // TODO: Change the reference link to point to the cargo semver reference
+    //       once it has a section on attribute #[must_use].
+    reference_link: Some("https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute"),
+    query: r#"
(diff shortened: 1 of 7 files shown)


src/lints/enum_must_use_added.ron
```diff
@@ -0,0 +1,64 @@
...
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "must_use": "must_use",
+        "zero": 0,
+    },
+    error_message: "An enum has been marked with #[must_use]. This can cause downstream crates that did not use this enum's value to get a compiler lint.",
```
obi1kenobi, line 62: Let's try to keep these error messages as concise as possible, while still mentioning all the necessary information. Here's a suggestion:
```suggestion
    error_message: "An enum is now #[must_use]. Downstream crates that did not use its value will get a compiler lint.",
```

test_crates/enum_must_use_added/new/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+// the attribute, including deletion, should NOT be reported by this rule.
+
+pub enum MustUseEnumToEnum {
+    Bar,
+}
+
+#[must_use = "Foo"]
+pub enum MustUseEnumToMustUseMessageEnum {
+    Bar,
+}
```
obi1kenobi, line 25: Nice one!

test_crates/enum_must_use_added/new/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+
+// This enum is private and should NOT be reported by this rule.
+
+#[must_use]
+enum MustUsePrivateEnum {
+    Bar,
+}
+
+
+// This enum was added in the new version of the crate with it's attribute.
```
obi1kenobi, line 55: typo:
```suggestion
// This enum was added in the new version of the crate with its attribute.
```

test_crates/enum_must_use_added/new/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+// It should NOT be reported by this rule because adding a new enum is not
+// a breaking change.
```
obi1kenobi, line 57: This rule isn't semver-major so it doesn't look for breaking changes, and I was confused by this comment.
```suggestion
// It should NOT be reported by this rule to avoid duplicate lints.
// It should be reported as a new pub type that is part of the crate's API.
```

obi1kenobi (commented):
Looks good overall! Just a few tiny suggested tweaks.

Did you mean to leave this as a draft?

obi1kenobi (approved):

Issue #241, opened 2022-12-20
Tracking issue: Required schema additions to make new lints possible


obi1kenobi:
In approximate order of increasing difficulty:

- [ ] expose auto-trait / unsafe trait info on [`type Trait`](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/0b1f021c22aab6e48f38156e52c7b8331447149a/src/rustdoc_schema.graphql#L382)
    - new lints enabled by this:
        - [trait became unsafe](https://github.com/obi1kenobi/cargo-semver-checks/issues/231) (so its `impl` blocks need to be `unsafe impl` now)

- [ ] expose constants: [`const` and `static` values](https://doc.rust-lang.org/stable/rust-by-example/custom_types/constants.html)
    - even without type information, there are lints to be checked here
    - new lints enabled by this:
        - module-level `const` missing (cannot be imported, so renamed or deleted)
        - module-level `static` missing (cannot be imported, so renamed or deleted)
        - `const` in type's `impl` block missing (cannot be imported, so renamed or deleted)
        - `static` in type's `impl` block missing (cannot be imported, so renamed or deleted)

- [ ] expose associated type info on [`type Trait`](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/0b1f021c22aab6e48f38156e52c7b8331447149a/src/rustdoc_schema.graphql#L382)
    - even without type information, we can still write lints about this
    - new lints enabled by this:
        - associated type missing (renamed or deleted, will break `Trait::Assoc` uses)

- [ ] expose enum variant fields on [`interface Variant`](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/0b1f021c22aab6e48f38156e52c7b8331447149a/src/rustdoc_schema.graphql#L192)
    - they are already available on `type StructVariant`
    - the edge on `type PlainVariant` should exist but never produce any vertices
    - the edge on `type TupleVariant` should produce fields similarly to how tuple structs produce their fields
    - new lints enabled by this:
        - tuple variant field removed
        - exhaustive tuple variant field added

- [ ] expose field ordering information
    - probably via a separate edge to a new type, like `field_index: [FieldIndex]` where that new type contains the field index number and points to a `StructField`
    - new lints enabled by this:
        - `repr(C)` plain struct has its fields reordered

[...]

Issue #149, opened 2022-10-15
Represent type information in the Trustfall schema


obi1kenobi:
Currently, the schema that lints query does not represent type information. This means that lints like "changed the type of a pub field" can't be written since we can't compare the before-and-after types of fields, function parameters, etc.

The current schema cargo-semver-checks is defined here: https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/main/src/rustdoc_schema.graphql

Any schema design needs to consider (and provide examples for):
- How we'd represent generic types (`Vec<usize>`), lifetime information (`&'a u64`), tuples, `dyn Trait`, etc.
- What queries that compare types would look like.
- How to ensure there are no false-positives in semver lints, while also keeping the schema as simple as possible.

kornelski:
Rust's type system is [turing complete](https://sdleffler.github.io/RustTypeSystemTuringComplete/), so it may be infeasible to represent Rust types with a full fidelity. 

I suggest supporting a wildcard placeholder for any type or part of a type (like `_` in Rust). When constructing the types, if some part of it is too complex or isn't precise enough in rustdoc's output, then replace it with `_`. When checking if two types/lifetimes/generics are compatible, treat `_` as always compatible. Make lints skip types with `_` if necessary.

This would enable starting small, and incrementally expand support for more sophisticated types. Having some lints for types is better than having none at all.

For example, at minimum you could support primitive non-borrowed types, and replace everything else with `_`. Then you may build it up to support `Vec<_>`, `&'_ str`, and so on.

obi1kenobi:
We're pursuing an attempt at a more complete solution as part of Google Summer of Code, Turing completeness and all.

I'll write more about it on my blog at some point, but the gist is that we're using the only stable API the Rust compiler has: "Can you please compile this Rust file for me?" The idea is related to the [witness programs](https://predr.ag/blog/semver-violations-are-common-better-tooling-is-the-answer/#automated-validation-via-witnesses) we used to demonstrate SemVer breakage in the study from a couple of years ago.

kornelski:
Would it be worth having type information in the schema anyway?

It might be useful to exclude obviously matching functions from compilation of witnesses.

obi1kenobi:
That's the plan! We'll only attempt a witness-based check if it looks like
something might have changed.

That initial check will be shallow and can have false positives but isn't
allowed false negatives. The former means we generate a witness that we
[...]

PR #283, opened 2023-01-10
New lint: Inherent method `#[must_use]` added

SmolSir:
After some testing it turned out it is not possible to have a single lint for both `Struct` and `Trait` methods. Here is the reason:
- assume we want to support both scenarios, Struct inherent method becomes `#[must_use]` and Trait method becomes `#[must_use]` at the same time. The former method has `public` visibility, the latter has `default`, while both Struct and Trait must have `public` visibility.
- in a single query, we have to give exact parameters to both `baseline` and `current` parts of the query. So we opt for method visibility to be either `public` or `default`, because otherwise we will miss some Struct or Trait methods.
- now imagine that we have one `pub Struct Foo` and one `pub Trait Bar`, that both have a method that becomes `#[must_use]`. If we run the query that looks for `public` or `default` methods in both `baseline` and `current` crates, the query will find matches for all 4 possible pairings, despite there being only 2 of them sensible to the user. To avoid this, the query would have to "on Struct look for a matching Struct, then for the matching method, on Trait look for a matching Trait, then for the matching method", which even if was possible would be begging to get split into 2 separate queries (at least I see it this way).
- because of this we have to precise our query to a single visibility value for either `baseline` or `current` version.

And since we decided in #268 that there is no need for a lint that detects addition of `#[must_use]` for a method that is also moved to Trait, we eliminated the only cross-type case from the 3 possibilities I introduced in that PR and are left with:
- `Struct method -> Struct #[must_use] method`, with the lint `inherent_method_must_use_added` implemented in this PR,
- `Trait method -> Trait #[must_use] method`, with the lint `trait_method_must_use_added`, on which I am still working in #284.


obi1kenobi:
I think this implementation has an interesting bug.

The PR description describes this lint as `Struct method -> Struct #[must_use] method`, but that is actually too narrow. Enums and unions can also have inherent `impl` blocks whose methods may be marked with `#[must_use]`. The lint [correctly anticipates this](https://github.com/obi1kenobi/cargo-semver-checks/pull/283/files#diff-3ea0ab91a721d62936d9327b5cc08c2d38cd37d257153de37389aba3dcf71474R15) and uses the coercion `... on ImplOwner` instead of `... on Struct`. This means it will correctly flag enum methods (and, when implemented, union methods as well).

Except, the `current` and `baseline` branches of the query never ensure that the kind of item matches between the baseline and the current code. So I predict that the following code will cause a false-positive here:

baseline:
```rust
pub struct Foo {};

impl Foo {
    pub fn bar() {}
}
```

current:
```rust
pub enum Foo {
    A,
}

impl Foo {
    #[must_use]
    pub fn bar() {}
}
```

I predict that the current code will flag `Foo::bar` as having gained a `#[must_use]` attribute, when in fact that's a completely new method on a completely new type.

It would be great to add "true-positive" test cases for enum and union methods just like the existing ones for structs, even though the union ones won't be caught (opne an issue about this and link it in a comment, so we remember that it's okay to for that test output to change when we implement unions in the adapter).

It would also be good to have "true-negative" test cases where structs/enums/unions get each other's names like this.

With respect to fixing the lint query, there are two options:
- Each vertex has a special implicit field `__typename: String!` that contains the exact type of the vertex: on `ImplOwner` it would be `Struct` or `Enum` or `Union` (when implemented). You could tag and filter on that value across the baseline / current branches of the query to make sure the item type matches up.
- Alternatively, if you think there's an advantage to doing so, you could split the lint into separate lints for `Struct / Enum / Union` types and replace `... on ImplOwner` with `... on Struct` etc. This doesn't seem to offer any obvious advantages over the alternative, and is more complex, but I wanted to bring it up for your consideration in case I missed something.

SmolSir:
That's a very interesting bug, I wasn't aware of the `inherent_impl` existing for types other than `Struct`. At first glance I also think it will be best to just broaden the query to handle both `Struct` and `Enum`, `Union` cases, but I will look deeper into this first before proceeding with changes.

> Alternatively, if you think there's an advantage to doing so, you could split the lint into separate lints for Struct / Enum / Union types and replace ... on ImplOwner with ... on Struct etc. This doesn't seem to offer any obvious advantages over the alternative, and is more complex, but I wanted to bring it up for your consideration in case I missed something.

The only "advantage" I can point out for now is that this would result in multiple smaller test crates, one per lint, instead of one big test crate handling all of the cases for one, common lint. But if the test crate turns out too large, we will be able to easily split it into smaller test cases and leave the common lint unchanged (I will keep it in mind to write the tests as easily splittable because of this).

SmolSir:
> I will keep it in mind to write the tests as easily splittable because of this.

The best option was to use the hybrid approach mentioned here https://github.com/obi1kenobi/cargo-semver-checks/pull/282#issuecomment-1382352356, so I already split the tests into relevant-named files. I also added a "test file pair" against the changed type false-positive scenario which was shown above.

**This is currently blocked on** #149 & #292.

If the lint was split into more specialized ones for `Enum`, `Struct`& `Union`, this would not require implementing #149 and would allow for adding `Enum` & `Struct` lints quickly, while the `Union` lint would wait until #292 is complete (and could be mentioned in the issue itself to be added once `Union` is in the schema, among some other lints that it will enable).

@obi1kenobi considering the above, despite it requiring a bit more work with splitting the lint, I think it could in return make the workflow a lot smoother and faster (with #149 being rated as the hardest addition in #241, I am not sure if this is a good idea to block this PR on that issue). If you have nothing against it, I would proceed with this approach to thin out the #[must_use] issue as much as possible.

obi1kenobi:
I don't think I see why this is blocked on #149. That issue is required for representing the type of fields, function arguments, generics, etc. I'm confused why those are involved here.

Are we not able to use [the `ImplOwner` type's `inherent_impl` edge](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/rustdoc-v24/src/rustdoc_schema.graphql#L336)? The `__typename` special field on `ImplOwner` will be one of `"Struct" / "Enum" / "Union"` so it seems like we should be able to use the standard `@tag + @filter` combination across `current` and `baseline` to determine when the types are of the same kind (struct / enum / union).

src/lints/inherent_method_must_use_added.ron
```diff
@@ -25,42 +26,43 @@ SemverQuery(
...
                 item {
                     ... on ImplOwner {
                         visibility_limit @filter(op: "=", value: ["$public"])
                         name @filter(op: "=", value: ["%name"])
+                        __typename @filter(op: "=", value: ["%owner_type"])
 
                         importable_path {
                             path @filter(op: "=", value: ["%path"])
-                        }
+                        }
```
obi1kenobi, line 54:
```suggestion
                        }
```
You should be able to configure your editor to strip trailing whitespace like this automatically. It's nice so that other people opening the file don't end up accidentally editing it when _their_ editor strips the whitespace introduced by someone else.

SmolSir:
It is indeed not blocked, my reading has just fooled me on this bit
> Each vertex has a special implicit field __typename: String! that contains the exact type of the vertex: on ImplOwner it would be Struct or Enum or Union (when implemented).

I applied the "when implemented" part to the whole sentence, that is `__typename` and assumed it was a part of #149, while it of course was just about the `Union`...
With that problem out of the way, this should be working now! I also updated the issue for adding `Union` #292 to remind that once it is released, the output file will become invalid.

src/lints/inherent_method_must_use_added.ron
SmolSir: I applied the change and also added source control recognizing such changes instead of ignoring them. Sorry for making you correct such mistakes, this should not happen again now.

src/lints/inherent_method_must_use_added.ron
```diff
@@ -0,0 +1,80 @@
...
+    reference_link: Some("https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute"),
+    query: r#"
+    {
+        CrateDiff {
+            current {
+                item {
+                    ... on ImplOwner {
+                        visibility_limit @filter(op: "=", value: ["$public"]) @output
+                        name @tag @output
+                        owner_type: __typename @tag
```
obi1kenobi, line 18:
```suggestion
                        owner_type: __typename @tag @output
```
Just for the sake of easier debugging, if we ever need it.

Trustfall guarantees that `@output` directives are processed only after all filtering / type coercions etc. is done, so adding `@output` has zero performance cost on results that aren't actually returned. Any results that are returned are lints, so there are almost certainly not even remotely enough of them to make any noticeable performance difference whatsoever.

So it's probably a good idea just for our debugging convenience to always `@output` values that are used in with a `@tag`.

src/lints/inherent_method_must_use_added.ron
```diff
@@ -0,0 +1,80 @@
...
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "must_use": "must_use",
+        "zero": 0,
+    },
+    error_message: "An inherent method is now #[must_use]. Downstream crates that did not use its return value will get a compiler lint.",
+    per_result_error_template: Some("inherent method {{join \"::\" path}}::{{method_name}} in {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 79:
```suggestion
    per_result_error_template: Some("method {{join \"::\" path}}::{{method_name}} in {{span_filename}}:{{span_begin_line}}"),
```
I think we can afford to omit `inherent` here, in the interest of keeping this message shorter:
- It makes the span info appear earlier, and following the span info will get the reader to the right method regardless of whether they've remembered that the method is inherent or not.
- The lint name and error message already contain "inherent" in the name, so the odds are good the user has noticed at least one of them anyway.

obi1kenobi (approved):
Looks great! I'll apply my two tiny suggestions and merge.

Issue #5, opened 2022-07-18
Tracking issue: Useful non-semver warnings 


obi1kenobi:
There are certain changes that are not breaking and don't even require a minor version, but can still be frustrating in downstream crates without a minor or major version bump.

For example, based on [this poll](https://twitter.com/PredragGruevski/status/1548327174360354818) (with small sample size: ~40 respondents), ~40% of users expect that upgrading to a new patch version of a crate should not generate new lints or compiler warnings. The split between expecting a new minor version and a new major version was approximately 3-to-1.

This issue tracks a list of such warnings that would be useful to have. Crates should be able to opt into these warnings on an individual basis.

- [ ] Adding `#[must_use]` on an item.
  - Code in downstream crates that did not use a value of that type will get a compiler lint.
- [ ] Removing `#[non_exhaustive]` from an item.
  - This change is difficult to reverse, and warrants closer attention. Per semver, removing `#[non_exhaustive]` can be done in a patch release, but adding it back would then require a new major version.

CAD97:
> Own public types should be Debug:

That's already available in the compiler as `#[warn(missing_debug_implementations)]`, and already only fires for `pub` types.

obi1kenobi:
> That's already available in the compiler as `#[warn(missing_debug_implementations)]`, isn't it?

Oh, neat, TIL. It appears to be [allowed by default](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html) and has to be enforced by manually enabling the check. In that case, perhaps the wish-listed query should be checking that `#![deny(missing_debug_implementations)]` is set instead.

epage:
This also gets into a conversation that I think we only had over zulip so good to summarize here.

Especially if we want this in cargo some day, I think we should clearly define the scope.

`cargo clippy` is meant for linting an API as it exists

`cargo semver-checks` would be meant for linting changes in an API
- `missing_debug_implementations` is an example of something that imo doesn't belong in `cargo semver-checks`
- Linting that a lint is enabled is both getting a bit meta and again something that should be out of scope

Misc notes
- Making it easier to add lints to `clippy` is a conversation with the clippy folks and they are interested in solving it
- User-generated lints in either type of tool shipped with rustup would likely be marked as unstable initially.  A path to being stable is dependent on how comfortable people are on stabilizing the query language and the data model which is a large surface area
- In the mean time, there could be room for a linter that handles user defined lints.

obi1kenobi:
[...]

PR #284, opened 2023-01-10
New lint: Trait method `#[must_use]` added

SmolSir:
This is a part of solving https://github.com/obi1kenobi/cargo-semver-checks/issues/159, as well as splitting https://github.com/obi1kenobi/cargo-semver-checks/pull/268 into more manageable, smaller PRs.

Implements the check against adding `#[must_use]` attribute to a public Trait's method.

### DIFF
--- a/test_crates/trait_method_must_use_added/new/Cargo.toml
+++ b/test_crates/trait_method_must_use_added/new/Cargo.toml
@@ -1,4 +1,5 @@
 [package]
+publish = false
 name = "trait_method_must_use_added"
 version = "0.1.0"
 edition = "2021"
--- a/test_crates/trait_method_must_use_added/new/Cargo.toml
+++ b/test_crates/trait_method_must_use_added/new/Cargo.toml
@@ -1,0 +1,6 @@
+[package]
+name = "trait_method_must_use_added"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
--- a/test_crates/trait_method_must_use_added/new/src/lib.rs
+++ b/test_crates/trait_method_must_use_added/new/src/lib.rs
@@ -68,6 +68,70 @@ pub trait TraitWithProvidedMustUseMethods {
 }
 
 
+pub trait TraitWithDeclaredToProvidedMustUseMethods {
+
+    // These methods did not have the #[must_use] attribute in the old version.
+    // Addition of the attribute should be reported.
+
+    #[must_use]
+    fn DeclaredMethodToProvidedMustUseMethod(&self) {}
+
+    #[must_use = "Foo"]
+    fn DeclaredMethodToProvidedMustUseMessageMethod(&self) {}
+
+
+    // These methods had the #[must_use] attribute in the old version. Changes of
+    // the attribute, including deletion, should not be reported.
+
+    fn DeclaredMustUseMethodToProvidedMethod(&self) {}
+
+    #[must_use = "Foo"]
+    fn DeclaredMustUseMethodToProvidedMustUseMessageMethod(&self) {}
+
+
+    // These methods had the #[must_use] attribute in the old version.
+    // They also included the user-defined warning message. Changes of
+    // the attribute, including deletion, should not be reported.
+
+    fn DeclaredMustUseMessageMethodToProvidedMethod(&self) {}
+
+    #[must_use]
+    fn DeclaredMustUseMessageMethodToProvidedMustUseMethod(&self) {}
+
+    #[must_use = "Baz"]
+    fn DeclaredMustUseMessageMethodToProvidedMustUseMessageMethod(&self) {}
+}
+
+
+// This trait's methods were provided in the old version of the crate, but
+// their bodies have been removed turning them into declared methods.
+// They should NOT be reported by this rule to avoid duplicate lints.
+// They should be reported as changes of the trait's provided methods into
+// declared methods.
+
+pub trait TraitWithProvidedToDeclaredMustUseMethods {
+
+    #[must_use]
+    fn ProvidedMethodToDeclaredMustUseMethod(&self);
+
+    #[must_use = "Foo"]
+    fn ProvidedMethodToDeclaredMustUseMessageMethod(&self);
+
+    fn ProvidedMustUseMethodToDeclaredMethod(&self);
+
+    #[must_use = "Foo"]
+    fn ProvidedMustUseMethodToDeclaredMustUseMessageMethod(&self);
+
+    fn ProvidedMustUseMessageMethodToDeclaredMethod(&self);
+
+    #[must_use]
+    fn ProvidedMustUseMessageMethodToDeclaredMustUseMethod(&self);
+
+    #[must_use = "Baz"]
+    fn ProvidedMustUseMessageMethodToDeclaredMustUseMessageMethod(&self);
+}
+
+
 // This trait is private and adding #[must_use] to its methods
 // should NOT be reported.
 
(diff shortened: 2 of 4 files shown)


obi1kenobi:
Looks good so far!

SmolSir:
Unfortunately I'm in a spot of bother with the query. Say we have the following code:
```rust
pub trait Foo {

    #[must_use]
    fn bar();
}
```
The thing is - how do we access the method `bar()` inside `Foo` to get its attributes? I looked through the schema, but could not figure out a way to get there. Even if there was an `impl` block (unlike above) in which we can find the method and its `implemented_trait`, we don't need its attributes there (because of (1) from [this comment](https://github.com/obi1kenobi/cargo-semver-checks/pull/282#issue-1527754161)) and it pretty much comes down to the code above - given a trait, access its method and attribute. Maybe you have a quick answer to this, just like with `__typename` that I missed?

obi1kenobi:
I think it requires a bit of new schema. The method `bar()` is _kind of_ like the existing `Method` type, but its body is optional (it could be `fn bar();` or `fn bar() {}`) unlike `Method` which currently must have a body.

The body presence or absence information comes from the `has_body` JSON field here: https://docs.rs/rustdoc-types/latest/rustdoc_types/struct.Function.html

We have a few different options here:
- We could add a property (e.g. `has_body`) to the existing `Method` type, and then guarantee that `has_body` is always `true` on everything except traits.
- We could add a separate `FunctionLike & Item` type analogous to `Method` but with the extra `has_body` field, and only use it for trait methods.
- We could turn `Method` into an interface, and derive the new type from `Method` while also adding the new `has_body` field there.

Perhaps there are other alternatives as well?

What do you think would be the best approach?

SmolSir:
The `has_body` field surely would be handy here! Thanks to your ideas, I think I came up with a possible solution.
> We could add a property (e.g. `has_body`) to the existing `Method` type, and then guarantee that `has_body` is always `true` on everything except traits.

This is almost perfect, but I see one point of contention - the `fn bar () {}` does have a body, but would be marked with `has_body = false` when inside a trait? If so, how would we check that the method is inside the same trait in both `baseline` and `current` version using the `Method` schema? (we sadly can't use `ImplOwner` here)

> We could add a separate `FunctionLike & Item` type analogous to `Method` but with the extra `has_body` field, and only use it for trait methods.

This is the sweet spot in my opinion - if we could solve the problem of "what trait am I inside?" by adding a field for that as well, we would be able to query for trait methods and even distinguish declared and provided methods. I would also add the `has_body` to the regular `Method` as well for potential future use.

> We could turn `Method` into an interface, and derive the new type from `Method` while also adding the new `has_body` field there.

This could be more wise thinking about future lints than previous one if we could know what trait the method is in with the derived type. I'm only not sure if there would be anything else deriving from it in the future - if not, would it be worth it to have an Interface with just two objects deriving it? I'm not experienced enough with project structuring to tell, though.

Overall, I would probably go with the middle option since it appears to me as the one that is not just sufficient, but also reliable and beneficial for the whole schema.
What do you think about this, knowing the adapter best and having a lot more experience with project design?

obi1kenobi:
> This is almost perfect, but I see one point of contention - the `fn bar () {}` does have a body, but would be marked with `has_body = false` when inside a trait?

Ah sorry for the confusion, on traits `has_body` would only be true if the method has a body, like that `bar` method for example. It could be true or false on trait methods, we only guarantee it's true outside of traits.

> Even if so, how would we check that the method is inside the same trait in both `baseline` and `current` version using the `Method` schema? (we sadly can't use `ImplOwner` here)

Would our current way of ensuring types are "the same" across `baseline` and `current` not work for some reason? Right now `Method` is only reachable after crossing the `impl` or `inherent_impl` edges, and I think we'd add one more edge on `Trait` for accessing its methods. Then we'd make sure it's the same trait in the usual way.

> This is the sweet spot in my opinion - if we could solve the problem of "what trait am I inside?" by adding a field for that as well, we would be able to query for trait methods and even distinguish declared and provided methods. I would add the `has_body` to the regular `Method` as well for potential future use.

If we're adding `has_body` to the regular `Method`, then I'm not sure I see the advantage to making a separate type for trait methods. It seems like they'd have the same fields and would share the same underlying rustdoc representation, and at that point may as well also share the same type in the schema.

I'm guessing my confusing wording of the first option probably disqualified it immediately from your consideration. Has any of this changed your mind?

SmolSir:
> I'm guessing my confusing wording of the first option probably disqualified it immediately from your consideration. Has any of this changed your mind?

Yes, the first option is now definitely the best one! And combined with
> Right now `Method` is only reachable after crossing the `impl` or `inherent_impl` edges, and I think we'd add one more edge on Trait for accessing its methods. Then we'd make sure it's the same trait in the usual way.

will allow for successfully accessing the trait methods. If this is the final choice, I will go for these additions to the schema, as well as the `Union` item one.

One more question, related to the lint but not the schema - turns out we don't have a check against turning the method `provided` in `baseline` into a `declared` method in `current` version, which obviously should not be allowed. Making the change the opposite way is fine I believe. 
Should we consider it when writing this lint, or just ignore the `has_body` field and focus on the attribute?
Should I open an issue for this case with request for a new lint once the schema additions will be complete?

obi1kenobi:
> If this is the final choice, I will go for these additions to the schema, as well as the `Union` item one.

Sounds great! Go for it!

> One more question, related to the lint but not the schema - turns out we don't have a check against turning the method `provided` in `baseline` into a `declared` method in `current` version, which obviously should not be allowed. Making the change the opposite way is fine I believe. Should we consider it when writing this lint, or just ignore the `has_body` field and focus on the attribute? Should I open an issue for this case with request for a new lint once the schema additions will be complete?

I think that should be a new lint, and not reported here — it would be great to have a test case for it. If you're interested in writing that new lint as well, go for it! And yes, feel free to open an issue for it. I believe this new lint is mentioned in #5 so just make sure to link it to the appropriate entry (or make one if I'm misremembering and it isn't actually in the list there).

SmolSir:
> Sounds great! Go for it!

> I think that should be a new lint, and not reported here — it would be great to have a test case for it. If you're interested in writing that new lint as well, go for it! And yes, feel free to open an issue for it. I believe this new lint is mentioned in https://github.com/obi1kenobi/cargo-semver-checks/issues/5 so just make sure to link it to the appropriate entry (or make one if I'm misremembering and it isn't actually in the list there).

Will do all of these! Thank you for answering with great suggestions and details, I believe there are no more uncertainties as far as this lint goes for now, great to have it all be clear :)

obi1kenobi:
Awesome, I'm happy to help and excited to help you keep building!

obi1kenobi:
@SmolSir just checking in since it's been a couple of months since the last activity on this PR — are you still interested in working on this?

SmolSir:
Yes, I just shifted the focus to writing our thesis after the winter break. I planned to code again after writing one more chapter in about a week from now if that's not a problem!

obi1kenobi:
@SmolSir checking in again, are you still planning on finishing up this PR or should I reassign it to someone else?

SmolSir:
I'm sorry for blocking this for so long, last time when I said that I will sit down and finish it we got overwhelmed by homeworks 😞  Currently my plan is to study hard for our bachelor's exam at the end of June, so I can get back to this no sooner than July. If you have somebody to assign this to sooner than that, feel free to do so (I will still keep my eye on this in case there were any questions about what has been done so far).

obi1kenobi:
No worries, good luck with exams!

obi1kenobi:
Closing for staleness (no worries!), and I'm sure another contributor can pick this up soon.

PR #286, opened 2023-01-11
Optimize the `#[must_use]` lint queries by improving filtering.

obi1kenobi:
Most items are not `#[must_use]`. By moving the more-selective branch of the query (the `current` one that looks for `#[must_use]`), the query will evaluate fewer items because it will discard more of them earlier in the process.

The speedup is approximately a factor of 1600-2000x when measured on a particularly large crate: `aws-sdk-ec2`
https://github.com/awslabs/aws-sdk-rust/tree/main/sdk/ec2 

Running just the 4 affected lints on that crate:
- Before: 401-413s total scan time
- After: 0.21-0.25s total scan time

The numbers only include lint execution time, excluding things such as the time needed to build the rustdoc JSON.

This is an optimization that requires knowing the underlying shape of the dataset being queried, so it's not something that Trustfall could have applied automatically without access to statistics.

### DIFF
--- a/src/lints/enum_must_use_added.ron
+++ b/src/lints/enum_must_use_added.ron
@@ -10,32 +10,14 @@ SemverQuery(
     query: r#"
     {
         CrateDiff {
-            baseline {
-                item {
-                    ... on Enum {
-                        visibility_limit @filter(op: "=", value: ["$public"]) @output
-                        name @tag @output
-
-                        importable_path {
-                            path @tag @output
-                        }
-
-                        attribute @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
-                            content {
-                                base @filter(op: "=", value: ["$must_use"])
-                            }
-                        }
-                    }
-                }
-            }
             current {
                 item {
                     ... on Enum {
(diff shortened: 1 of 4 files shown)


obi1kenobi:
Adding @SmolSir since he's working on the other `#[must_use]` lints.

Perhaps also of general interest to @tonowak, @staniewzki, and @mgr0dzicki.

SmolSir:
That's astonishing! Because our test crates don't contain such enormous tests, I've never thought about their efficiency before (prioritizing the "does it work well" side). I will take this into account from now on when writing lints, as well as apply the change to the existing `#[must_use]` branches.

Is it a general rule that we should follow to do the more selective part of the query first like here (if possible), or is there some other aspect with such high impact?

obi1kenobi:
> Is it a general rule that we should follow to do the more selective part of the query first

Yes, this is a core operation in statistics-based query planning (e.g. in SQL databases): try to load and process as little data as possible.

It's a good idea to look through your past lints and see if any of them might benefit from flipping the current and baseline branches. Feel free to ping me if you aren't sure.

Issue #297, opened 2023-01-20
False positive: Changing struct with no pub fields to another kind of type is not breaking by itself


obi1kenobi:
Consider the following code change, from:
```rust
pub struct Foo {
    x: i64
}

impl Foo {
  pub fn new(x: i64) -> Self {
    Self { x }
  }
}
```
to:
```rust
pub enum Foo {
    Bar(i64),
}

impl Foo {
  pub fn new(x: i64) -> Self {
    Self::Bar(x)
  }
}
```

This isn't a breaking change. The struct's field wasn't visible, the struct wasn't externally constructible with a literal (due to the private field), and its pub methods are preserved across the change.

Instead, this is semver-minor: the new code publicly allows `Foo::Bar(x)` whereas the old code had no such ability. I believe changing the struct to a union would also be semver-minor for similar reasons.

obi1kenobi:
Here's a decision tree to decide whether a struct deletion / alteration is breaking:
```
deleted entirely? -> breaking

not deleted but changed type ->
    any pub fields -> breaking (field access)
        enums don't have fields
        union fields require `unsafe` to access

    no fields at all, but exhaustive -> breaking (constructible)
        enums require a variant to construct
        unions cannot have zero fields
```

The best way forward is:
- [ ] alter `struct_missing` to handle the "deleted entirely" case only
- [ ] confirm that the `struct_pub_field_missing` lint catches the "field access" breaking change when structs are converted to other types
- [ ] add a lint to cover the last, "constructible fieldless struct" case

obi1kenobi:
@epage is changing a struct into a trait worth covering off as an edge case, or is it too rare to worry about? With the plan above, structs changed into traits would be reported as "removed" in the same way as the earlier struct -> enum case you brought up. The fix is doable but a bit more work, and I'm not sure if it's the most pressing thing at the moment.

epage:
Replacing a struct with a trait is likely rare enough to be ok but it feels a bit hacky.  Personally, I would structure it around [namespaces / universes](https://doc.rust-lang.org/stable/reference/names/namespaces.html) for what is removed and then have other lints for what happens within a namespace

obi1kenobi:
Makes sense. I'll open a couple of separate issues for the "replacing struct with trait" and similar edge cases.

CAD97:
No, replacing a `struct` with an `enum` is breaking:

```rust
fn breaks(value: Chameleon) {
[...]

PR #299, opened 2023-01-20
Non-constructible struct becoming an enum is not per se breaking.

obi1kenobi:
Add a test case where a non-constructible struct becomes an enum. This checks that:
- Inherent methods on the struct are found and not reported as missing by `inherent_method_missing`.
- The struct changing to an enum is not reported by `struct_missing`. This check is currently failing, and the false-positive outcome is checked in at the moment, in the interest of strengthening our test coverage.
 
#297 tracks fixing this false-positive.


epage (commented):
Looks good

PR #300, opened 2023-01-20
Only parse test rustdocs once, for ~2x test time speedup.

obi1kenobi:
The original code used to parse each test crate's rustdoc JSON once per test case, which was wasteful. Each lint's test runs the lint on every test crate, so all test crates are always loaded even when running only a single test. The new test code loads test crates only once total, and runs approximately ~2x faster as a result.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -106,6 +106,67 @@ mod tests {
     use crate::query::SemverQuery;
     use crate::templating::make_handlebars_registry;
 
+    lazy_static::lazy_static! {
+        static ref TEST_CRATE_NAMES: Vec<String> = get_test_crate_names();
+
+        /// Mapping test crate (pair) name -> (old rustdoc, new rustdoc).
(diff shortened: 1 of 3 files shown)


src/query.rs
```diff
@@ -115,8 +176,8 @@ mod tests {
 
     #[test]
     fn all_queries_parse_correctly() {
-        let current_crate = load_pregenerated_rustdoc("template", "new");
-        let indexed_crate = VersionedIndexedCrate::new(&current_crate);
+        let current_crate = &TEST_CRATE_RUSTDOCS["template"].1;
```
tonowak, line 179: Wouldn't a `(_previous_crate, current_crate)` be more readable?

tonowak: Instead of `.1`

obi1kenobi: Yeah, good call.

tonowak: Also, `TEST_CRATE_RUSTDOC` kind of suggests that the map contains values that are one crate / one rustdoc, which is not really the case.

tonowak (approved):

Issue #309, opened 2023-01-21
The app doesn't compare between the local version and the published one


MarcoIeni:
### Steps to reproduce the bug with the above code

```sh
git clone https://github.com/rust-github/rust-gh-example.git
cd rust-gh-example
cargo semver-checks check-release
```

output:

```
$ cargo semver-checks check-release
    Updating index
     Parsing rust-gh-example v0.1.8 (current)
     Parsing rust-gh-example v0.1.7 (baseline)
    Checking rust-gh-example v0.1.7 -> v0.1.7 (no change)
   Completed [   0.008s] 33 checks; 33 passed, 0 unnecessary
```

Question: why does it compare 0.1.7 with 0.1.7? I would expect it compares current and baseline.

Anyway, now I try to break the api by adding the `#[must_use]` attribute:

```rust
#[must_use]
pub fn answer() -> u32 {
    42
}
``` 

I run `cargo semver-checks check-release` again.

### Actual Behaviour

The check still pass:

```
$ cargo semver-checks check-release
    Updating index
     Parsing rust-gh-example v0.1.8 (current)
     Parsing rust-gh-example v0.1.7 (baseline)
    Checking rust-gh-example v0.1.7 -> v0.1.7 (no change)
   Completed [   0.008s] 33 checks; 33 passed, 0 unnecessary
```

### Expected Behaviour

I would expect the program to tell me I have added the must_use attribute.

#### Software version

cargo-semver-checks 0.15.2

#### Operating system

macOS 13.1 (Darwin 22.2.0)

#### Command-line

```bash
/Users/marcoieni/.cargo/bin/cargo-semver-checks semver-checks --bugreport
```

#### cargo version

```
> cargo -V
cargo 1.66.1 (ad779e08b 2023-01-10)
```

#### Compile time information

- Profile: release
- Target triple: aarch64-apple-darwin
- Family: unix
- OS: macos
- Architecture: aarch64
- Pointer width: 64
- Endian: little
- CPU features: aes,crc,dit,dotprod,dpb,dpb2,fcma,fhm,flagm,fp16,frintts,jsconv,lor,lse,neon,paca,pacg,pan,pmuv3,ras,rcpc,rcpc2,rdm,sb,sha2,sha3,ssbs,vh
- Host: aarch64-apple-darwin

### Additional Context

I'm trying to use this library, but I didn't find a way to simulate a semver violation. I'm sure I'm missing something, but I tried many things. How can I test this library locally?
[...]

PR #310, opened 2023-01-21
Update issue template to reference `CARGO_` env vars.

obi1kenobi:
In #309, the build configuration was in an env var, not in a `cargo.toml`. Advise bug reporters about this possibility.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -29,7 +29,7 @@ body:
   - type: textarea
     attributes:
       label: "Build Configuration"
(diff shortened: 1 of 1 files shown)


tonowak (commented):
Maybe it would be good to also write a command for listing env vars.

obi1kenobi:
> Maybe it would be good to also write a command for listing env vars.

By the time we cover all the supported architectures and adequately advise users on how to filter out sensitive env vars they shouldn't be putting on GitHub, I think we'd be better off doing #311 instead.

This is a quick-and-dirty fix meant to last only until we can get #311 done, and I don't think we want to invest too much effort into it since it's all going away when #311 is completed.

tonowak (approved):

Issue #67, opened 2022-08-12
Allow usage as a library, not just as a binary.


drahnr:
Providing a library alongside the cli tool would be much appreciated for tools like https://GitHub.com/paritytech/cargo-unleash even before it becomes part of cargo

Related: https://github.com/obi1kenobi/cargo-semver-check/issues/61

CC @xanewok @gnunicorn

epage:
The cargo-release issue for this is https://github.com/crate-ci/cargo-release/issues/62 (though I suspect we won't auto-select version but check if the version is appropriate).

My main concern for cargo-release will be the performance of cargo-semver-check due to the time it takes to generate the rustdoc json.

btw @drahnr any interest in collaborating on release tools?

drahnr:
Hey @epage - I'd love to, but I am not sure I can carve out any kind of significant time for the remainder of at least this year.

obi1kenobi:
> My main concern for cargo-release will be the performance of cargo-semver-check due to the time it takes to generate the rustdoc json.

A few notes to help bring up to speed folks coming across this issue, since @epage and I discussed it in person. TL;DR: JSON generation is likely always going to be the long pole in the runtime, and there's plenty of low-hanging fruit to speed up the checking itself that we haven't bothered with yet because the JSON is the slow part.

When rustdoc JSON is stabilized and docs.rs [starts hosting the JSON](https://github.com/rust-lang/docs.rs/issues/1285) we'll be able to cut that time in half: we'll download the baseline version's JSON instead of running a `git checkout` + generating it again. It's also possible that once the JSON format is stabilized, we might be able to get some speedups there as well. Right now, the focus seems to rightly be on stabilization, not performance.

To make the checking itself faster, with fairly trivial effort I think we can get small multiples of 10x speedup on the checking time. With a bit more effort I think we can get more orders of magnitude. Consider that an `E-help-wanted + E-mentor` opportunity :)

> Providing a library alongside the cli tool would be much appreciated

A few other folks have reached out to ask for the same thing, and not necessarily just for use in 3rd party cargo plugins. I'd be happy to do it as time allows, and would appreciate help. I hope you won't mind me renaming the issue to generalize it to the library aspect rather than the cargo plugins.

MarcoIeni:
Hey Predrag! As we discussed on Twitter, I'm happy to be a "beta tester" for this feature and integrate it in https://github.com/MarcoIeni/release-plz :)
[...]

PR #314, opened 2023-01-21
extract library

MarcoIeni:
A step towards #67 
To keep things easier I just created a new lib.rs
In the future, we could extract this lib.rs file into a `core` crate, that doesn't depend on clap.


src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+    /// Git revision to lookup for a baseline.
+    Revision(String),
+    /// Directory containing baseline crate source.
+    Root(PathBuf),
+    /// The rustdoc json file to use as a semver baseline.
+    RustDoc(PathBuf),
+    /// Latest version published to the cargo registry.
+    #[default]
+    LatestVersion,
+}
```
MarcoIeni, line 41: I tried to group conflicting cli arguments together in enums.

src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+
+/// Current version of the project to analyze.
+#[derive(Default)]
+enum Current {
+    /// Path to the manifest of the current version of the project.
+    /// It can be a workspace or a single package.
+    Manifest(PathBuf),
+    /// The rustdoc json of the current version of the project.
+    RustDoc(PathBuf),
```
MarcoIeni, line 50: I saw that this rustdoc also requires the Baseline rustdoc.
At the moment this is not captured by the type system. How would you deal with this?

src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+    Packages(Vec<String>),
+    /// All packages in the workspace. Equivalent to `--workspace`.
+    Workspace,
+    /// Default members of the workspace.
+    #[default]
+    DefaultMembers,
+}
+
+impl Scope {
+    fn selected_packages<'m>(
```
MarcoIeni, line 75: this function is very similar to [partition_packages](https://docs.rs/clap-cargo/latest/clap_cargo/struct.Workspace.html#method.partition_packages).
Initially, I wanted to use it, but since the struct `clap_cargo::Workspace` is non-exhaustive, it cannot be constructed.

src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+
+impl Check {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn with_manifest(&mut self, path: PathBuf) -> &mut Self {
+        self.current = Current::Manifest(path);
+        self
+    }
```
MarcoIeni, line 120: builder pattern: https://rust-lang.github.io/api-guidelines/type-safety.html#non-consuming-builders-preferred

src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        self.log_level = Some(log_level);
+        self
+    }
+
+    fn manifest_path(&self) -> anyhow::Result<PathBuf> {
+        let path = match &self.current {
+            Current::Manifest(path) => path.clone(),
+            Current::RustDoc(_) => {
+                anyhow::bail!("error: RustDoc is not supported with these arguments.")
+            }
```
MarcoIeni, line 172: I'm not sure about this.

src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        let metadata = self.manifest_metadata_no_deps()?;
+        let target = metadata.target_directory.as_std_path().join(util::SCOPE);
+        let registry = baseline::RegistryBaseline::new(&target, config)?;
+        Ok(registry)
+    }
+}
+
+pub struct Report {
+    success: bool,
+}
```
MarcoIeni, line 304: the report should contain the failing checks.
For now, the library prints to stdout and just reports if it fails or succeeds.

src/config.rs
```diff
@@ -3,14 +3,20 @@ use termcolor::{ColorChoice, StandardStream};
...
     stdout: StandardStream,
     stderr: StandardStream,
     handlebars: handlebars::Handlebars<'static>,
 }
 
+impl Default for GlobalConfig {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
MarcoIeni, line 18: I added this to fix a clippy lint

obi1kenobi (commented):
Very excited about this, thanks for putting it together!

src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+        Self::Version(None)
+    }
+}
+
+/// Current version of the project to analyze.
+#[derive(Default)]
+enum Current {
+    /// Path to the manifest of the current version of the project.
+    /// It can be a workspace or a single package.
+    Manifest(PathBuf),
```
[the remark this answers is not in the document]
MarcoIeni: In this way, I don't have to mess with lifetimes :)

src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+                    )?;
+                    rustdoc_paths.push((crate_name.clone(), baseline_path, rustdoc_path));
+                }
+                rustdoc_paths
+            }
+        };
+        let mut success = true;
+        for (crate_name, baseline_path, current_path) in rustdoc_paths {
+            let baseline_crate = load_rustdoc(&baseline_path)?;
+            let current_crate = load_rustdoc(&current_path)?;
```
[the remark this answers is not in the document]
MarcoIeni: Shall we merge this pr after #270?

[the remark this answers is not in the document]
tonowak: #270 has been merged.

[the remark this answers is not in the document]
MarcoIeni: merged here in [7712fae](https://github.com/obi1kenobi/cargo-semver-checks/pull/314/commits/7712fae0eb1c07dbe7c77e6d699f06b5c810ad02)

src/lib.rs
MarcoIeni: So we want to create a RustdocSource struct that can be used both by the Baseline and Current rustdoc?

> already generated, just load it from an existing file

```rust
enum RustdocSource {
    RustDoc(PathBuf)
}
```

> generate it based on code at some path, optionally first checking out a gitrev (maybe better split into two cases? what do you think?)



```rust
enum RustdocSource {
    RustDoc(PathBuf),
    Root(PathBuf),
    // (path to root, rev)
    Revision(PathBuf, String)
}
```

> generate it based on a registry version

```rust
enum RustdocSource {
    RustDoc(PathBuf),
    Root(PathBuf),
    // (path to root, rev)
    Revision(PathBuf, String),
    Version(Option<String>),
}
```

This looks similar to the current Baseline enum. Is it valid for `Current`, too?
If I didn't get it, can you give me the structure of the enum you were thinking about?

obi1kenobi: I think it looks reasonable. It currently isn't valid for `Current` in the CLI, but I think it should be in the lib. Not much reason not to: that would make the types more complex while providing less functionality.

Small nit: let's call the variant `Rustdoc` instead, since the underlying tool doesn't capitalize the `D` either.

MarcoIeni:
Hey @tonowak, if I'm blocking you, feel free to contribute to this pr, yes 👍 
For example, today I won't have time to look into [this](https://github.com/obi1kenobi/cargo-semver-checks/pull/314#discussion_r1083390720)

src/lib.rs
```diff
@@ -0,0 +1,367 @@
...
+        }
+    }
+
+    fn registry(&self, config: &mut GlobalConfig) -> anyhow::Result<baseline::RegistryBaseline> {
+        let manifest_dir = match &self.source {
+            RustdocSource::Root(manifest_dir) | RustdocSource::Revision(manifest_dir, _) => {
+                manifest_dir
+            }
+            RustdocSource::Version(_) | RustdocSource::Rustdoc(_) => {
+                anyhow::bail!("not supported yet")
```
MarcoIeni, line 85: probably for this pr this is good enough.
As a future improvement, maybe we could use a temporary directory as "target".

obi1kenobi: I'm quite confused about what this function does, and the name isn't helping me out much. Is there a more descriptive name we could use for it?

I'm also confused by the fact that the function only works for the `Root` and `Revision` variants, but generates a `RegistryBaseline` result on success.

MarcoIeni: I inlined the function. In this way, it should be easier to review because it is easier to compare with the [original code](https://github.com/obi1kenobi/cargo-semver-checks/blob/ee1fa79054ccaaab762845f920bd95782e246fff/src/main.rs#L117). 👍

MarcoIeni:
> Feel free to commit directly into the branch, or to raise prs against it.

Nevermind Tomasz, I found some time to work on this tonight 👍

src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+struct Scope {
+    selection: ScopeSelection,
+    excluded_packages: Vec<String>,
+}
+
+/// Which packages to analyze.
+#[derive(Default, PartialEq, Eq)]
+enum ScopeSelection {
+    /// Package to process (see `cargo help pkgid`)
+    Packages(Vec<String>),
```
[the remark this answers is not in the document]
MarcoIeni: In 1190770 I have made it impossible for this to happen

tonowak:
No problem! I see you're progressing faster than I can help :)

src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+use trustfall_rustdoc::{load_rustdoc, VersionedCrate};
+
+use dump::RustDocCommand;
+use itertools::Itertools;
+use semver::Version;
+use std::collections::HashSet;
+use std::path::{Path, PathBuf};
+
+/// Test a release for semver violations.
+pub struct Check {
```
obi1kenobi, line 26: We should probably `#[derive(Debug)]` and `#[non_exhaustive]` at least for all these pub structs and enums, unless there's a reason not to?

src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+    /// Which packages to analyze.
+    scope: Scope,
+    current: Rustdoc,
+    baseline: Rustdoc,
+    log_level: Option<log::Level>,
```
obi1kenobi, line 31: Is this different than `GlobalConfig.log_level`? I'm confused by the duplication.

src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+
+impl Rustdoc {
+    /// Use an existing rustdoc file.
+    pub fn from_path(rustdoc_path: impl Into<PathBuf>) -> Self {
+        Self {
+            source: RustdocSource::Rustdoc(rustdoc_path.into()),
+        }
+    }
+
+    /// Generate the rustdoc file from the project root directory, i.e. the directory containing the crate source.
```
obi1kenobi, line 46: Just breaking a long line.
```suggestion
    /// Generate the rustdoc file from the project root directory,
    /// i.e. the directory containing the crate source.
```

src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+    /// It can be a workspace or a single package.
+    /// Same as `from_git_revision`, but with the current git revision.
```
obi1kenobi, line 48: Is Rust doc comment style to include or omit the parens when naming functions? I genuinely am not sure, we should probably mirror what `std` does here.

src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+    /// Generate the rustdoc file from the largest-numbered non-yanked non-prerelease version
+    /// published to the cargo registry. If no such version, uses
+    /// the largest-numbered version including yanked and prerelease versions.
+    pub fn from_latest_version() -> Self {
+        Self {
+            source: RustdocSource::Version(None),
+        }
+    }
+
+    pub fn from_version(version: impl Into<String>) -> Self {
```
obi1kenobi, line 74: We probably want a doc comment here as well.

src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+        Self {
+            source: RustdocSource::Version(Some(version.into())),
+        }
+    }
+}
+
+enum RustdocSource {
+    /// Path to the Rustdoc json file. Use this option when you have already generated the rustdoc file.
```
obi1kenobi, line 82: Just breaking another long line. Generally would be nice to keep line length below 100 chars max.
```suggestion
    /// Path to the Rustdoc json file.
    /// Use this option when you have already generated the rustdoc file.
```

src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    pub fn from_version(version: impl Into<String>) -> Self {
+        Self {
+            source: RustdocSource::Version(Some(version.into())),
+        }
+    }
+}
+
+enum RustdocSource {
```
obi1kenobi, line 81:
```suggestion
#[derive(Debug)]
enum RustdocSource {
```
for all types unless there's a reason not to?

src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+                    rev,
+                    &mut config,
+                )?)
+            }
+            RustdocSource::Version(version) => {
+                let mut registry = {
+                    let manifest_dir = match &self.current.source {
+                        RustdocSource::Root(manifest_dir)
+                        | RustdocSource::Revision(manifest_dir, _) => manifest_dir,
+                        RustdocSource::Version(_) | RustdocSource::Rustdoc(_) => {
```
obi1kenobi, line 254: Hmm is `--current-rustdoc` not supported together with `--baseline-version`? I'm not sure why that might be the case? I'd have expected it to just assume that the manifest is in the current directory and up-to-date in that case, and bail if it is not.

If indeed it doesn't work today, then I will update my own prior — no code change needed in that case :)

src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+                                &mut config,
+                                crate_name,
+                                current_crate,
+                                baseline_crate,
+                            )?)
+                        }
+                    })
+                    .collect()
+            }
+            RustdocSource::Revision(_, _) | RustdocSource::Version(_) => todo!(),
```
obi1kenobi, line 339: I believe these are currently unimplemented in the CLI, so I'm fine with leaving a panic in this branch.

But let's make it `unimplemented!()` with a suitable message instead of a bare `todo!()`, which makes it seem like we forgot to add it into the PR rather than making an explicit choice to not include it for now.

src/query.rs
```diff
@@ -40,20 +40,20 @@ impl ActualSemverUpdate {
 /// A query that can be executed on a pair of rustdoc output files,
 /// returning instances of a particular kind of semver violation.
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub(crate) struct SemverQuery {
-    pub(crate) id: String,
+pub struct SemverQuery {
```
obi1kenobi, line 43:
```suggestion
#[non_exhaustive]
pub struct SemverQuery {
```
We may want to add new fields in the future here.

src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+        };
+        let success = all_outcomes
+            .into_iter()
+            .fold_ok(true, std::ops::BitAnd::bitand)?;
+
+        Ok(Report { success })
+    }
+}
+
+pub struct Report {
```
obi1kenobi, line 349:
```suggestion
#[non_exhaustive]
#[derive(Debug)
pub struct Report {
```
I know `#[non_exhaustive]` is redundant here at the moment, but better safe than sorry. We don't want to end up in a situation where we've accidentally removed the last private field without realizing, and we're then locked into a constructible exhaustive struct until the next major version.

One day `cargo-semver-checks` will warn about "semver-regrettable" situations like that.

src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    pub fn with_excluded_packages(&mut self, packages: Vec<String>) -> &mut Self {
+        self.excluded_packages = packages;
+        self
+    }
+}
+
+#[derive(Default, PartialEq, Eq, Clone)]
+pub enum ScopeSelection {
```
obi1kenobi, line 136:
```suggestion
#[non_exhaustive]
#[derive(Debug, Default, PartialEq, Eq, Clone)]
pub enum ScopeSelection {
```
Unless there's a reason not to?

obi1kenobi (commented):
Sorry for the delay! I'm pretty happy with the shape of things here. Except for that one block that I'd love to get @epage's eyes on, I think this is looking pretty good. Just some minor "polish-level" suggestions that should be quickly resolvable.

Excited to ship this soon!

src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+                        let version = &selected.version;
+
+                        let is_implied = matches!(
+                            self.scope.mode,
+                            ScopeMode::DenyList(PackageSelection {
+                                selection: ScopeSelection::Workspace,
+                                ..
+                            })
+                        ) && selected.publish == Some(vec![]);
+                        if is_implied && selected.publish == Some(vec![]) {
```
tonowak, line 306: One of the `selected.publish == ...` is not needed.

src/lib.rs
MarcoIeni: Do you want to add `#[non_exhaustive]` even if the struct fields are private?
I don't think it has any effect in this case.

src/lib.rs
MarcoIeni: done in https://github.com/obi1kenobi/cargo-semver-checks/pull/314/commits/f626d7ebaed9becbf30edaaf24ff273a75fa0830

src/query.rs
MarcoIeni: done in https://github.com/obi1kenobi/cargo-semver-checks/pull/314/commits/368ee5b0c0e866e8dedc00d079ff4ac63cd764e1

src/lib.rs
MarcoIeni: done in [5a334ba](https://github.com/obi1kenobi/cargo-semver-checks/pull/314/commits/5a334bac2269731663c2c0f377911b61a1bc8b83)

src/lib.rs
MarcoIeni: By looking at [this](https://doc.rust-lang.org/std/env/index.html) page, it seems that the std uses the () convention.
I changed this in a496d420ecd96a1fe338bdc96a92f8960350a7a5

src/lib.rs
tonowak: > It currently isn't valid for Current in the CLI, but I think it should be in the lib. 

FYI @MarcoIeni, the recent merged chain of PRs (which ended with https://github.com/obi1kenobi/cargo-semver-checks/pull/351) made it so that the `current` crate's rustdoc is generated in the same way as the `baseline`'s rustdoc, so now it's possible to create the `current` rustdoc from registry and from git revision -- your `RustdocSource` now is also fully valid for the `current` crates.

tonowak: And also, this chain of PRs renamed quite a lot of things in the source code, so there might be a big merge conflict, sorry :( I can try to merge main into this branch when I'll have some free time.

obi1kenobi: If you could help with the merge, that would be great @tonowak. Probably the highest impact next thing you can do right now, TBH.

tonowak: Ok, I'll do it in max few days.

tonowak:
I've merged `main` into this branch. @obi1kenobi can you *carefully* review the merge? I had to change some logic in it.

tests/verify_binary_contains_lints.rs
```diff
@@ -5,11 +5,7 @@ fn verify_binary_contains_lints() {
     let assert_on_crate_pair = |crate_pair: &str| {
         let mut cmd = Command::cargo_bin("cargo-semver-checks").unwrap();
         cmd.current_dir(format!("test_crates/{crate_pair}/new"))
-            .args([
-                "semver-checks",
-                "check-release",
-                "--baseline-root=../old/Cargo.toml",
-            ])
+            .args(["semver-checks", "check-release", "--baseline-root=../old/"])
```
tonowak, line 8: I've made this change temporarily so that the test passes locally, we should revert this change.

From what I understand, the flag `--baseline-root` was meant to take a directory to the project, but the old code somehow worked both when it received a directory and when it received a path to a manifest (and I believe that we didn't even know what we had this feature). Passing a directory is "more correct" in the sense that it aligns with the description from `--help`, but I think there might be some users who might rely on the accidental feature of passing manifests (like this test), so I think that we should make the current code also work for both directories and manifests.

src/lib.rs
```diff
@@ -0,0 +1,458 @@
...
+fn manifest_metadata_no_deps(manifest_dir: &Path) -> anyhow::Result<cargo_metadata::Metadata> {
+    let manifest_path = manifest_from_dir(manifest_dir);
+    let mut command = cargo_metadata::MetadataCommand::new();
+    let metadata = command.manifest_path(manifest_path).no_deps().exec()?;
+    Ok(metadata)
+}
+
+fn get_target_dir_from_config() -> Option<PathBuf> {
+    // TODO: I'll implement this function probably tomorrow
+    None
```
tonowak, line 439: As the comment says, I'll come back to this function in a few days.

src/lib.rs
```diff
@@ -0,0 +1,458 @@
...
+        } else if let Some(path) = get_target_dir_from_project_root(source)? {
+            path
+        } else if let Some(path) = get_target_dir_from_project_root(&self.current.source)? {
+            path
+        } else if let Some(path) = get_target_dir_from_project_root(&self.baseline.source)? {
+            path
+        } else {
+            std::env::current_dir()?.join("target")
+        })
+    }
```
tonowak, line 248: Now that I think about this -- the logic behind this function is not obvious at all, I should document this.

tests/verify_binary_contains_lints.rs
obi1kenobi: I also think it's reasonable to be able to extrapolate the directory if we're passed the manifest. Seems like an easy way to do the right thing even if the user did the wrong thing.

Nice catch! We got lucky this time, so let's add a test so we don't have to rely on luck for next time?

src/lib.rs
MarcoIeni: done! :)

src/lib.rs
MarcoIeni: done :)

src/lib.rs
MarcoIeni: After the changes from [tonowak](https://github.com/tonowak) the `todo` is not there anymore 👍

src/lib.rs
MarcoIeni: fixed in [371ccab](https://github.com/obi1kenobi/cargo-semver-checks/pull/314/commits/371ccabd22e9253d6aa0ef85edd7774fbe327f42)

src/lib.rs
MarcoIeni: After the changes from [tonowak](https://github.com/tonowak), the `anyhow::bail` is not there anymore 👍

src/lib.rs
MarcoIeni: I decided not to expose GlobalConfig to the user. I'm only exposing the log level.
That's because if the user wants to use `cargo-semver-checks` as a library, I imagine `cargo-semver-checks` won't write to stdout or stderr in the future. So it doesn't make sense to allow the user to customize stdout and stderr.

MarcoIeni:
Thanks for helping Tomasz 😄 
I think I have addressed all the PR comments. I won't edit this PR anymore unless you still need something else from my side 👍

tests/verify_binary_contains_lints.rs
tonowak: OK, I've implemented it and I've added some tests. Also, the previous error messages of passing directory to `--manifest-path` were not clear. I'm not sure whether I made the right choice, but I've made it so it's possible to pass both a manifest and a directory path to the `--manifest-path` flag.

src/lib.rs
tonowak: Done in 2449acd377ed1792127a10c0ffed211094664d1b.

src/lib.rs
tonowak: We discussed on Zulip whether this is the best idea and we've decided to instead rely on the XDG cache directory: 2449acd377ed1792127a10c0ffed211094664d1b

src/lib.rs
tonowak: Done.

src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+    /// Project root directory, i.e. the directory containing the crate source.
+    /// It can be a workspace or a single package.
+    Root(PathBuf),
+    /// Project root directory and Git Revision.
+    Revision(PathBuf, String),
+    /// Version from cargo registry to lookup. E.g. "1.0.0".
+    /// If `None`, uses the largest-numbered non-yanked non-prerelease version
+    /// published to the cargo registry. If no such version, uses
+    /// the largest-numbered version including yanked and prerelease versions.
+    Version(Option<String>),
```
tonowak, line 101: Maybe it would be better to call it `Registry` instead of `Version`? It would make it more clear that it requires internet access to use it and from where this specific version is being taken from.

MarcoIeni: What about `RegistryVersion`? 😄 
The user when seeing `Registry` with a string might think about `Registry("crates.io")` for example

MarcoIeni: Naming is hard 😁
`VersionFromRegistry` is more verbose, but more clear maybe. 🤔

obi1kenobi: I'm fine with either `RegistryVersion` or `VersionFromRegistry`. I'm not that worried about people thinking that they need to pass a version of a registry, but I'm willing to accept I may be overly optimistic if you think the other option is better.

tonowak: What do you both think about 8485cc5dfe51a91ba2e8fe2b071c631f24f67cfb?

src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+            .silence(!config.is_verbose());
+
+        let current_loader = self.get_rustdoc_generator(&mut config, &self.current.source)?;
+        let baseline_loader = self.get_rustdoc_generator(&mut config, &self.baseline.source)?;
+
+        let all_outcomes: Vec<anyhow::Result<bool>> = match &self.current.source {
+            RustdocSource::Rustdoc(_)
+            | RustdocSource::Revision(_, _)
+            | RustdocSource::VersionFromRegistry(_) => {
+                let name = "<unknown>";
```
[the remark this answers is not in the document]
tonowak: In #207 I've made a temporary fix 7ecf5873242bc0defbb1eb739876cdc778614161 so that I can progress further. In the case where this fix is good enough, I can move it here, otherwise we should discuss better alternatives.

tonowak:
The CLI needs some polishing, but I'm not experienced in stuff like that -- @obi1kenobi can you review it?

I'm thinking that `semver-crater` might be useful not only for data gathering, but also for convincing people that the tool is worthy. The `semver-crater` allows to easily check whether there has been any issues in the user's crate (just run a single `cargo` command that install `cargo-semver-checks` and runs it on all releases of the user's crate) and to show people "look, some popular crates (like clap) had issues, here is how in a single command you can check that it is indeed true and the tool even detects it".

Assuming `semver-crater` would be used in this manner, we have to think about how to make its CLI convenient to use.

src/lib.rs
MarcoIeni: I think it's fine :)
I would not try to make the library perfect in this PR. Otherwise we will have to deal with too many merge conflicts.

I guess (and hope) that the library will be extracted in a separate crate, so we can still change the library API in future PRs.

obi1kenobi: This PR is on my to-do list to review today. Sorry for the delay! Trustfall ended up in the spotlight (it was [GitHub's #​1 most-starred project](https://twitter.com/PredragGruevski/status/1623711641127747584) a couple of days ago) and I've been super busy over there for the last couple of days.

I know I probably just merged another PR that causes another merge conflict, and I'm happy to fix the mess of my own making 😅

src/lib.rs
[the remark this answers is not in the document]
obi1kenobi: In https://github.com/obi1kenobi/cargo-semver-checks/commit/7ecf5873242bc0defbb1eb739876cdc778614161 it seems like there's still an `unknown` branch. That looks like it could work if `current` is given by gitrev, but won't work with a registry version nor with a premade rustdoc file.

Can we add an explicit `panic!()` in the cases we know for sure won't work, instead of setting a bogus value and then breaking in a confusing way?

src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+    fn get_target_dir(&self, source: &RustdocSource) -> anyhow::Result<PathBuf> {
+        Ok(
+            if let Some(path) = get_target_dir_from_project_root(source)? {
+                path
+            } else if let Some(path) = get_target_dir_from_project_root(&self.current.source)? {
+                path
+            } else if let Some(path) = get_target_dir_from_project_root(&self.baseline.source)? {
+                path
+            } else {
+                get_xdg_cache_dir()
```
obi1kenobi, line 252: I suspect this won't work on Windows, unfortunately. Let's instead use [the `directories` crate](https://crates.io/crates/directories) which is the known-working cross-platform way to do this:
```rust
if let Some(proj_dirs) = ProjectDirs::from("cargo-semver-checks") {
    proj_dirs.config_dir();
    // probable dirs (still need to be created! the library won't make them for us)
    // Lin: /home/alice/.config/cargo-semver-checks
    // Win: C:\Users\Alice\AppData\Roaming\cargo-semver-checks\config
    // Mac: /Users/Alice/Library/Application Support/cargo-semver-checks
}
```

obi1kenobi (commented):
This looks in pretty good shape. Let's target merging tomorrow or Monday morning if possible, so I can include this in the next release (also planned for Monday at the moment).

obi1kenobi:
I took care of the merge, it wasn't that bad because I reviewed the code that caused a conflict. Lmk what you think about the `panic!()` branch and using the cross-platform-compatible app dir crate!

Really looking forward to merging this soon! 🚀🎉

MarcoIeni:
> Lmk what you think about the panic!() branch and using the cross-platform-compatible app dir crate!

Is this question for me? If yes, what's the panic!() branch?

obi1kenobi:
The two remaining items needed to merge this are:
- [This `<unknown>` branch](https://github.com/obi1kenobi/cargo-semver-checks/pull/314/files#r1103730310) that seems likely to cause an obscure error if someone hits it, and I'm recommending an explicit `panic!()` instead. It's based on changes @tonowak made.
- [Using the cross-platform `directories` crate instead of `xdg`](https://github.com/obi1kenobi/cargo-semver-checks/pull/314#discussion_r1103730464) so that our non-Unix users aren't left behind.

I unfortunately have had to shift my primary focus to Trustfall for a bit, given [the recent massive uptick in interest in the project](https://twitter.com/StarHistoryHQ/status/1625421205321793536). I think the remaining items aren't much work and I trust both of you to do a good job on them.

MarcoIeni:
I replaced `xdg` with `directories`. I leave the other fix to tonowak :)

tonowak:
Thanks :) I'll probably do the other fix in the next 48h.

obi1kenobi (approved):
This looks polished enough to release, let's not block @MarcoIeni's `release-plz` work any longer. We will certainly keep improving the library APIs over time, we don't have to get to "perfect" on day 1.

Thanks, everyone!

MarcoIeni:
Thanks!
This week I'll work on extracting the library into a separate crate.
In this way, the users of the library don't need to compile `clap` and other specific libraries used only in the cli.

Some naming ideas:

- `semver-checks`
- `cargo-semver-checks-core`

Please, reserve in crates.io the one you prefer :)

obi1kenobi:
I grabbed both, just to be extra safe. `semver-checks` seems good to me, but I don't have a strong preference.

My 2 cents: splitting the library into a separate crate would definitely be nice, but it doesn't seem that critical to me. The community impact of having semver-checking made automatic via `release-plz` seems bigger, and we can always split the library into a separate crate and cut down compile times later.

Of course, `release-plz` is your crate, this is your work, and it's your call to make :)

tonowak:
Maybe it would be reasonable to have the `clippy` and CLI under a default flag?

MarcoIeni:
> CLI under a default flag

This would prevent us to ship breaking changes independently for the cli and for the library.

For example, if we change a command line argument, should it be a breaking change?

Having two separate crates allows us to ship breaking changes for cli and the lib independenlty

Issue #317, opened 2023-01-23
compiling semver-checks fails


demoray:
### Steps to reproduce the bug with the above code

cargo install cargo-semver-checks

### Actual Behaviour
```
   --> /home/USER/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-semver-checks-0.15.2/src/baseline.rs:387:31
    |
387 |             if err.class() == git2::ErrorClass::Index && err.code() == git2::ErrorCode::Locked {
    |                               ^^^^^^^^^^^^^^^^^^^^^^^ expected enum `git2::ErrorClass`, found enum `ErrorClass`
    |
    = note: enum `ErrorClass` and enum `git2::ErrorClass` have similar names, but are actually distinct types
note: enum `ErrorClass` is defined in crate `git2`
   --> /home/USER/.cargo/registry/src/github.com-1ecc6299db9ec823/git2-0.15.0/src/lib.rs:225:1
    |
225 | pub enum ErrorClass {
    | ^^^^^^^^^^^^^^^^^^^
note: enum `git2::ErrorClass` is defined in crate `git2`
   --> /home/USER/.cargo/registry/src/github.com-1ecc6299db9ec823/git2-0.16.1/src/lib.rs:225:1
    |
225 | pub enum ErrorClass {
    | ^^^^^^^^^^^^^^^^^^^
    = note: perhaps two different versions of crate `git2` are being used?

error[E0308]: mismatched types
   --> /home/USER/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-semver-checks-0.15.2/src/baseline.rs:387:72
    |
387 |             if err.class() == git2::ErrorClass::Index && err.code() == git2::ErrorCode::Locked {
    |                                                                        ^^^^^^^^^^^^^^^^^^^^^^^ expected enum `git2::ErrorCode`, found a different enum `git2::ErrorCode`
    |
    = note: enum `git2::ErrorCode` and enum `git2::ErrorCode` have similar names, but are actually distinct types
note: enum `git2::ErrorCode` is defined in crate `git2`
   --> /home/USER/.cargo/registry/src/github.com-1ecc6299db9ec823/git2-0.15.0/src/lib.rs:165:1
    |
165 | pub enum ErrorCode {
    | ^^^^^^^^^^^^^^^^^^
note: enum `git2::ErrorCode` is defined in crate `git2`
[...]

PR #318, opened 2023-01-23
update git2 dependency

demoray:
Prior to this PR, `crates-index` and this crate used different git2 versions, which causes compile errors if the project is built without '--locked'.

This PR updates git2 dependency directly, as well as updates the other dependencies using `cargo update`


demoray:
This PR fixes #317

obi1kenobi (approved):

obi1kenobi:
Thank you!

Issue #272, opened 2023-01-08
dependency crates-index with vendored-openssl feature fails with an SSL error


tonowak:
### Steps to reproduce the bug with the above code

`cargo semver-checks check-release`

### Actual Behaviour

Updating index
Error: SSL error: unknown error; class=Ssl (16)

Caused by:
    SSL error: unknown error; class=Ssl (16)

### Expected Behaviour

It should work.

#### Software version

cargo-semver-checks 0.12.1

#### Operating system

Linux 6.1.1-arch1-1

#### Command-line

```bash
/home/tonowak/.cargo/bin/cargo-semver-checks semver-checks --bugreport
```

#### cargo nightly version

```
> cargo +nightly -V
cargo 1.68.0-nightly (c994a4a63 2022-12-18)
```

#### Compile time information

- Profile: release
- Target triple: x86_64-unknown-linux-gnu
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,llvm14-builtins-abi,sse,sse2
- Host: x86_64-unknown-linux-gnu

### Additional Context

I've asked people and tried other machines and it seems to not work only on my laptop.

The issues happens exactly here: https://github.com/obi1kenobi/cargo-semver-checks/blob/f6b41f43cd8c70683e7fd901c34c76250baf3ef5/src/baseline.rs#L187

The tool works after removing `vendored-openssl` from the list of features.

Further inspection of the bug requires going into `crates-index` code and debugging it there. I'll do that in some free time -- this issue is a tracking issue for the bug.

obi1kenobi:
Should we make vendoring SSL an optional feature instead? Is that even
feasible? I haven't looked into it, just throwing an idea out there.

tonowak:
Oh, that might be a good idea. From what I remember, vendored SSL is used, because there was some problem in the CI and using it was the easiest solution, right?

tonowak:
That won't fix the underlying problem and it won't prevent an user from encountering the same bug, but I'm happy with such a solution.

obi1kenobi:
Agreed. It's a reasonable workaround until we can find the root cause.
Especially if we can catch the error when it happens and suggest disabling
the feature + link to the tracking issue, so the user doesn't have to go
look at the readme to figure out what happened and what to do about it.

staniewzki:
I have also just run into this exact error

tonowak:
[...]

PR #324, opened 2023-01-25
Add vendored-openssl feature

tonowak:
This PR adds a `vendored-openssl` default feature that enables `vendored-openssl` in `crates-index` dependency.

Because there were some issues in the CI https://github.com/obi1kenobi/cargo-semver-checks/pull/208, there was a decision to enable by default the `vendored-openssl` in `cargo-semver-checks`, but lately there have been some issues with it in some users https://github.com/obi1kenobi/cargo-semver-checks/issues/272. Now a temporary fix for those users would be to disable default features in `cargo-semver-checks` (thus disabling the `vendored-openssl` feature). Because this project is a binary, not a library, we've decided to make `vendored-openssl` a default feature so that users won't have to separately install `openssl` on their system.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,4 +43,5 @@ assert_cmd = "2.0"
 lazy_static = "1.4.0"
 
 [features]
+default = ["vendored-openssl"]
 vendored-openssl = ["crates-index/vendored-openssl"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ ignore = "0.4.18"
 clap-verbosity-flag = "2.0.0"
 log = "0.4.17"
 git2 = { version = "0.16.0", default-features = false }
-crates-index = { version = "0.19.0", features = ["vendored-openssl"] }
+crates-index = { version = "0.19.0" }
 human-panic = "1.0.3"
 bugreport = "0.5.0"
 itertools = "0.10.5"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -41,3 +41,6 @@ toml = "0.5.9"
 [dev-dependencies]
 assert_cmd = "2.0"
 lazy_static = "1.4.0"
+
+[features]
+vendored-openssl = ["crates-index/vendored-openssl"]
```
obi1kenobi, line 46: Were you still planning on making this feature be default-enabled to preserve the current behavior, or did you change your mind on it again?

tonowak: Oh sorry, I forgot to make a commit for making it a default feature. Gimme a sec.

tonowak: Thanks for the remainder. adcf9ef9d08f5c6109c16695327a68a6528c3efe

tonowak:
It seems #272 is no longer present in my local machine. I can't reproduce this issue, thus I can't test whether the feature works. :(

obi1kenobi:
Can you just update the top comment in the PR? When we merge, that's what will end up in the commit message, and it currently has incorrect information on whether this feature is on or off by default.

tonowak:
> Can you just update the top comment in the PR? When we merge, that's what will end up in the commit message, and it currently has incorrect information on whether this feature is on or off by default.

Done.

obi1kenobi (approved):

PR #329, opened 2023-01-28
Speed up tests by 10x by compiling dependencies with optimizations.

obi1kenobi:
In dev and test configurations, compile all dependencies with optimizations enabled, but still checking debug assertions and overflows.

This produces a more than 10x speedup on tests, since the tests check dozens of test crates with dozens of Trustfall queries.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,11 +46,17 @@ lazy_static = "1.4.0"
 default = ["vendored-openssl"]
 vendored-openssl = ["crates-index/vendored-openssl"]
 
-# When testing, compile all dependencies with optimizations enabled,
+# In dev and test profiles, compile all dependencies with optimizations enabled,
 # but still checking debug assertions and overflows.
 #
 # This produces a more than 10x speedup on tests, since the tests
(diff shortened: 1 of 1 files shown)


obi1kenobi:
For some reason, I couldn't get the same speedup by enabling optimizations for only `trustfall_core`, `trustfall_rustdoc`, and `trustfall-rustdoc-adapter`. It also didn't look like cargo's rebuild detection was correctly realizing which dependencies' settings had changed: adding or changing overrides for one of those packages didn't always cause a rebuild of that package.

The other thing that wasn't clear is whether overrides for `trustfall-rustdoc-adapter` would apply to _all_ major versions being included (we have 4), or whether the override was just not applying to anything and needed to instead be specified individually per-version. The [cargo docs](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides) didn't seem to say one way or the other. Possible documentation issue?

cc @epage, perhaps of interest.

obi1kenobi:
This is my first time editing profile settings, so I'd love a review from @epage just in case I messed something up.

epage (approved):
Eh, I don't touch this much either

Issue #202, opened 2022-12-02
False-positives due to renaming re-exports + name collisions in the same crate


obi1kenobi:
> Our `cargo semver-checks` CI steps currently fail for `libp2p-dcutr` and `libp2p-relay`, even though neither of them merged any changes since the last [`libp2p` `v0.50.0` release](https://github.com/libp2p/rust-libp2p/releases/tag/v0.50.0).
> 
> - `libp2p-dcutr` https://github.com/libp2p/rust-libp2p/actions/runs/3576177811/jobs/6013685638
> - `libp2p-relay` https://github.com/libp2p/rust-libp2p/actions/runs/3576177811/jobs/6013686091

_Originally posted by @mxinden in https://github.com/libp2p/rust-libp2p/issues/2647#issuecomment-1335189994_

I was able to replicate the issue locally. The generated rustdoc for `libp2p-dcutr` at first glance doesn't seem to have anything out of place. I spot-checked one of the enums and variants that reported an error, and they were present in both the baseline and current rustdoc with identical IDs and reasonable-looking data.

This issue needs further triage.

obi1kenobi:
Unfortunately, I found that this is because of a combination of renaming re-exports causing confusion between multiple enums with the same name. For example, a `pub enum UpgradeError` exists in all of:
```
protocols/dcutr/src/protocol/outbound.rs
protocols/dcutr/src/protocol/inbound.rs
protocols/dcutr/src/behavior.rs
```
The false-positives happen because the lint tries to match up all the variants across different enums, which obviously isn't going to do the right thing.

`cargo-semver-checks` doesn't currently handle re-export renames (and re-exports are already tricky because of the possibility of having infinitely many re-exported names for each item: https://github.com/obi1kenobi/cargo-semver-checks/issues/152 ), so this isn't a trivial fix.

That said, it's still important to fix and this should be sufficient motivation to come up with a permanent solution to the re-exports problem :)

obi1kenobi:
It's tempting to use the `canonical_path` edge instead of `importable_path` edge in the affected lints. Right now, this wouldn't work, since re-exported items do not currently have a `canonical_path` available, since that path uses the rustdoc index information which currently does not include re-exported items.

Relevant rustdoc issues:
https://github.com/rust-lang/rust/issues/93522
https://github.com/rust-lang/rust/issues/99513

PR #330, opened 2023-01-28
Detect and don't report non-breaking renaming re-exports.

### LINKED ISSUE #42: Handle imports that rename the imported item

### LINKED ISSUE #288: Tracking issue: import handling revamp
- [ ] support type aliases: #215
- [ ] support renaming re-exports: #42 and #202
- [x] don't crash on recursive re-exports: #152 and #287
- [x] support glob imports and glob re-exports: #34

Type alias support is more valuable than renaming re-exports, per https://github.com/libp2p/rust-libp2p/pull/2647#issuecomment-1343467075

obi1kenobi:
Only use importable paths to match items across versions.

Declared names (e.g. `Foo` in `pub struct Foo;`) are subject to change in renaming `pub use` and `pub type` statements (e.g. `pub use Foo as Bar`), so relying on them to match items across versions means we miss renames. The importable paths provided by `trustfall_rustdoc v0.8.2+` account for this, and list all valid paths under which the item can be imported. These paths are the only way we can match items across versions.

The above logic change needed to be applied to all currently-written lints, and will need to also apply to all future lints as well. Heads-up for @SmolSir @tonowak @staniewzki @mgr0dzicki.

The new logic is also somewhat more expensive, since it increases the search space and eliminates some of the pruning that filtering on the name could get us. This is unfortunate, but unavoidable at the moment — we don't want to sacrifice correctness for speed. I'll focus on perf improvements next and I'm confident we can win back all the lost performance and then some.

Fixes #288, #215, #202.
Fixes #42.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,7 @@ jobs:
       - lint
       - rust-tests
       - run-on-rust-libp2p
+      - run-on-libp2p-dcutr-relay
       - run-on-core-graphics
       - run-on-bevy-core
       - run-on-bevy-gltf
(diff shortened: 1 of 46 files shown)


obi1kenobi:
Still TODO here:
- [ ] add all crates that had these false-positives into the GitHub Actions regression suite
- [ ] measure the perf impact of the new lint changes

I think the impact won't be spectacular on most crates, but still better to measure and make an explicit decision based on data. Hope is not a strategy.

epage (approved):
Didn't get into too much of the details but glad to see this done

obi1kenobi:
An overnight perf run didn't show any significant perf changes one way or the other on even the biggest crates.

Issue #223, opened 2022-12-17
Add `witness` crates that demonstrate how each breaking change is breaking


obi1kenobi:
What is and is not a breaking change is really tricky, and #5 contains many examples of breaking changes that are very much not obvious. We're already writing test crates that exemplify the breaking changes when implementing new lints: we add an `old` and `new` crate, where the `new` crate makes a breaking change.

It would be great to include a third `witness` crate that compiles fine against the `old` crate, and fails to compile due to the breaking change. This would complete the example of why that change is breaking, and would be constructive proof that we are indeed reporting an actually-breaking change.

Mentioned in #203.

Diggsey:
It would be cool (and not that hard?) to automatically generate a witness crate for the *actual* issue identified, not just a generic witness crate for that type of issue.

This could also be a good way to detect false-positives in advance, since it would be hard to write witness generation for a flawed lint... Certianly it would at least make you think about edge cases more.

Diggsey:
This could also inform severity - eg. losing `Sync` trait might be considered a less severe breaking change if the type was never constructable. (Not being constructable would impact what kind of witness crate you could generate).

obi1kenobi:
> It would be cool (and not that hard?) to automatically generate a witness crate for the _actual_ issue identified, not just a generic witness crate for that type of issue.

Cool idea! Would you perhaps be interested in prototyping it? It sounds like you have a viable approach in mind already.

> edit: This could also inform severity - eg. losing `Sync` trait might be considered a less severe breaking change if the type was never constructable. (Not being constructable would impact what kind of witness crate you could generate).

It's certainly possible that losing `Sync` might not be breaking at all, even: if the type isn't directly constructible with a literal or any function that returns a value of that type. Any cascading losses of `Sync` on other types causing breaking changes can arguably be blamed on those types' loss of `Sync` in turn.

PR #337, opened 2023-01-31
Test case: replace type with a more generic one without major change.

obi1kenobi:


obi1kenobi:
@epage could you check if you think this test crate contains any breaking changes, if you have a second?

To me, it seems like the answer is "no" but I'd love to check if there's some edge case I've missed.

If "no" is the right answer, then this is the remaining class of false-positives vaguely related to the import system. This isn't resolved by the changes in 0.17 since the `pub type` here is *not* just a rename and re-export — it's a new type that isn't semantically identical to the underlying.

test_crates/type_alias_over_more_generic_underlying/new/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+
+pub type OriginalStructMixOfFields = NewStructMixOfFields<()>;
+
+// ---
+
+pub struct NewStructConstructible<T> {
+    pub x: T,
+}
+
+pub type OriginalStructConstructible = NewStructConstructible<i64>;
```
epage, line 39: I think this is the case you need to look into of whether it is semver compatible.  I have a vague recollection of someone warning about replacing a type with `type` and I think it was with `{}` construction syntax.

obi1kenobi: This would be another great use of witness crates (#223), except to demonstrate that some syntax _does not_ break, instead of demonstrating that it breaks.

obi1kenobi: I either found the issue you had in mind or something related: #338

TL;DR: `pub type` replacement of a constructible unit or tuple struct is breaking, since the unit/tuple constructor is not usable on the type alias.

QuineDot: [Trait bounds on type aliases aren't checked.](https://github.com/rust-lang/rust/issues/21903)  This is linted (sometimes an error, sometimes a warning), but the lint has false positives.

Additionally -- and as far as I know this is not documented anywhere -- [while lifetime constraints aren't enforced per se, they do have semantic impact.](https://github.com/rust-lang/rust/issues/100270#issuecomment-1293331893)  Therefore, not having the same bounds (or lack of bounds) on the `type` as are on the original struct is a breaking change.  But if it doesn't cause local breakage, people who get the lint warning are likely to apply the suggestion and remove the semantically relevant bounds.

I suspect most people don't know the `dyn Trait` lifetime default mechanism in the first place, and [RFC 2093](https://rust-lang.github.io/rfcs/2093-infer-outlives.html#impact-on-semver) made the potential for breakage a lot more nuanced to boot, so that may be an area to look into independently of `type` aliases as well.

Issue #214, opened 2022-12-09
Make CI runs faster


thomaseizinger:
### Describe your use case

We are using this in CI and it contributes about ~1-2min to the CI run of each crate despite:

- `cargo semver-checks` being installed
- `cargo doc` being cached

See https://github.com/thomaseizinger/rust-libp2p/actions/runs/3653859145/jobs/6173748557 for an example.

Is there a way for how this can be improved? When I run it locally, it is a lot faster (after the first invocation) so I am guessing there is a cache somewhere? Which files do I need to cache?

### Describe the solution you'd like

`cargo semver-checks` to finish within seconds, not minutes.

obi1kenobi:
I'm not familiar with your setup for caching `cargo doc` — does it work for only the current state of the package, or does it also cache the `cargo doc` run for the baseline (crates.io) version as well? Based on the log, it seems like the actual checking of the rustdoc files finished in 0.044s, so the slow parts are building one or both of the rustdoc JSON files.

#208 should merge very soon, and might make it easier to not have to cache the `cargo install` step.

Since the baseline version isn't changing all that often (especially not with every PR), it should be possible to cache the baseline rustdoc JSON file across runs. This is [how the GitHub Action builds that file](https://github.com/obi1kenobi/cargo-semver-checks-action/blob/main/action.yml#L38-L68), and in principle it could be improved to cache the file as well. `cargo-semver-checks` can be directed to use a pre-generated baseline rustdoc JSON file with the `--baseline-rustdoc <file>` option.

After running `cargo check --all-features` or `cargo test --all-features`, building the rustdoc JSON for the current version should be quite fast since everything is already mostly checked and built. So running `cargo-semver-checks` in the same runner where the tests already ran would probably make it much faster.

In general, I believe your `cargo-semver-checks` setup is the most sophisticated one in existence (at least that I know of). If you're interested, I'd be thrilled to work with you on building all these capabilities into [the cargo-semver-checks GitHub Action](https://github.com/obi1kenobi/cargo-semver-checks-action) itself rather than each of us maintaining a separate workflow.

thomaseizinger:
My workflow doesn't build it for the baseline version, no. I thought that one was downloaded from docs.rs or something?

Does it build the baseline version also locally?

I am happy to collaborate on an/the action. Fitting that one out with some pre-configured caching would be really nice actually :)

#208 is nice to see!

obi1kenobi:
[...]

PR #339, opened 2023-02-02
Cache generated rustdoc for registry baselines

thomaseizinger:
This implements the suggestion discussion in https://github.com/obi1kenobi/cargo-semver-checks/issues/214. It turned out surprisingly simple :)

### DIFF
--- a/src/baseline.rs
+++ b/src/baseline.rs
@@ -348,9 +348,9 @@ impl BaselineLoader for RegistryBaseline {
             return Ok(cached_rustdoc);
         }
 
-        let base_root = self.target_root.join(crate_identifier);
-        std::fs::create_dir_all(&base_root)?;
-        let manifest_path = base_root.join("Cargo.toml");
+        let build_dir = self.target_root.join(crate_identifier);
+        std::fs::create_dir_all(&build_dir)?;
+        let manifest_path = build_dir.join("Cargo.toml");
 
         let crate_baseline = crate_
             .versions()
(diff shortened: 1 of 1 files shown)


thomaseizinger:
I've tested it locally on the `rust-libp2p` repository. It works after a `cargo clean` and cleans up after itself. A subsequent run detects the cached file and reuses it which speeds up things a fair bit.

obi1kenobi (approved):

obi1kenobi:
@thomaseizinger if I were to release 0.18 with this caching change soon, when would be the best time for you? I'd love to have you onboard to test it in "the real world" as soon as it comes out, in case something doesn't behave quite as well as we had hoped.

thomaseizinger:
> @thomaseizinger if I were to release 0.18 with this caching change soon, when would be the best time for you? I'd love to have you onboard to test it in "the real world" as soon as it comes out, in case something doesn't behave quite as well as we had hoped.

We can test it as soon as there is a binary. Why don't we release `0.18.0-alpha` and we use that in our CI for a bit?

thomaseizinger:
> > @thomaseizinger if I were to release 0.18 with this caching change soon, when would be the best time for you? I'd love to have you onboard to test it in "the real world" as soon as it comes out, in case something doesn't behave quite as well as we had hoped.
> 
> We can test it as soon as there is a binary. Why don't we release `0.18.0-alpha` and we use that in our CI for a bit?

I am gonna be on NZ timezone for the next two weeks. With a pre-release, I can adjust our CI to create a cache on master and use that on PR workflow runs and play around what happens when we nuke the cache etc.

tonowak:
@obi1kenobi it would be good to first merge #340 before creating the alpha release, since it touched some of this PR's code.

PR #340, opened 2023-02-02
Refactor of baseline.rs for easier next additive changes

tonowak:
After this PR, I plan to generate the current&baseline rustdoc from a local copy with the same method as the registry baseline. It will allow specifying features and it won't touch the project's `Cargo.lock`.

This PR is mostly moving functions and splitting functions into multiple smaller ones -- hopefully no existing logic has been changed. It prepares the code for the new current&baseline rustdoc generation (some new functions are created, but not implemented).

### DIFF
--- a/src/baseline.rs
+++ b/src/baseline.rs
@@ -1,4 +1,5 @@
 use std::collections::BTreeSet;
+use std::path::{Path, PathBuf};
 
 use anyhow::Context as _;
 use crates_index::Crate;
@@ -7,6 +8,180 @@ use crate::dump::RustDocCommand;
 use crate::util::slugify;
 use crate::GlobalConfig;
 
+fn get_all_crate_features_from_registry(crate_: &crates_index::Version) -> Vec<String> {
+    // Sometimes crates ship types with fields or variants that are included
+    // only when certain features are enabled.
+    //
+    // By default, we want to generate rustdoc with `--all-features`,
+    // but that option isn't available here so we have to implement it ourselves.
+    // TODO: this comment is not yet true, but it'll be in some next PR. Then move the whole comment
+    // outside and remove this TODO.
+    //
+    // Implicit features from optional dependencies have to be added separately
+    // from regular features: https://github.com/obi1kenobi/cargo-semver-checks/issues/265
+    let mut implicit_features: BTreeSet<_> = crate_
+        .dependencies()
+        .iter()
+        .filter_map(|dep| dep.is_optional().then_some(dep.name()))
+        .map(|x| x.to_string())
+        .collect();
+    for feature_defn in crate_.features().values().flatten() {
+        // "If you specify the optional dependency with the dep: prefix anywhere
+        //  in the [features] table, that disables the implicit feature."
+        // https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies
+        if let Some(optional_dep) = feature_defn.strip_prefix("dep:") {
+            implicit_features.remove(optional_dep);
+        }
+    }
+    let regular_features: BTreeSet<_> = crate_.features().keys().cloned().collect();
+    let mut all_crate_features = implicit_features;
+    all_crate_features.extend(regular_features);
+    all_crate_features.into_iter().collect()
+}
+
+#[allow(unused_variables)]
+fn get_all_crate_features_from_manifest(path: &Path) -> Vec<String> {
+    unimplemented!()
+}
+
+#[allow(dead_code)]
+enum CrateSource<'a> {
+    Registry { crate_: &'a crates_index::Version },
+    ManifestPath { path: &'a Path, name: String },
+}
+
+/// To get the rustdoc of the baseline, we first create a placeholder project somewhere
+/// with the baseline as a dependency, and run `cargo rustdoc` on it.
+fn create_placeholder_rustdoc_manifest(crate_source: &CrateSource) -> cargo_toml::Manifest<()> {
+    use cargo_toml::*;
+
+    Manifest::<()> {
+        package: {
+            let mut package = Package::new("rustdoc", "0.0.0");
+            package.publish = Inheritable::Set(Publish::Flag(false));
+            Some(package)
+        },
+        workspace: Some(Workspace::<()>::default()),
+        lib: {
+            let product = Product {
+                path: Some("lib.rs".to_string()),
+                ..Product::default()
+            };
+            Some(product)
+        },
+        dependencies: {
+            let project_with_features: DependencyDetail = match crate_source {
+                CrateSource::Registry { crate_ } => DependencyDetail {
+                    // We need the *exact* version as a dependency, or else cargo will
+                    // give us the latest semver-compatible version which is not we want.
+                    // Fixes: https://github.com/obi1kenobi/cargo-semver-checks/issues/261
+                    version: Some(format!("={}", crate_.version())),
+                    features: get_all_crate_features_from_registry(crate_),
+                    ..DependencyDetail::default()
+                },
+                CrateSource::ManifestPath { path, name: _ } => DependencyDetail {
+                    path: Some(
+                        // The Manifest will be saved in some other directory,
+                        // so for convenience, we're using absolute paths.
+                        path.canonicalize()
+                            .expect("Failed to canonicalize Manifest path")
+                            .to_str()
+                            .expect("Manifest path is not valid UTF-8")
+                            .to_string(),
+                    ),
+                    features: get_all_crate_features_from_manifest(path),
+                    ..DependencyDetail::default()
+                },
+            };
+            let mut deps = DepsSet::new();
+            deps.insert(
+                match crate_source {
+                    CrateSource::Registry { crate_ } => crate_.name().to_string(),
+                    CrateSource::ManifestPath { name, path: _ } => name.to_string(),
+                },
+                Dependency::Detailed(project_with_features),
+            );
+            deps
+        },
+        ..Default::default()
+    }
+}
+
+fn save_placeholder_rustdoc_manifest(
+    placeholder_build_dir: &Path,
+    placeholder_manifest: cargo_toml::Manifest<()>,
+) -> anyhow::Result<PathBuf> {
+    std::fs::create_dir_all(placeholder_build_dir)?;
+    let placeholder_manifest_path = placeholder_build_dir.join("Cargo.toml");
+
+    // Possibly fixes https://github.com/libp2p/rust-libp2p/pull/2647#issuecomment-1280221217
+    let _: std::io::Result<()> = std::fs::remove_file(placeholder_build_dir.join("Cargo.lock"));
+
+    std::fs::write(
+        &placeholder_manifest_path,
+        toml::to_string(&placeholder_manifest)?,
+    )?;
+    std::fs::write(placeholder_build_dir.join("lib.rs"), "")?;
+    Ok(placeholder_manifest_path)
+}
+
+fn generate_rustdoc(
+    config: &mut GlobalConfig,
+    rustdoc: &RustDocCommand,
+    target_root: PathBuf,
+    crate_source: CrateSource,
+) -> anyhow::Result<PathBuf> {
+    let (name, version) = match crate_source {
+        CrateSource::Registry { crate_ } => (crate_.name().to_string(), crate_.version()),
+        CrateSource::ManifestPath { .. } => unimplemented!(),
+    };
+    let (build_dir, cache_dir) = match crate_source {
+        CrateSource::Registry { .. } => {
+            let crate_identifier = format!("registry-{}-{}", slugify(&name), slugify(version));
+            let cache_dir = target_root
+                .join("cache")
+                .join(format!("{crate_identifier}.json"));
+
+            // We assume that the generated rustdoc is untouched.
+            // Users should run cargo-clean if they experience any anomalies.
+            if cache_dir.exists() {
+                return Ok(cache_dir);
+            }
+
+            let build_dir = target_root.join(crate_identifier);
+            (build_dir, cache_dir)
+        }
+        CrateSource::ManifestPath { .. } => {
+            unimplemented!()
+        }
+    };
+
+    let placeholder_manifest = create_placeholder_rustdoc_manifest(&crate_source);
+    let placeholder_manifest_path =
+        save_placeholder_rustdoc_manifest(build_dir.as_path(), placeholder_manifest)?;
+
+    config.shell_status("Parsing", format_args!("{name} v{version} (baseline)"))?;
+    // TODO: replace (baseline) with something else
+
+    let rustdoc_path = rustdoc.dump(
+        placeholder_manifest_path.as_path(),
+        Some(&format!("{name}@{version}")),
+        false,
+    )?;
+
+    match crate_source {
+        CrateSource::Registry { .. } => {
+            // Clean up after ourselves.
+            std::fs::create_dir_all(&cache_dir)?;
+            std::fs::copy(rustdoc_path, &cache_dir)?;
+            std::fs::remove_dir_all(build_dir)?;
+        }
+        CrateSource::ManifestPath { .. } => {}
+    };
+
+    Ok(cache_dir)
+}
+
 pub(crate) trait BaselineLoader {
     fn load_rustdoc(
         &self,
@@ -14,15 +189,15 @@ pub(crate) trait BaselineLoader {
         rustdoc: &RustDocCommand,
         name: &str,
         version_current: Option<&semver::Version>,
-    ) -> anyhow::Result<std::path::PathBuf>;
+    ) -> anyhow::Result<PathBuf>;
 }
 
 pub(crate) struct RustdocBaseline {
-    path: std::path::PathBuf,
+    path: PathBuf,
 }
 
 impl RustdocBaseline {
-    pub(crate) fn new(path: std::path::PathBuf) -> Self {
+    pub(crate) fn new(path: PathBuf) -> Self {
         Self { path }
     }
 }
@@ -34,14 +209,14 @@ impl BaselineLoader for RustdocBaseline {
         _rustdoc: &RustDocCommand,
         _name: &str,
         _version_current: Option<&semver::Version>,
-    ) -> anyhow::Result<std::path::PathBuf> {
+    ) -> anyhow::Result<PathBuf> {
         Ok(self.path.clone())
     }
 }
 
 pub(crate) struct PathBaseline {
-    root: std::path::PathBuf,
-    lookup: std::collections::HashMap<String, (String, std::path::PathBuf)>,
+    root: PathBuf,
+    lookup: std::collections::HashMap<String, (String, PathBuf)>,
 }
 
 impl PathBaseline {
@@ -74,7 +249,7 @@ impl BaselineLoader for PathBaseline {
         rustdoc: &RustDocCommand,
         name: &str,
         _version_current: Option<&semver::Version>,
-    ) -> anyhow::Result<std::path::PathBuf> {
+    ) -> anyhow::Result<PathBuf> {
         let (version, manifest_path) = self
             .lookup
             .get(name)
@@ -153,7 +328,7 @@ impl BaselineLoader for GitBaseline {
         rustdoc: &RustDocCommand,
         name: &str,
         version_current: Option<&semver::Version>,
-    ) -> anyhow::Result<std::path::PathBuf> {
+    ) -> anyhow::Result<PathBuf> {
         self.path
             .load_rustdoc(config, rustdoc, name, version_current)
     }
@@ -174,7 +349,7 @@ fn bytes2str(b: &[u8]) -> &std::ffi::OsStr {
 }
 
 pub(crate) struct RegistryBaseline {
-    target_root: std::path::PathBuf,
+    target_root: PathBuf,
     version: Option<semver::Version>,
     index: crates_index::Index,
 }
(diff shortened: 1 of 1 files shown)


src/baseline.rs
```diff
@@ -7,22 +8,196 @@ use crate::dump::RustDocCommand;
...
 use crate::GlobalConfig;
 
+fn get_all_crate_features_from_registry(crate_: &crates_index::Version) -> Vec<String> {
+    // Sometimes crates ship types with fields or variants that are included
+    // only when certain features are enabled.
+    //
+    // By default, we want to generate rustdoc with `--all-features`,
+    // but that option isn't available here so we have to implement it ourselves.
+    // TODO: this comment is not yet true, but it'll be in some next PR. Then move the whole comment
+    // outside and remove this TODO.
```
obi1kenobi, line 18: What's not true about this? The code seems to me to do what it says at the moment?

src/baseline.rs
```diff
@@ -7,22 +8,196 @@ use crate::dump::RustDocCommand;
...
+    all_crate_features.into_iter().collect()
+}
+
+#[allow(unused_variables)]
+fn get_all_crate_features_from_manifest(path: &Path) -> Vec<String> {
+    unimplemented!()
+}
+
+#[allow(dead_code)]
+enum CrateSource<'a> {
```
obi1kenobi, line 48:
```suggestion
#[derive(Debug, Clone)]
enum CrateSource<'a> {
```
Unless there's reason not to? Almost all types should be `Debug` and (slightly fewer of them) `Clone`.

src/baseline.rs
```diff
@@ -7,22 +8,196 @@ use crate::dump::RustDocCommand;
...
+            let project_with_features: DependencyDetail = match crate_source {
+                CrateSource::Registry { crate_ } => DependencyDetail {
+                    // We need the *exact* version as a dependency, or else cargo will
+                    // give us the latest semver-compatible version which is not we want.
+                    // Fixes: https://github.com/obi1kenobi/cargo-semver-checks/issues/261
+                    version: Some(format!("={}", crate_.version())),
+                    features: get_all_crate_features_from_registry(crate_),
+                    ..DependencyDetail::default()
+                },
+                CrateSource::ManifestPath { path, name: _ } => DependencyDetail {
```
obi1kenobi, line 82:
```suggestion
                CrateSource::ManifestPath { path, .. } => DependencyDetail {
```
Perhaps a bit easier to read? Your call.

src/baseline.rs
```diff
@@ -7,22 +8,196 @@ use crate::dump::RustDocCommand;
...
+    )?;
+
+    match crate_source {
+        CrateSource::Registry { .. } => {
+            // Clean up after ourselves.
+            std::fs::create_dir_all(&cache_dir)?;
+            std::fs::copy(rustdoc_path, &cache_dir)?;
+            std::fs::remove_dir_all(build_dir)?;
+        }
+        CrateSource::ManifestPath { .. } => {}
```
obi1kenobi, line 179:
```suggestion
        CrateSource::ManifestPath { .. } => {
            // Explain why we don't do caching / copying / cleaning here?
        }
```

obi1kenobi (commented):
Looks good overall, some minor suggestions.

src/baseline.rs
tonowak: Oh, indeed, it is true. So the only thing left is to somehow lift the comment up so it references both the `get_all_crate_features_*` functions.

obi1kenobi: Move the two functions into their own submodule and make this kind of thing a comment on the module itself?

src/baseline.rs
```diff
@@ -7,22 +8,203 @@ use crate::dump::RustDocCommand;
...
+    )?;
+
+    match crate_source {
+        CrateSource::Registry { .. } => {
+            // Clean up after ourselves.
+            std::fs::create_dir_all(&cache_dir).with_context(|| "failed to create cache dir")?;
+            let cache_rustdoc_path = cache_dir.join("rustdoc.json");
+            std::fs::copy(rustdoc_path, &cache_rustdoc_path)
+                .with_context(|| "failed to copy cache dir")?;
+            std::fs::remove_dir_all(build_dir).with_context(|| "failed to remove build dir")?;
```
tonowak, line 183: Do all of the `with_context` make sense? I'm not yet proficient in using Rust errors.

src/baseline.rs
tonowak: Done in 9cc50d33d5c644c50133388c122b59c3dcd96779.

src/baseline.rs
tonowak: Done in f54a285b77811935d4ecdee3ab128b1575d980f5.

tonowak:
Should be ready to merge.

src/baseline.rs
```diff
@@ -7,22 +7,210 @@ use crate::dump::RustDocCommand;
...
+            Ok(cache_rustdoc_path)
+        }
+        CrateSource::ManifestPath { .. } => {
+            // We don't do any caching here -- since the crate is saved locally,
+            // it could be modified by the user since it was cached.
+            Ok(rustdoc_path)
```
thomaseizinger, line 193: We can never get here or am I missing something? With the previous paths being unimplemented, this function will only work for `Registry`.

src/baseline.rs
```diff
@@ -7,22 +7,210 @@ use crate::dump::RustDocCommand;
...
+    /// Sometimes crates ship types with fields or variants that are included
+    /// only when certain features are enabled.
+    ///
+    /// By default, we want to generate rustdoc with `--all-features`,
+    /// but that option isn't available here so we have to implement it ourselves.
+
+    pub(crate) fn get_all_crate_features_from_registry(
+        crate_: &crates_index::Version,
+    ) -> Vec<String> {
+        //
```
obi1kenobi, line 20:
```suggestion
```

src/baseline.rs
```diff
@@ -7,22 +7,210 @@ use crate::dump::RustDocCommand;
...
+        CrateSource::ManifestPath { .. } => unimplemented!(),
+    };
+    let (build_dir, cache_dir) = match crate_source {
+        CrateSource::Registry { .. } => {
+            let crate_identifier = format!("registry-{}-{}", slugify(&name), slugify(version));
+            let cache_dir = target_root.join("cache").join(&crate_identifier);
+
+            // We assume that the generated rustdoc is untouched.
+            // Users should run cargo-clean if they experience any anomalies.
+            if cache_dir.exists() {
```
thomaseizinger, line 153: We need to check for existence of the JSON file, not the directory.

src/baseline.rs
```diff
@@ -7,22 +7,210 @@ use crate::dump::RustDocCommand;
 use crate::util::slugify;
 use crate::GlobalConfig;
 
+mod crate_features {
+    /// Sometimes crates ship types with fields or variants that are included
+    /// only when certain features are enabled.
+    ///
+    /// By default, we want to generate rustdoc with `--all-features`,
+    /// but that option isn't available here so we have to implement it ourselves.
```
obi1kenobi, line 15:
```suggestion
    /// but that option isn't available outside of the current crate,
    /// so we have to implement it ourselves.
```

thomaseizinger (commented):
I only looked at the caching code, nothing else but I think this may still need some work!

src/baseline.rs
```diff
@@ -7,22 +7,210 @@ use crate::dump::RustDocCommand;
...
+                    ..DependencyDetail::default()
+                },
+                CrateSource::ManifestPath { path, .. } => DependencyDetail {
+                    path: Some(
+                        // The Manifest will be saved in some other directory,
+                        // so for convenience, we're using absolute paths.
+                        path.canonicalize()
+                            .with_context(|| "failed to canonicalize Manifest path")?
+                            .to_str()
+                            .with_context(|| "Manifest path is not valid UTF-8")?
```
obi1kenobi, line 95:
```suggestion
                        // The manifest will be saved in some other directory,
                        // so for convenience, we're using absolute paths.
                        path.canonicalize()
                            .with_context(|| "failed to canonicalize manifest path")?
                            .to_str()
                            .with_context(|| "manifest path is not valid UTF-8")?
```

src/baseline.rs
```diff
@@ -7,22 +7,210 @@ use crate::dump::RustDocCommand;
...
+    let placeholder_manifest_path = placeholder_build_dir.join("Cargo.toml");
+
+    // Possibly fixes https://github.com/libp2p/rust-libp2p/pull/2647#issuecomment-1280221217
+    let _: std::io::Result<()> = std::fs::remove_file(placeholder_build_dir.join("Cargo.lock"));
+
+    std::fs::write(
+        &placeholder_manifest_path,
+        toml::to_string(&placeholder_manifest)?,
+    )
+    .with_context(|| "failed to write placeholder Manifest")?;
```
obi1kenobi, line 130:
```suggestion
    .with_context(|| "failed to write placeholder manifest")?;
```

src/baseline.rs
tonowak: Thanks for reviewing!
Yes, the current code only works for Registry -- this PR is a preparation for implementing the Manifest path (in the next PR I'll implement it). I've split it to make the review easier.

src/baseline.rs
```diff
@@ -338,21 +456,7 @@ impl BaselineLoader for RegistryBaseline {
...
 
-        let crate_identifier = format!("registry-{}-{}", slugify(name), slugify(&base_version));
-
-        let cache_dir = self.target_root.join("cache");
-        let cached_rustdoc = cache_dir.join(format!("{crate_identifier}.json"));
-
-        // We assume that the generated rustdoc is untouched. Users should run cargo-clean if they experience any anomalies.
-        if cached_rustdoc.exists() {
-            return Ok(cached_rustdoc);
-        }
```
obi1kenobi, line 349: @tonowak I don't think you ported this bit correctly. The old logic here is correct but the new logic added in this PR has several different bugs.

src/baseline.rs
thomaseizinger: Got it, so we only ever call this with the `Registry` variant?

src/baseline.rs
tonowak: From what I remember, I've changed the code a bit so that the cached rustdoc is in the format `cache/registry-xxx-yyy/rustdoc.json` instead of `cache/registry-xxx-yyy.json`, with the reasoning that the code would be a bit easier to understand after the changes I've made (because of the `let (x, y) = z`), and I won't deny it -- this change is mostly stylistic. I think the code you've pointed out after my changes is correct -- can you confirm?

tonowak: Oh, I see that @obi1kenobi has also commented this code. I'll look into it in detail.

tonowak: I'll revert to how it was before.

obi1kenobi: I don't think the directory-per-file layout is desirable, and I since we already have something that we think works and that we want to test in the real world, I think we probably don't want to change the behavior here and now.

src/baseline.rs
obi1kenobi: Yeah, they are fine. I just un-capitalized `manifest` throughout since it isn't a name.

src/baseline.rs
thomaseizinger: Your PR description said "no functionality was changed" which is why I commented :)

I think a flatter file hierarchy is preferable but I am obviously biased 😁

Whatever you go with, test it locally please that it still works as we don't have any automated tests for this!

src/baseline.rs
tonowak: Yes. Sorry for not making it clear in the PR

src/baseline.rs
tonowak: 42f0fb99e99f332da156f6dac7d10f663dd0d0bb reverts it. Thanks for the help.

tonowak: > Your PR description said "no functionality was changed" which is why I commented :)

Right, that totally was my mistake. I'm still learning :)

src/baseline.rs
```diff
@@ -7,22 +7,210 @@ use crate::dump::RustDocCommand;
...
+        },
+        ..Default::default()
+    })
+}
+
+fn save_placeholder_rustdoc_manifest(
+    placeholder_build_dir: &Path,
+    placeholder_manifest: cargo_toml::Manifest<()>,
+) -> anyhow::Result<PathBuf> {
+    std::fs::create_dir_all(placeholder_build_dir).with_context(|| "failed to create build dir")?;
```
thomaseizinger, line 120: If you don't allocate, you can just use `.context`. The `with` variant is for when you allocate to only do it in case of an error.

src/baseline.rs
```diff
@@ -7,22 +7,210 @@ use crate::dump::RustDocCommand;
...
+                },
+                Dependency::Detailed(project_with_features),
+            );
+            deps
+        },
+        ..Default::default()
+    })
+}
+
+fn save_placeholder_rustdoc_manifest(
```
thomaseizinger, line 116: This could be an extension trait on `cargo_toml::Manifest`:

```rust
trait ManifestExt {
    fn save_to(&self, path: &Path) -> io::Result<>;
}

impl ManifestExt for cargo_toml::Manifest { }
```

Just a different way of doing things though :)

src/baseline.rs
```diff
@@ -7,22 +7,210 @@ use crate::dump::RustDocCommand;
...
+fn generate_rustdoc(
+    config: &mut GlobalConfig,
+    rustdoc: &RustDocCommand,
+    target_root: PathBuf,
+    crate_source: CrateSource,
+) -> anyhow::Result<PathBuf> {
+    let (name, version) = match crate_source {
+        CrateSource::Registry { crate_ } => (crate_.name().to_string(), crate_.version()),
+        CrateSource::ManifestPath { .. } => unimplemented!(),
+    };
```
thomaseizinger, line 145: Should maybe be a function on `CrateSource`.

src/baseline.rs
```diff
@@ -201,76 +389,6 @@ impl RegistryBaseline {
     }
 }
 
-/// To get the rustdoc of the baseline, we first create a placeholder project somewhere
-/// with the baseline as a dependency, and run `cargo rustdoc` on it.
-fn create_rustdoc_manifest_for_crate_version(
```
thomaseizinger, line 206: Are we just moving this fn around? If so, we could reduce the diff of this PR and make review easier.

thomaseizinger (commented):
Took a deeper look, some suggestions :)

src/baseline.rs
obi1kenobi: Extension traits always felt a bit too magical for my taste, except in cases like `itertools` where the extension methods are truly general-purpose and applicable everywhere. I appreciate you highlighting the option, though!

src/baseline.rs
obi1kenobi: > we don't have any automated tests for this

By a happy accident, we actually do — and they caught a minor bug:
https://github.com/obi1kenobi/cargo-semver-checks/actions/runs/4079374830/jobs/7030684022

That test runs effectively the same command over the same crate in two different ways (since they were broken due to separate reasons), and it happens to run back to back in the same runner. So we can see that the caching worked (yay! 🙌) but we can also see that when the cache gets hit, the `(baseline)` printout is just inexplicably gone (confusing 🤨).
image

Let's add a user-facing message there that indicates the cache hit. Suggestion: `Parsing libp2p-core v0.37.0 (baseline, cached)`

obi1kenobi: @thomaseizinger you might find the subject crate in the test familiar 😁

obi1kenobi:
It's the middle of the night for @tonowak right now, so I'll publish `0.18.0-alpha` without this PR for now and we can publish more alphas if needed in the future.

obi1kenobi:
I just realized -- @thomaseizinger did you need me to generate prebuilt binaries? I published the pre-release version manually to avoid spamming everyone watching the repo, but that also means no prebuilt binaries since the GitHub workflow didn't run.

src/baseline.rs
tonowak: Oh, what do you mean? Is there a way to tell GitHub's diff that some code has moved? Or do you propose to not move this function?

src/baseline.rs
tonowak: Moved in 69fa5a846327736f9cb272892140b7951efa71d3.

src/baseline.rs
tonowak: Done in ff7c57118d70ef99a155b120885e933d0299be2d. Do I understand correctly that e.g. `.with_context(|| anyhow::format_err!("{} not found in registry", name))` allocates, so it should be in `with_context`?

src/baseline.rs
tonowak: > Let's add a user-facing message there that indicates the cache hit. Suggestion: `Parsing libp2p-core v0.37.0 (baseline, cached)`

Done in 54e8b67b3fb46fc69fbb24207ecf102ba38f59c6.

src/baseline.rs
tonowak: @obi1kenobi do I understand correctly that you propose to *not* implement this change? I'm confused.

tonowak: I think implementing this change would make the code a bit harder to understand -- this code removes a `Cargo.lock` in the same directory, which is a fix specific only to the current use case, but that change would suggest that it is a function for general use.

src/baseline.rs
thomaseizinger: Yes!

src/baseline.rs
thomaseizinger: I would not move it unless strictly necessary if the PR also contains functional changes.

tonowak: I see. Since this PR is pretty much reviewed, I think I prefer to leave it as it is currently and in my next PRs I'll try to do it like you suggest (splitting the PRs into functional changes and just moving code around).

src/baseline.rs
obi1kenobi: Yes, I wouldn't do it.

src/baseline.rs
```diff
@@ -7,22 +7,232 @@ use crate::dump::RustDocCommand;
...
+                    features: crate_features::get_all_crate_features_from_manifest(path),
+                    ..DependencyDetail::default()
+                },
+            };
+            let mut deps = DepsSet::new();
+            deps.insert(
+                match crate_source {
+                    CrateSource::Registry { crate_ } => crate_.name().to_string(),
+                    CrateSource::ManifestPath { name, path: _ } => name.to_string(),
+                },
```
obi1kenobi: This is the same function as defined above. Assuming you accept my suggestion there, this could be just:
```suggestion
                crate_source.name().to_string(),
```

src/baseline.rs
```diff
@@ -7,22 +7,232 @@ use crate::dump::RustDocCommand;
...
+            Self::ManifestPath { .. } => unimplemented!(),
+        }
+    }
+
+    fn get_version(&self) -> String {
+        match self {
+            Self::Registry { crate_ } => crate_.version().to_string(),
+            Self::ManifestPath { .. } => unimplemented!(),
+        }
+    }
```
obi1kenobi, line 70: Two suggestions:
- The underlying code uses `.name()` and `.version()` for these accessor methods, so let's match that naming convention for simplicity.
- Let's return a reference and avoid allocating, since not all uses of this function will care about owning the result, and ones that do are welcome to do `.to_string()` themselves. This preserves maximum flexibility for the caller and again matches the upstream API.
```suggestion
    fn name(&self) -> &str {
        match self {
            Self::Registry { crate_ } => crate_.name(),
            Self::ManifestPath { .. } => unimplemented!(),
        }
    }

    fn version(&self) -> &str {
        match self {
            Self::Registry { crate_ } => crate_.version(),
            Self::ManifestPath { .. } => unimplemented!(),
        }
    }
```

src/baseline.rs
```diff
@@ -7,22 +7,232 @@ use crate::dump::RustDocCommand;
...
+}
+
+fn generate_rustdoc(
+    config: &mut GlobalConfig,
+    rustdoc: &RustDocCommand,
+    target_root: PathBuf,
+    crate_source: CrateSource,
+) -> anyhow::Result<PathBuf> {
+    let name = crate_source.get_name();
+    let version = crate_source.get_version();
```
obi1kenobi, line 156:
```suggestion
    let name = crate_source.name();
    let version = crate_source.version();
```
To match my suggestion above.

src/baseline.rs
```diff
@@ -7,22 +7,232 @@ use crate::dump::RustDocCommand;
...
+
+    let (build_dir, cache_dir, cached_rustdoc) = match crate_source {
+        CrateSource::Registry { .. } => {
+            let crate_identifier = format!("registry-{}-{}", slugify(&name), slugify(&version));
```
obi1kenobi, line 160:
```suggestion
            let crate_identifier = format!("registry-{}-{}", slugify(name), slugify(version));
```
This is the place where we use those values without caring whether they are owned or not. We just saved two copies and allocations!

src/baseline.rs
tonowak: I just did it a few seconds before you reviewed it :)

src/baseline.rs
tonowak: I see. So the convention is to by default return `&str` instead of `String`?

obi1kenobi: Generally, it's better to return types that assume less and enable more for the caller. The caller can easily make `&str` into a `String` if they want to, but if we return `String` here, they can't retroactively opt out of the allocation and copy.

tonowak: Done in d284a1692ae9885ab16602bb66c8f0de14e76819 and 91114184ecadea4331ffbb920381507ec2ee0c45.

obi1kenobi (approved):

PR #341, opened 2023-02-03
Changed all baseline rustdoc generations to use placeholder manifest

tonowak:
This PR is a continuation of #340.

This PR changes the way baseline rustdoc of `PathBaseline` is being generated -- similarly to  `RegistryBaseline`, a placeholder project is created (with a dependency to the true project) on which the rustdoc is generated.

I was worried about two possible places which could have bugs, but after writing this PR, it turned out to not be a problem:
- the dependency is given an absolute path, instead of a relative path, so the target directory in which the placeholder project is placed doesn't matter,
- getting all features from a manifest (instead of the registry) turned out to be *extremely* similar to getting all features from registry. I've split the feature functions into multiple smaller functions, which shows similarities for easier comparison and lesser chance of making a bug.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -377,11 +377,11 @@ jobs:
         continue-on-error: true
         run: |
           cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-root="subject-baseline/Cargo.toml" > output
-          echo "$?" > return_code
+          touch failed_successfully
 
-      - name: Check return code
+      - name: Check whether it failed
         run: |
-          if [[ "$(cat return_code)" == "0" ]]; then exit 1; else exit 0; fi
+          if [ -f failed_successfully ]; then exit 1; else exit 0; fi
 
       - name: Check output
         run: |
@@ -391,17 +391,18 @@ jobs:
 
       - name: Cleanup
         run: |
-          rm output return_code
+          rm output
+          rm -f failed_successfully
 
       - name: Run semver-checks with --baseline-rev
         continue-on-error: true
         run: |
           cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-rev=3.1.18 > output
-          echo "$?" > return_code
+          touch failed_successfully
 
-      - name: Check return code
+      - name: Check whether it failed
         run: |
-          if [[ "$(cat return_code)" == "0" ]]; then exit 1; else exit 0; fi
+          if [ -f failed_successfully ]; then exit 1; else exit 0; fi
 
       - name: Check output
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -397,7 +397,7 @@ jobs:
       - name: Fetch v3.1.18 in subject-current
         run: |
           cd subject-current
-          git fetch origin e36cf1a67ee6a447d3615a70b065d2b4f5e60
+          git fetch origin 524e36cf1a67ee6a447d3615a70b065d2b4f5e60
           cd ..
 
       - name: Run semver-checks with --baseline-rev
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -394,10 +394,15 @@ jobs:
           rm output
           rm -f failed_successfully
 
+      - name: Fetch v3.1.18 in subject-current
+        run: |
+          cd subject-current
+          git fetch origin e36cf1a67ee6a447d3615a70b065d2b4f5e60
+          cd ..
+
       - name: Run semver-checks with --baseline-rev
         continue-on-error: true
         run: |
-          git fetch origin e36cf1a67ee6a447d3615a70b065d2b4f5e60
           cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-rev=524e36cf1a67ee6a447d3615a70b065d2b4f5e60 > output
           touch failed_successfully
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -377,11 +377,11 @@ jobs:
         continue-on-error: true
         run: |
           cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-root="subject-baseline/Cargo.toml" > output
-          touch failed_successfully
+          touch unexpectedly_did_not_fail
 
       - name: Check whether it failed
         run: |
-          if [ -f failed_successfully ]; then exit 1; else exit 0; fi
+          if [ -f unexpectedly_did_not_fail ]; then exit 1; else exit 0; fi
 
       - name: Check output
         run: |
@@ -392,7 +392,7 @@ jobs:
       - name: Cleanup
         run: |
           rm output
-          rm -f failed_successfully
+          rm -f unexpectedly_did_not_fail
 
       - name: Fetch v3.1.18 in subject-current
         run: |
@@ -404,11 +404,11 @@ jobs:
         continue-on-error: true
         run: |
           cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-rev=524e36cf1a67ee6a447d3615a70b065d2b4f5e60 > output
-          touch failed_successfully
+          touch unexpectedly_did_not_fail
 
       - name: Check whether it failed
         run: |
-          if [ -f failed_successfully ]; then exit 1; else exit 0; fi
+          if [ -f unexpectedly_did_not_fail ]; then exit 1; else exit 0; fi
 
       - name: Check output
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -397,7 +397,7 @@ jobs:
       - name: Run semver-checks with --baseline-rev
         continue-on-error: true
         run: |
-          cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-rev=v3.1.18 > output
+          cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-rev=524e36cf1a67ee6a447d3615a70b065d2b4f5e60 > output
           touch failed_successfully
 
       - name: Check whether it failed
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -397,6 +397,7 @@ jobs:
       - name: Run semver-checks with --baseline-rev
         continue-on-error: true
         run: |
+          git fetch origin e36cf1a67ee6a447d3615a70b065d2b4f5e60
           cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-rev=524e36cf1a67ee6a447d3615a70b065d2b4f5e60 > output
           touch failed_successfully
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -373,10 +373,30 @@ jobs:
         with:
           key: clap
 
-      - name: Run semver-checks
+      - name: Run semver-checks with --baseline-root
         continue-on-error: true
         run: |
           cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-root="subject-baseline/Cargo.toml" > output
+          echo "$?" > return_code
+
+      - name: Check return code
+        run: |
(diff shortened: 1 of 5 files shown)


src/baseline.rs
```diff
@@ -152,12 +180,21 @@ fn generate_rustdoc(
...
-
-    let (build_dir, cache_dir, cached_rustdoc) = match crate_source {
+    let name = crate_source.name()?;
+    let version = crate_source.version()?;
+
+    let crate_identifier = format!(
+        "{}-{}-{}",
+        match crate_source {
+            CrateSource::Registry { .. } => "registry",
+            CrateSource::ManifestPath { .. } => "local",
```
obi1kenobi, line 190:
```suggestion
            // Identifiers of manifest-based crates cannot be used for caching,
            // since they probably correspond to a specific (and unknown) gitrev and git state
            // so cached entries cannot be checked to see if they are a match or not.
            CrateSource::ManifestPath { .. } => "local",
```
Just a good reminder for our future selves, once this code is no longer "warm" in our minds.

src/baseline.rs
```diff
@@ -200,6 +234,8 @@ fn generate_rustdoc(
     match crate_source {
         CrateSource::Registry { .. } => {
             // Clean up after ourselves.
+            let cache_dir = cache_dir.unwrap();
+            let cached_rustdoc = cached_rustdoc.unwrap();
```
obi1kenobi, line 238: Please use `.expect()` with a suitable message on both. It isn't obvious why we should expect these to be `Some`, and it will only be less obvious over time as this code gets more complex, more refactored, and less fresh in our minds.

src/baseline.rs
```diff
@@ -4,69 +4,91 @@ use anyhow::Context as _;
...
+    }
+
+    fn version(&self) -> anyhow::Result<String> {
+        Ok(match self {
+            Self::Registry { crate_ } => crate_.version().to_string(),
+            Self::ManifestPath { manifest } => crate::manifest::get_package_version(manifest)?,
+        })
+    }
+
+    fn regular_features(&self) -> Vec<String> {
```
obi1kenobi, line 32: Which features are "regular"? That isn't standard terminology (not in cargo, for example) so this function could really use a doc comment.

src/baseline.rs
```diff
@@ -4,69 +4,91 @@ use anyhow::Context as _;
...
+                .into_iter()
+                .collect(),
         }
     }
 
-    fn version(&self) -> &str {
-        match self {
-            Self::Registry { crate_ } => crate_.version(),
-            Self::ManifestPath { .. } => unimplemented!(),
+    fn implicit_features(&self) -> std::collections::BTreeSet<String> {
```
obi1kenobi, line 45: Another doc comment here would be good.

src/baseline.rs
```diff
@@ -247,27 +283,29 @@ impl BaselineLoader for RustdocBaseline {
...
-    lookup: std::collections::HashMap<String, (String, PathBuf)>,
+    lookup: std::collections::HashMap<String, Manifest>,
 }
 
 impl PathBaseline {
-    pub(crate) fn new(root: &std::path::Path) -> anyhow::Result<Self> {
+    pub(crate) fn new(
+        target_root: &std::path::Path,
+        root: &std::path::Path,
+    ) -> anyhow::Result<Self> {
```
obi1kenobi, line 295: Please document these parameters. When there was one parameter, it was decently clear what it did. Now that there are two parameters of the same type, it isn't obvious which one does what.

Also, consider whether this order or the opposite order is better and less likely to lead to bugs.

obi1kenobi (commented):
I like the direction. Just needs some polishing for maintainability.

I'd also feel a lot more comfortable with these changes if we had automated tests over them that use path and gitrev arguments in the CLI. Consider adding a new job that looks to catch one of the known (but not yanked) semver violations by path and by gitrev, and ensures the output shows the semver issue was found?

src/baseline.rs
tonowak: > Also, consider whether this order or the opposite order is better and less likely to lead to bugs.

I based the order on `RegistryBaseline`. From what I understand, you suggest that the "more important" argument should be first. Then, should I also change the order in `RegistryBaseline`?

src/baseline.rs
tonowak: Done in 80e56feaec2acb6aeaabac74375dbfb8733ffe7a.

src/baseline.rs
tonowak: Done in 941dc0fb034be387c2173280fbf22eb0ec9f1220.

src/baseline.rs
tonowak: Done in 941dc0fb034be387c2173280fbf22eb0ec9f1220.

tonowak:
> Consider adding a new job that looks to catch one of the known (but not yanked) semver violations by path and by gitrev, and ensures the output shows the semver issue was found?

Do you have specific crates / versions I could use?

src/baseline.rs
tonowak: Done in 5ba44475ffc0e8a4e7003327a823acad7d1890c9. Is that what you had in mind?

src/baseline.rs
```diff
@@ -98,26 +124,32 @@ fn create_placeholder_rustdoc_manifest(
...
                     version: Some(format!("={}", crate_.version())),
-                    features: crate_features::get_all_crate_features_from_registry(crate_),
+                    features: crate_source.all_features(),
                     ..DependencyDetail::default()
                 },
-                CrateSource::ManifestPath { path, .. } => DependencyDetail {
-                    path: Some(
+                CrateSource::ManifestPath { manifest } => DependencyDetail {
+                    path: Some({
+                        assert!(manifest.path.ends_with("Cargo.toml"));
```
obi1kenobi, line 132: Two suggestions:
- `assert!()` takes additional parameters equivalent to `format!()`, let's use them to help ourselves debug this if it ever gets tripped. Otherwise we'll just get a "expected false to be true" error and we won't know anything more about the circumstances.
- Which piece of code has it as its job to check that the manifest path is valid? Based on this code, I'm guessing it's the job of the code that constructs the `CrateSource::ManifestPath` value passed here. Do we have a test to make sure it actually constructs them correctly? If not, is it feasible to add one? Because this seems like something that could sneak through to users and then subsequently both look bad on us and be annoying to debug.

src/baseline.rs
```diff
@@ -4,69 +4,95 @@ use anyhow::Context as _;
...
+            Self::ManifestPath { manifest } => crate::manifest::get_package_name(manifest)?,
+        })
+    }
+
+    fn version(&self) -> anyhow::Result<String> {
+        Ok(match self {
+            Self::Registry { crate_ } => crate_.version().to_string(),
+            Self::ManifestPath { manifest } => crate::manifest::get_package_version(manifest)?,
+        })
+    }
```
obi1kenobi, line 30: I'm tempted to say we should go in the other direction on these methods: make the `crate::manifest::get_*` functions return `Result<&'str>` instead of making this function return `Result<String>`. What do you think?

src/baseline.rs
```diff
@@ -247,29 +294,34 @@ impl BaselineLoader for RustdocBaseline {
...
 
 impl PathBaseline {
-    pub(crate) fn new(root: &std::path::Path) -> anyhow::Result<Self> {
+    pub(crate) fn new(
+        // Path to a directory with the manifest or with subdirectories with the manifests.
+        project_root: &std::path::Path,
+
+        // Path to a directory where the placeholder manifest / rustdoc can be created.
+        target_root: &std::path::Path,
+    ) -> anyhow::Result<Self> {
```
obi1kenobi, line 309: The comment content is great!

Let's use standard doc comment style for them so the text becomes visible to other tools, like rust-analyzer and editor autocomplete: https://doc.rust-lang.org/rust-by-example/meta/doc.html

obi1kenobi (commented):
Looks good so far. The tests are the biggest remaining thing, since I believe our entire test suite currently uses registry-based baselines.

I suggested a couple of tiny incremental improvements as well.

src/baseline.rs
tonowak: Assert msg done in 32a99535231d480e653fdf6568c901a67c8a0ae1.

src/baseline.rs
tonowak: Done in 641c2f4283a6af2e8cb7597cd3ec66cc65359add.

tonowak: Nope, nevermind.

```
doc comments are not allowed here
```

tonowak: I've reverted the changes. What do you propose to do now?

src/baseline.rs
tonowak: Done in efd3b5884a5e4872781379b6e392dce3d3db5dc0.

src/baseline.rs
tonowak: > Which piece of code has it as its job to check that the manifest path is valid? 

The `PathBaseline::new` searches for the manifests, parses them (thus we know for sure they are valid manifests) and saves the parsed manifests and their paths to the `lookup`. The `PathBaseline::load_rustdoc` then just uses `lookup` to pass it to `generate_rustdoc`.

I haven't fully understood your suggestion about tests (e.g. what exactly to test). Can you explain it a bit more? (and also, are they needed? The new code I've written gets the manifest paths that are for sure valid, because the parsing was successful in the steps done before my code is executed).

tonowak:
> I know the affected version for clap is still up, it's a solid candidate.

I'm not entirely sure whether I understood what you mean, but I've added a test in the CI. And also I've found a bug in the previous CI :)

.github/workflows/ci.yml
```diff
@@ -397,7 +397,7 @@ jobs:
       - name: Run semver-checks with --baseline-rev
         continue-on-error: true
         run: |
-          cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-rev=v3.1.18 > output
+          cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-rev=524e36cf1a67ee6a447d3615a70b065d2b4f5e60 > output
```
obi1kenobi, line 400: Shouldn't both of these work?

obi1kenobi: I take it back 😅

tonowak:
This PR also fixes https://github.com/obi1kenobi/cargo-semver-checks/issues/344.

src/baseline.rs
tonowak: Done in https://github.com/obi1kenobi/cargo-semver-checks/pull/341/commits/4a70618e34c59fb2880356211c1d010e86f73bcd.

src/baseline.rs
obi1kenobi: Right, so my question was around whether the "steps done before this code is executed" are actually tested. Because _right now_ this is all totally fine, but the correctness of this code is relying on that other code not changing and becoming incorrect or being removed.

The tests you're adding using `clap` cover this now, I think.

tonowak: Right, but a `Manifest` is being passed as an argument to this function, which has a `Path` and a parsed manifest that is a witness that the path is valid. One could argue that in the future we could change the code so that the path and the parsed manifest being sent to the function are from two different manifests, but that isn't reasonable.

.github/workflows/ci.yml
```diff
@@ -373,15 +373,42 @@ jobs:
         with:
           key: clap
 
-      - name: Run semver-checks
+      - name: Run semver-checks with --baseline-root
         continue-on-error: true
         run: |
           cargo run semver-checks check-release --manifest-path="subject-current/Cargo.toml" --baseline-root="subject-baseline/Cargo.toml" > output
-          echo "$?" > return_code
+          touch failed_successfully
```
obi1kenobi, line 380: Perhaps `unexpectedly_did_not_fail` might be a better name? `failed_successfully` sounds like the opposite of what this file's existence signifies, no?

tonowak: Done in ace4bbcd024fa889d8df91fbd3a7f147606e05a5.

obi1kenobi (approved):
Awesome job!

src/baseline.rs
obi1kenobi: > that isn't reasonable

I agree in this instance, it's unlikely. But there are cases where such a thing is not only reasonable but actively exploitable for malicious purposes in ways that are really annoying to deal with: https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use

PR #347, opened 2023-02-04
Changed some load_rustdoc arguments to a struct

tonowak:
This PR is a continuation of #341 and #340.

It doesn't yet introduce functional changes -- the only change is that I've defined some new enums/structs and I'm passing them to `load_rustdoc` functions, instead of several arguments.

The reasoning is that some arguments were not obvious / should have been documented, but it's hard to do it near each `load_rustdoc` function. It's also convenient to group those struct fields together -- they are strictly related and independent of other `load_rustdoc` arguments. The new enum allows to execute the new rustdoc generation through the placeholder for the current crate (which will be done in a next PR). The struct also allows for an easier additive change in some next PR, which will add the possibility of choosing which features to enable.

### DIFF
--- a/src/baseline.rs
+++ b/src/baseline.rs
@@ -177,6 +177,7 @@ fn save_placeholder_rustdoc_manifest(
 }
 
 #[allow(dead_code)]
+#[derive(Debug)]
 pub(crate) enum CrateType<'a> {
     Current,
     Baseline {
@@ -187,6 +188,7 @@ pub(crate) enum CrateType<'a> {
     },
 }
 
+#[derive(Debug)]
 pub(crate) struct CrateDataForRustdoc<'a> {
     pub(crate) crate_type: CrateType<'a>,
     pub(crate) name: &'a str,
--- a/src/baseline.rs
+++ b/src/baseline.rs
@@ -241,8 +241,6 @@ fn generate_rustdoc(
                         crate_data.crate_type.to_string()
                     ),
                 )?;
-                // TODO: replace "baseline" with a string passed as a function argument
-                // (the plan is to make this function work for both baseline and current).
                 return Ok(cached_rustdoc);
             }
 
@@ -261,8 +259,6 @@ fn generate_rustdoc(
         "Parsing",
         format_args!("{name} v{version} ({})", crate_data.crate_type.to_string()),
     )?;
-    // TODO: replace "baseline" with a string passed as a function argument
-    // (the plan is to make this function work for both baseline and current).
 
     let rustdoc_path = rustdoc.dump(
         placeholder_manifest_path.as_path(),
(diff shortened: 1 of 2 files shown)


src/baseline.rs
```diff
@@ -33,14 +33,8 @@ impl<'a> CrateSource<'a> {
...
-            Self::Registry { crate_ } => crate_.features().keys().cloned().into_iter().collect(),
-            Self::ManifestPath { manifest } => manifest
-                .parsed
-                .features
-                .keys()
-                .cloned()
-                .into_iter()
-                .collect(),
+            Self::Registry { crate_ } => crate_.features().keys().cloned().collect(),
+            Self::ManifestPath { manifest } => manifest.parsed.features.keys().cloned().collect(),
```
tonowak, line 37: This change was made, because the new clippy complains.

src/baseline.rs
```diff
@@ -182,11 +176,39 @@ fn save_placeholder_rustdoc_manifest(
...
+
+impl<'a> ToString for CrateType<'a> {
+    fn to_string(&self) -> String {
+        match self {
+            CrateType::Current => "current",
+            CrateType::Baseline { .. } => "baseline",
+        }
+        .to_string()
+    }
+}
```
tonowak, line 204: It feels wasteful to create a `String` when it could be `&'static str` -- is there a way to do it with this trait, or should I for example remove this trait?

src/baseline.rs
```diff
@@ -182,11 +176,39 @@ fn save_placeholder_rustdoc_manifest(
...
+        /// that is the same or older than the version of the current crate.
+        highest_allowed_version: Option<&'a semver::Version>,
+    },
+}
+
+pub(crate) struct CrateDataForRustdoc<'a> {
+    pub(crate) crate_type: CrateType<'a>,
+    pub(crate) name: &'a str,
+    // TODO: pass an enum describing which features to enable
+}
```
tonowak, line 194: I'm not sure whether I like all the names here -- do you have some better ideas @obi1kenobi ?

src/baseline.rs
obi1kenobi: Ah sorry, I merged out of order. In #348 I made this a regular method returning `&'static str` since we don't actually care about `ToString` as a trait by itself (we'll never write `impl ToString` and mean this type).

obi1kenobi:
Accidentally merged out of order, so these changes got merged in #348.

obi1kenobi (approved):
I managed to stop #348 from merging out of order, so let's do this properly 😅

src/baseline.rs
obi1kenobi: They're ... okay. I don't love them either but none of my ideas are substantially better, and we need this feature much more than we need fancy names here. These will do for now.

Issue #167, opened 2022-11-01
Semver errors when exposing types from 3rd party crates that are not at the latest version


jrmuizel:
Scenario:
- Crate `A` has a dependency on crate `B = "1.0"` and exposes crate `B`'s types in its own public API.
- The lockfile for crate `A` currently using `B` version `1.0`. Version `1.1` of `B` exists and contains additional pub methods on its pub types.
- The current version of `A` is semver-checked using the default (registry-based) baseline:
  - The current rustdoc will use `B v1.0` and will show the v1.0 methods from `B` as part of `A`'s public API.
  - The baseline rustdoc for `A` will be generated, which will cause a new lockfile to be created with the latest allowed version of `B`: v1.1.
  - The baseline's public API for `A` will show additional methods  that don't exist in the current version: those added in `B v1.1` but not present in `B v1.0.
  - cargo-semver-checks will report those as major breaking changes.

### Steps to reproduce the bug with the above code

Checkout https://github.com/servo/core-foundation-rs/tree/cocoa-v0.24.1/cocoa
Run `rustup run nightly cargo semver-checks check-release`

### Actual Behaviour

```
--- failure inherent_method_missing: pub method removed or renamed ---

Description:
A publicly-visible method or associated fn is no longer available under its prior name. It may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-check/tree/v0.12.0/src/queries/inherent_method_missing.ron

Failed in:
  NSEventMask::from_bits_unchecked, previously in file /Users/jrmuizel/.cargo/registry/src/github.com-1ecc6299db9ec823/cocoa-0.24.1/src/appkit.rs:2334
  NSEventMask::intersection, previously in file /Users/jrmuizel/.cargo/registry/src/github.com-1ecc6299db9ec823/cocoa-0.24.1/src/appkit.rs:2334
  NSEventMask::union, previously in file /Users/jrmuizel/.cargo/registry/src/github.com-1ecc6299db9ec823/cocoa-0.24.1/src/appkit.rs:2334
  NSEventMask::difference, previously in file /Users/jrmuizel/.cargo/registry/src/github.com-1ecc6299db9ec823/cocoa-0.24.1/src/appkit.rs:2334
[...]

PR #350, opened 2023-02-04
Baseline variable renames

tonowak:
This PR is stacked on https://github.com/obi1kenobi/cargo-semver-checks/pull/348.

In the next PR, the trait `BaselineLoader` and its structs will be also user for the `current` crate, the names won't be relevant anymore. This PR renames the trait and structs.

PR #351, opened 2023-02-04
Using placeholder rustdoc generation for current crate

tonowak:
This PR is stacked on https://github.com/obi1kenobi/cargo-semver-checks/pull/350.

It changes the logic of the tool. Instead of generating the rustdoc of the current crate directly, it uses the baseline's way of creating a placeholder project and running rustdoc on it.

### DIFF
--- a/src/manifest.rs
+++ b/src/manifest.rs
@@ -1,3 +1,5 @@
+use anyhow::Context as _;
+
 #[derive(Debug, Clone)]
 pub(crate) struct Manifest {
     pub(crate) path: std::path::PathBuf,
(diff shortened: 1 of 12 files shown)


tonowak:
This PR should fix https://github.com/obi1kenobi/cargo-semver-checks/issues/167, but I don't have an Apple I could use to check it. @obi1kenobi can you test whether this PR fixes the issue?

src/manifest.rs
```diff
@@ -1,3 +1,5 @@
+use anyhow::Context as _;
```
obi1kenobi, line 1: Do we need the `as _` bit here?

obi1kenobi (commented):
This looks good too, just the one nitpick.

obi1kenobi:
> This PR should fix #167, but I don't have an Apple I could use to check it. @obi1kenobi can you test whether this PR fixes the issue?

I don't have an Apple device either. Maybe we can ask the original reporter of #167 to try it once we release this.

src/manifest.rs
tonowak: Done in 3b18ce47cc9e14eaa9e63281968e8dc9eaf3420b. I was basing this code on `manifest.rs` which also had `as _`, so I removed it there also.

obi1kenobi (approved):

src/manifest.rs
obi1kenobi: Ah I think I just realized what was going on here. `Context` is a trait, probably? And the `as _` means that the trait's methods are brought into scope without bringing the trait itself into scope.

Issue #346, opened 2023-02-04
The lint job in the CI is missing a `cargo rustdoc` step 


tonowak:
There might be problems in the code with the doc comments, but they are not checked through clippy:

```
➜  cargo-semver-checks git:(baseline-all-rustdocs-from-placeholder) ✗ cargo clippy --workspace --all-features --all-targets -- -D warnings --allow deprecated
    Checking cargo-semver-checks v0.17.1 (/home/tonowak/zpp/cargo-semver-checks)
    Finished dev [unoptimized + debuginfo] target(s) in 0.86s
➜  cargo-semver-checks git:(baseline-all-rustdocs-from-placeholder) ✗ cargo rustdoc
    Checking bstr v1.1.0
    Checking globset v0.4.10
    Checking ignore v0.4.20
 Documenting cargo-semver-checks v0.17.1 (/home/tonowak/zpp/cargo-semver-checks)
warning: this URL is not a hyperlink
  --> src/baseline.rs:33:9
   |
33 |     /// https://doc.rust-lang.org/cargo/reference/features.html#the-features-section
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use an automatic link instead: `<https://doc.rust-lang.org/cargo/reference/features.html#the-features-section>`
   |
   = note: bare URLs are not automatically turned into clickable links
   = note: `#[warn(rustdoc::bare_urls)]` on by default

warning: this URL is not a hyperlink
  --> src/baseline.rs:48:9
   |
48 |     /// https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use an automatic link instead: `<https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies>`
   |
   = note: bare URLs are not automatically turned into clickable links

warning: `cargo-semver-checks` (bin "cargo-semver-checks" doc) generated 2 warnings
    Finished dev [unoptimized + debuginfo] target(s) in 2.18s
```

I suggest adding a `cargo rustdoc` step in the CI.

PR #352, opened 2023-02-04
Added `cargo doc` in lint job

tonowak:
Fixes https://github.com/obi1kenobi/cargo-semver-checks/issues/346

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,9 +61,9 @@ jobs:
           args: -- --check
 
       - name: cargo doc
-        uses: actions-rs/cargo@v1
-        with:
-          command: doc
+        env:
+          RUSTDOCFLAGS: -D warnings
+        run: cargo doc --workspace --all-features --no-deps --document-private-items
 
   rust-tests:
     name: Run tests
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -60,6 +60,11 @@ jobs:
           command: fmt
           args: -- --check
 
+      - name: cargo doc
+        uses: actions-rs/cargo@v1
+        with:
+          command: doc
```
epage, line 66: Will this error on warnings?

I suspect not and this is one of the several reasons why I dislike the `cargo` action (its also a dependency that does not pull its weight).

I'd recommend something more like what I run in my other projects: https://github.com/rust-cli/argfile/blob/main/.github/workflows/ci.yml#L68

tonowak: Thanks for the review! Is this what you had in mind? 5a5a130ce5ba83c40453a866d410aa8630de0a85

obi1kenobi (approved):

.github/workflows/ci.yml
epage: That works!

PR #353, opened 2023-02-04
Allow specifying feature set

tonowak:
This PR is stacked on https://github.com/obi1kenobi/cargo-semver-checks/pull/351.

It allows to generate the rustdocs with some user-specified feature set (with the ability to disable default features).

This code is untested and unused -- it is a prototype. Based on this code, the CLI should be extended with new flags (and I prefer to leave it completely to @obi1kenobi) and then extensive integration tests should be written with some new test crates.


obi1kenobi:
Is it ever possible that we might want different features enabled on baseline vs current? 🤔

tonowak:
> Is it ever possible that we might want different features enabled on baseline vs current? thinking

Yes. Someone could want to, for example, check whether a feature only adds additive changes. Then one would want to compare the same code with a different set of features.

tonowak:
I certainly prefer to leave the CLI design to you -- could you write the flags etc, and then I would implement the logic behind the flags and the integration tests?

tonowak:
I've checked the existing github issues with features and there were two ideas there that I haven't yet implemented:
- ignoring features prefixed with `_` when executing with all features,
- adding a list of features to exclude from all features.

Should I also implement those changes?

src/rustdoc_gen.rs
```diff
@@ -90,10 +90,40 @@ impl<'a> CrateSource<'a> {
...
+pub(crate) struct FeatureSet {
+    pub(crate) default_features: bool,
+    pub(crate) features: Vec<String>,
+}
+
+#[allow(dead_code)]
+#[derive(Debug, Clone)]
+pub(crate) enum FeatureBehaviour {
+    AllFeatures,
+    SpecifiedFeatues(FeatureSet),
```
tonowak, line 103: Now that I think about it, making it an enum with named fields would be shorter and possibly more readable.

tonowak: Done in 44bde8abb45596dc39126788812e1bbd13727f07.

PR #357, opened 2023-02-09
cap-lints when building rustdoc JSON

Emilgardis:
Due to using `--document-private-items --document-hidden-items`, some new warnings not previously found by the creator(s) of a crate can be triggered as most do not check that rustdoc builds when documenting private/hidden items.

This also happens when rustdoc behaviour changes between versions, see https://github.com/obi1kenobi/cargo-semver-checks-action/issues/17

This fixes that

### DIFF
--- a/scripts/regenerate_test_rustdocs.sh
+++ b/scripts/regenerate_test_rustdocs.sh
@@ -27,7 +27,7 @@ for crate_pair in $(find "$TOPLEVEL/test_crates/" -maxdepth 1 -mindepth 1 -type 
                 echo "Generating: $crate"
 
                 pushd "$TOPLEVEL/test_crates/$crate"
-                RUSTC_BOOTSTRAP=1 $RUSTDOC_CMD -- -Zunstable-options --document-private-items --document-hidden-items --output-format=json
+                RUSTC_BOOTSTRAP=1 $RUSTDOC_CMD -- -Zunstable-options --document-private-items --document-hidden-items --cap-lints allow --output-format=json
                 mkdir -p "$TARGET_DIR/$crate"
                 mv "$RUSTDOC_OUTPUT_DIR/$crate_pair.json" "$TARGET_DIR/$crate/rustdoc.json"
                 popd
--- a/scripts/regenerate_test_rustdocs.sh
+++ b/scripts/regenerate_test_rustdocs.sh
@@ -27,7 +27,16 @@ for crate_pair in $(find "$TOPLEVEL/test_crates/" -maxdepth 1 -mindepth 1 -type 
                 echo "Generating: $crate"
 
                 pushd "$TOPLEVEL/test_crates/$crate"
-                RUSTC_BOOTSTRAP=1 $RUSTDOC_CMD -- -Zunstable-options --document-private-items --document-hidden-items --cap-lints allow --output-format=json
+
+                # Determine whether to warn on lints or allow them.
+                CAP_LINTS="warn"
+                if [[ "$crate" == "broken_rustdoc" ]]; then
+                    # This crate *intentionally* has broken rustdoc.
+                    # Don't warn on it. The warnings and errors are thing being tested.
+                    CAP_LINTS="allow"
+                fi
+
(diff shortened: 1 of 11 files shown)


Emilgardis:
Was bit by this when using 

`#![deny(rustdoc::broken_intra_doc_links)]`

obi1kenobi:
Wow, thanks -- this is awesome, I didn't know it existed and we've needed something like it before. A couple of questions:
- Could you share the crate and git hash/version where this came up? I'd love it if we could add it to our CI test suite as a regression test.
- Do you know if using `--cap-lints` forces a full rebuild or not? We were [considering using the `RUSTFLAGS` env var to allow all lints](https://github.com/obi1kenobi/cargo-semver-checks-action/issues/17), but decided against using that always & by default because it would always cause a full rebuild which would be unnecessarily expensive.

Emilgardis:
It doesn't trigger a rebuild for me atleast

got this specifically on `twitch_api v0.7.0-rc.4` as baseline, however I wouldn't use that as a test case as I'm most likely going to be janking all versions (including release-candidates) of that crate before 0.7.0 (and all versions <0.7.0 have already been janked)

obi1kenobi:
> It doesn't trigger a rebuild for me atleast

Awesome!

> got this specifically on `twitch_api v0.7.0-rc.4` as baseline

If you don't mind specifying a repo and gitsha for that, we can add it as a test like that too. It doesn't have to be published, we'll just check out that commit specifically and use the `--baseline-rev` flag to set a baseline.

Emilgardis:
[here's the commit](https://github.com/twitch-rs/twitch_api/commit/da7c6cc60963cf47a9f8e9394744eaf14643b488) da7c6cc60963cf47a9f8e9394744eaf14643b488 on https://github.com/twitch-rs/twitch_api

jyn514:
> Do you know if using `--cap-lints` forces a full rebuild or not? We were [considering using the `RUSTFLAGS` env var to allow all lints](https://github.com/obi1kenobi/cargo-semver-checks-action/issues/17), but decided against using that always & by default because it would always cause a full rebuild which would be unnecessarily expensive.

RUSTFLAGS affects all dependencies, including proc macros and dependencies not in the workspace. RUSTDOCFLAGS is only passed to rustdoc, so it only affects crates in the workspace, since you're passing `--no-deps` to rustdoc.

obi1kenobi:
Thanks @jyn514, that makes sense!

I'll add the test case to this PR shortly and then merge.

obi1kenobi:
Thanks for the heads-up! I'll take a look at why it's taking so long, hopefully something I might be able to fix while I'm at it.

scripts/regenerate_test_rustdocs.sh
```diff
@@ -27,7 +27,7 @@ for crate_pair in $(find "$TOPLEVEL/test_crates/" -maxdepth 1 -mindepth 1 -type
                 echo "Generating: $crate"
 
                 pushd "$TOPLEVEL/test_crates/$crate"
-                RUSTC_BOOTSTRAP=1 $RUSTDOC_CMD -- -Zunstable-options --document-private-items --document-hidden-items --output-format=json
+                RUSTC_BOOTSTRAP=1 $RUSTDOC_CMD -- -Zunstable-options --document-private-items --document-hidden-items --cap-lints allow --output-format=json
```
jyn514, line 30: Maybe use --cap-lints warn instead so the author still sees the warning?

I guess might not make sense for semver-checks since the fact it uses rustdoc is an implementation detail, though.

obi1kenobi: Thanks, good idea.

By itself, that would get a bit annoying since I just pushed a crate that is going to emit a bunch of warnings every time this is executed.

So I'll add a bit more logic there and `allow` for that crate while using `warn` for everything else.

obi1kenobi:
Good news and bad news on the slow check time.

The good news is that I know what the problem is and I have a solution in mind that's guaranteed to fix it: https://predr.ag/blog/speeding-up-rust-semver-checking-by-over-2000x/

It's because `twitch_api` is a large crate with ~69k items, around 30% of the size of the crate I used in the example in that blog post.

Bad news: it's not something I'll be able to fix today or this week. It needs a bit more polishing and, while I'm doing the best I can, I'm stretched quite thin.

I'll try to tweak some of the queries to see if I can get some quicker wins in the meantime.

Issue #326, opened 2023-01-25
Error output from rustdoc is stripped of color


jyn514:
### Steps to reproduce the bug with the above code

1. Create any cargo project (e.g. `cargo new --lib example`)
2. Modify it so it doesn't compile (e.g. `echo x > src/lib.rs`)
3. Run `cargo semver-checks check-release`

### Actual Behaviour

The errors are printed without color:

<img width="1458" alt="Screenshot 2023-01-25 at 11 15 04 AM" src="https://user-images.githubusercontent.com/23638587/214633761-f32d336f-eedc-45df-a650-09bbb847993f.png">

### Expected Behaviour

The errors are printed with color, just like `cargo doc`:

<img width="1449" alt="Screenshot 2023-01-25 at 11 15 47 AM" src="https://user-images.githubusercontent.com/23638587/214634072-de81e6ee-cb12-4d90-a39d-1faf4afb7f4d.png">

#### Software version

cargo-semver-checks 0.16.2

#### Operating system

macOS 13.1 (Darwin 22.2.0)

#### Command-line

```bash
/Users/jyn/.local/lib/cargo/bin/cargo-semver-checks semver-checks --bugreport 
```

#### cargo version

```
> cargo -V 
cargo 1.68.0-nightly (a5d47a725 2023-01-16)
```

#### Compile time information

- Profile: release
- Target triple: aarch64-apple-darwin
- Family: unix
- OS: macos
- Architecture: aarch64
- Pointer width: 64
- Endian: little
- CPU features: aes,crc,dit,dotprod,dpb,dpb2,fcma,fhm,flagm,fp16,frintts,jsconv,lor,lse,neon,paca,pacg,pan,pmuv3,ras,rcpc,rcpc2,rdm,sb,sha2,sha3,ssbs,vh
- Host: aarch64-apple-darwin

### Build Configuration

```
; env | grep CARGO_
CARGO_HOME=/Users/jyn/.local/lib/cargo
CARGO_TARGET_DIR=/Users/jyn/.local/lib/cargo/target
```

### Additional Context

I see that you're passing `--json=diagnostic-rendered-ansi`, so I'm not sure where the color stripping happens. Maybe cargo doesn't think it's being output to a terminal and you need to pass `--color` explicitly?

jyn514:
Interestingly, this works fine when I pass `-v`, it's only the default output that has color stripped.

obi1kenobi:
[...]

PR #360, opened 2023-02-09
Use `--color=always` in rustdoc command when stderr is tty

tonowak:
Fixes https://github.com/obi1kenobi/cargo-semver-checks/issues/326


src/rustdoc_cmd.rs
```diff
@@ -83,6 +83,9 @@ impl RustdocCommand {
         if all_features {
             cmd.arg("--all-features");
         }
+        if atty::is(atty::Stream::Stderr) {
+            cmd.arg("--color=always");
+        }
```
obi1kenobi, line 88: Should this reference the `GlobalConfig` struct's `stderr` value instead of the global stderr, just in case they are different?

This might need a bit more plumbing.

tonowak: But isn't `anyhow::bail!` using the normal stderr, and that's where this command's output will go?

tonowak: https://github.com/obi1kenobi/cargo-semver-checks/blob/4de32032f5eeb73e9c7bc95dcc8b4dedc207e43a/src/rustdoc_cmd.rs#L92

obi1kenobi: That is an excellent point!

We should change that too, before a "cargo-semver-checks-as-a-library" user opens an issue about their library polluting their stderr even though they configured its stderr behavior to be something else.

tonowak: I see. But nevertheless it should use the `GlobalConfig`, right?

obi1kenobi:
Yes, I think so.

src/rustdoc_cmd.rs
tonowak: Done in 8f7d3870d6212ee2181dbb8c71f0bce004fe304f.

tonowak:
Why is the CI failing? I don't understand it. The compilation error references line numbers in `rustdoc_cmd.rs` which don't exist (the file is shorter than the line numbers).

obi1kenobi:
> Why is the CI failing? I don't understand it. The compilation error references line numbers in `rustdoc_cmd.rs` which don't exist (the file is shorter than the line numbers).

Possibly due to merging `main` into this branch first. There's some poorly-documented default behavior of GitHub Actions where PR tests implicitly run on the current branch with `main` merged into it, IIRC.

I just merged `main` and if that causes the line numbers to more-or-less match up (because I merged #375 into `main` first, after this CI run), that might be the cause.

obi1kenobi (approved):
Nice!

PR #363, opened 2023-02-10
Fix syntax highlighting in README.md

Finomnis:
`console` is not a recognized syntax highlighting tag by most markdown viewers

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Lint your crate API changes for semver violations.
 
 ## Quick Start
 
-```console
+```sh
 $ cargo install cargo-semver-checks --locked
 
 # Check whether it's safe to release the new version:


obi1kenobi:
Is `sh` the right thing for a shell _session_ block? I thought it was just for shell files, which this isn't?

The GitHub Linguist project lists `console` as an alias, which is why it's used in the README at the moment.
https://github.com/github/linguist/blob/3c3b037910006fc2f1a9bb34b2c4e9cde062206c/lib/linguist/languages.yml#L5842-L5843

Finomnis:
> Is `sh` the right thing for a shell _session_ block? I thought it was just for shell files, which this isn't?
> 
> The GitHub Linguist project lists `console` as an alias, which is why it's used in the README at the moment. https://github.com/github/linguist/blob/3c3b037910006fc2f1a9bb34b2c4e9cde062206c/lib/linguist/languages.yml#L5842-L5843

There are no such things as `#` comments in a shell session. It works with the `sh`, but doesn't with `console` ;)

I tried here in github and on VSCode, and both are fine with the `sh` tag.

See for yourself:
- before:
  https://github.com/obi1kenobi/cargo-semver-checks/tree/14e77ee491733423d731b0e7ae62b08948d28514#quick-start
- after:
  https://github.com/obi1kenobi/cargo-semver-checks/blob/03e0f86ceb62cdf801b218e035540b2f0d3f97a5/README.md#quick-start

obi1kenobi:
I'm convinced! Thanks for spotting the problem and fixing it, much appreciated!

PR #364, opened 2023-02-10
Add "skip minor lints" flag

Finomnis:
Related to #359, but doesn't directly close it. But might make it obsolete.


Finomnis:
Rejected in favor of #361

Issue #370, opened 2023-02-16
workspace version is not supported (regression between 0.17 and 0.18)


skyzh:
### Steps to reproduce the bug with the above code

It seems that there's a regression between 0.17.1 and 0.18.0 that version in workspace is not supported any more?

Thanks for investigating into this issue!

In our root workspace config,

```
[workspace.package]
version = "0.13.0"
```

In the workspace member, we have:

```
[package]
name = "sqllogictest"
version = { workspace = true }
```

### Actual Behaviour

```
cargo install cargo-semver-checks --version ^0.18 --locked
git clone https://github.com/risinglightdb/sqllogictest-rs && cd sqllogictest-rs
cargo semver-checks check-release -p sqllogictest
    Updating index
Error: Failed to retrieve package version from /Users/skyzh/Work/sqllogictest-rs/sqllogictest/Cargo.toml: value from workspace hasn't been set
```

### Expected Behaviour

```
cargo install cargo-semver-checks --version ^0.17 --locked
git clone https://github.com/risinglightdb/sqllogictest-rs && cd sqllogictest-rs
cargo semver-checks check-release -p sqllogictest
    Updating index
     Parsing sqllogictest v0.13.0 (current)
     Parsing sqllogictest v0.13.0 (baseline)
    Checking sqllogictest v0.13.0 -> v0.13.0 (no change)
   Completed [   2.507s] 40 checks; 40 passed, 0 unnecessary
```

#### Software version

cargo-semver-checks 0.18.0

#### Operating system

macOS 13.1 (Darwin 22.2.0)

#### Command-line

```bash
/Users/skyzh/.cargo/bin/cargo-semver-checks semver-checks --bugreport
```

#### cargo version

```
> cargo -V
cargo 1.67.1 (8ecd4f20a 2023-01-10)
```

#### Compile time information

- Profile: release
- Target triple: aarch64-apple-darwin
- Family: unix
- OS: macos
- Architecture: aarch64
- Pointer width: 64
- Endian: little
- CPU features: aes,crc,dit,dotprod,dpb,dpb2,fcma,fhm,flagm,fp16,frintts,jsconv,lor,lse,neon,paca,pacg,pan,pmuv3,ras,rcpc,rcpc2,rdm,sb,sha2,sha3,ssbs,vh
- Host: aarch64-apple-darwin

skyzh:
fix on the way...

skyzh:
[...]

PR #371, opened 2023-02-16
support workspace version

skyzh:
close https://github.com/obi1kenobi/cargo-semver-checks/issues/370


obi1kenobi:
Sorry for the bug and thanks for the fix! I'd be happy to merge and release this ASAP.

Any idea why it's failing in CI? It seems to be unable to find the generated rustdoc file for one of our tests. I'm not familiar with workspace versions so I figured I'd ask in case something stood out to you.

I'd also love to have the crate and version where this broke to our [integration test suite](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/.github/workflows/ci.yml#L22-L28), so this doesn't ever happen again. You're welcome to add it or I can do it right before merging.

skyzh:
Sure! I'll investigate the CI failure and add our test cases tonight...

skyzh:
This looks like a parsing bug of upstream library `toml_cargo`. The `Cargo.toml`:

```
[package]
name = "renamed_lib"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[lib]
name = "librenamed"

[dependencies]
```

is parsed as:

```
    lib: Some(
        Product {
            path: Some(
                "src/lib.rs",
            ),
            name: Some(
                "renamed_lib",
            ),
```

where the lib name is wrong. I'll see how to workaround this...

obi1kenobi:
That's a bummer :( It might be a good idea to open an issue on their issue tracker if you have a chance, since the repo seems decently active: https://gitlab.com/crates.rs/cargo_toml/-/issues

skyzh:
I've workaround this issue by first parsing directly with toml, and then cargo_toml. I'm going to submit a fix to upstream and meanwhile I think this workaround will work for now, unless someone uses both renamed lib and workspace inheritance 🤣

skyzh:
upstream PR: https://gitlab.com/crates.rs/cargo_toml/-/merge_requests/16/

obi1kenobi (commented):
The code looks good so far. Unfortunately, it doesn't seem to be a complete solution to the problem.

I added an integration test (with two checks) against the specific library and version that encountered the issue, and as far as I can tell both checks are failing at the moment.

test_rustdoc/crate_in_workspace/crate1/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+publish = false
+name = "crate_in_workspace_crate1"
+version = "0.1.0"
```
epage, line 4: Isnt' this supposed to be `version.workspace = true`?

src/manifest.rs
```diff
@@ -8,11 +8,25 @@ pub(crate) struct Manifest {
...
-        let parsed = toml::from_str(manifest_text.as_str())
-            .map_err(|e| anyhow::format_err!("Failed to parse {}: {}", path.display(), e))?;
-
+        let parsed = match toml::from_str(&manifest_text) {
+            Ok(parsed) => parsed,
+            Err(_) => {
+                // Parsing failed, possibly because of workspace inheritance. Retry with cargo_toml.
+                cargo_toml::Manifest::from_path(&path)?
+            }
+        };
```
epage, line 29: Why do two phase loading instead of directly loading what is needed?

test_rustdoc/crate_in_workspace/crate1/Cargo.toml
skyzh: Oh yes sorry for that. I don't know what I'm thinking about yesterday night 🤣

src/manifest.rs
skyzh: cargo_toml failed to get the correct library name, and therefore this is a workaround to forcefully overwrite the library name parsed by cargo_toml. I have an upstream PR. Once it's fixed we can remove this workaround.

obi1kenobi: @epage is there something I could add or change about the comment above this block to make this clearer? I updated it [here](https://github.com/obi1kenobi/cargo-semver-checks/pull/371/commits/19e66b291a12c8c3a61f5c1561add0578393c31f) but it sounds like it was still confusing.

src/rustdoc_cmd.rs
```diff
@@ -207,4 +207,15 @@ mod tests {
...
+    #[test]
+    fn rustdoc_for_crate_in_workspace() {
+        RustdocCommand::default()
+            .dump(
+                Path::new("./test_rustdoc/crate_in_workspace/crate1/Cargo.toml"),
+                None,
+                true,
+            )
+            .expect("no errors");
+    }
```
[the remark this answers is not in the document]
obi1kenobi: @skyzh might it make sense to add a second test case, similar to this one but using the workspace's `Cargo.toml` and passing a package spec instead of that `None`? I believe that should work too, if all goes well.

skyzh:
Just pushed a commit and I believe this will be the real "dirty" fix 🤣 passing integration test for sqllogictest 0.13 locally and let's see if it works on CI. Thanks for your reviews!

src/rustdoc_cmd.rs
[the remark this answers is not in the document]
skyzh: sure, I'll update the test case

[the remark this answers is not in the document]
skyzh: done

src/manifest.rs
epage: I had missed it because the comment is not on the weird case but on the expected case.

skyzh:
I'll submit another PR once the upstream fix has been merged and released to crates.io.

src/manifest.rs
obi1kenobi: I'll try to move it closer 👍

src/manifest.rs
```diff
@@ -8,10 +8,33 @@ pub(crate) struct Manifest {
 
 impl Manifest {
     pub(crate) fn parse(path: std::path::PathBuf) -> anyhow::Result<Self> {
+        // Parsing via `cargo_toml::Manifest::from_path()` is preferable to parsing from a string,.
```
obi1kenobi, line 11:
```suggestion
        // Parsing via `cargo_toml::Manifest::from_path()` is preferable to parsing from a string,
```

obi1kenobi (approved):
Thanks for the help! I'm going to merge this and queue up a patch release.

PR #374, opened 2023-02-18
bump cargo-toml to 0.15.2 and remove workarounds

skyzh:
Thanks to @kornelski for a quick review for my PR to cargo_toml and we can remove the workaround now!

### DIFF
--- a/src/manifest.rs
+++ b/src/manifest.rs
@@ -11,30 +11,9 @@ impl Manifest {
         // Parsing via `cargo_toml::Manifest::from_path()` is preferable to parsing from a string,
         // because inspection of surrounding files is sometimes necessary to determine
         // the existence of lib targets and ensure proper handling of workspace inheritance.
-        let mut parsed = cargo_toml::Manifest::from_path(&path).unwrap();
 
-        // `cargo_toml` currently has buggy handling of renamed library targets:
-        // https://gitlab.com/crates.rs/cargo_toml/-/merge_requests/16/
-        //
-        // This is a workaround for that bug, added in:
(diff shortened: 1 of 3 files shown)


src/manifest.rs
```diff
@@ -8,33 +8,7 @@ pub(crate) struct Manifest {
...
-                *parsed
-                    .lib
-                    .as_mut()
-                    .expect("no lib element")
-                    .name
-                    .as_mut()
-                    .expect("no name in the lib") = name.clone();
-            }
-        }
+        let parsed = cargo_toml::Manifest::from_path(&path)?;
```
obi1kenobi, line 11: Can you keep the original `.map_err()` call that adds a bit of context on the failing operation? Just pull the same error message that was here before on the parsing call before the workaround PR.

skyzh:
Comment resolved! Actually we can use anyhow's `.context` here, which produces something like (as in [anyhow::Context](https://docs.rs/anyhow/latest/anyhow/trait.Context.html) example):

```
Error: Failed to read instrs from ./path/to/instrs.json

Caused by:
    No such file or directory (os error 2)
```

```rust
        let parsed = cargo_toml::Manifest::from_path(&path)
            .with_context(|e| format!("Failed when reading {}", path.display())?;
```

Do you feel this is a better way to throw an error? If so I can refactor map_err to context for manifest errors in this PR.

obi1kenobi:
Yes, definitely! Thanks!

skyzh:
done :)

obi1kenobi (approved):

obi1kenobi:
Awesome! Thank you so much!

PR #375, opened 2023-02-20
Fix false-positives from target-specific optional features

### LINKED ISSUE #369: Implicit features due to target-specific optional dependencies are not added in current version
> @obi1kenobi Upgrading to `v0.18` results in a failing semver check despite this crate not having received any significant changes. Could you look into this?

_Originally posted by @thomaseizinger in https://github.com/libp2p/rust-libp2p/issues/3469#issuecomment-1432218666_

Affected crate: `libp2p_uds` version `0.17`, repro commit `5b4eab7bafe6cba8fea6c806027ea680662b7cc6`

Its `Cargo.toml` contains the following block:
```toml
[target.'cfg(all(unix, not(target_os = "emscripten")))'.dependencies]
async-std = { version = "1.6.2", optional = true }
libp2p-core = { version = "0.38.0", path = "../../core" }
log = "0.4.1"
futures = "0.3.1"
tokio = { version = "1.15", default-features = false, features = ["net"], optional = true }
```

This implicitly defines features named `async-std` and `tokio`. These features are enabled in the baseline, but for some reason are not enabled in the "current":
```toml
[dependencies.libp2p-uds]
path = "/.../rust-libp2p/transports/uds"
features = []
default-features = false
```

tonowak:
Fixes #369

### DIFF
--- a/src/rustdoc_gen.rs
+++ b/src/rustdoc_gen.rs
@@ -48,13 +48,17 @@ impl<'a> CrateSource<'a> {
                 .filter_map(|dep| dep.is_optional().then_some(dep.name()))
                 .map(|x| x.to_string())
                 .collect(),
-            Self::ManifestPath { manifest } => manifest
-                .parsed
-                .dependencies
-                .iter()
-                .filter_map(|(name, dep)| dep.optional().then_some(name))
-                .map(|x| x.to_string())
-                .collect(),
+            Self::ManifestPath { manifest } => {
+                let mut dependencies = manifest.parsed.dependencies.clone();
+                for target in manifest.parsed.target.values() {
+                    dependencies.extend(target.dependencies.clone().into_iter());
+                }
+                dependencies
+                    .iter()
+                    .filter_map(|(name, dep)| dep.optional().then_some(name))
+                    .map(|x| x.to_string())
+                    .collect()
+            }
         };
 
         let feature_defns: Vec<&String> = match self {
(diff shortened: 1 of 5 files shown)


tonowak:
Is there a situation where we *don't* want to look into target-specific dependencies?

tonowak:
By the way, this change doesn't apply to `crates-index::Version`, because it doesn't differentiate between targets and their dependencies.

src/rustdoc_gen.rs
```diff
@@ -48,13 +48,17 @@ impl<'a> CrateSource<'a> {
...
-                .dependencies
-                .iter()
-                .filter_map(|(name, dep)| dep.optional().then_some(name))
-                .map(|x| x.to_string())
-                .collect(),
+            Self::ManifestPath { manifest } => {
+                let mut dependencies = manifest.parsed.dependencies.clone();
+                for target in manifest.parsed.target.values() {
+                    dependencies.extend(target.dependencies.clone().into_iter());
+                }
```
epage, line 55: How does enabling target-specific dependencies behave when the target isn't being built?

tonowak:
```
cargo new targetspecific
cd targetspecific
echo "[target.'cfg(windows)'.dependencies]" >> Cargo.toml
echo 'async-std = { version = "1.6.2", optional = true }' >> Cargo.toml
cargo run --features async-std
```
It seems to work on a unix machine.

epage: Seems like that would be a good test to have in the repo documented specifically for this purpose rather than relying only on the third-party reproductions

tonowak: Done in 2c2e7c991f32a8e8ac514e00577787a2adb47d40.

obi1kenobi (approved):

PR #389, opened 2023-02-23
Add test for precompiled binary failing on Windows to CI

mgr0dzicki:
This exposes the issue encountered while testing the new GitHub Action on Windows machine (see current workaround https://github.com/obi1kenobi/cargo-semver-checks-action/pull/21/commits/2d4767cfb82b228e158e7b30b3516859b26f44b4).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,7 @@ jobs:
       - run-on-clap
       - run-on-sqllogictest
       - run-on-ref-slice-fork
+      - run-on-ref-slice-fork-windows
     steps:
       - run: exit 0
 
@@ -612,6 +613,73 @@ jobs:
         run: |
           rm output
           rm -f unexpectedly_did_not_fail
+
+  run-on-ref-slice-fork-windows:
+    # Same as above, but run on a Windows machine and from a compiled binary after
+    # removing .cargo/registry/index. This broke in a workflow of the GitHub Action
+    # using a precompiled binary after a fresh Rust installation, when the
+    # registry index does not exist yet.
+    name: Run cargo-semver-checks on ref-slice fork (Windows)
+    runs-on: windows-latest
+    steps:
+      - name: Checkout cargo-semver-checks
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: true
+
+      - name: Checkout ref-slice fork semver break
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+          repository: 'mgr0dzicki/cargo-semver-action-ref-slice'
+          ref: 'major_change'
+          path: 'subject'
+
+      - name: Install rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: ref-slice-fork-windows
+
+      - name: Build cargo-semver-checks
+        run: cargo build
+
+      - name: Remove the cargo registry index
+        run: Remove-Item -Recurse -Force C:\Users\runneradmin\.cargo\registry\index
+
+      - name: Run cargo-semver-checks
+        continue-on-error: true
+        id: semver_checks
+        run: ./target/debug/cargo-semver-checks.exe semver-checks check-release --manifest-path="subject/Cargo.toml" 2>&1 | tee output
+
+      - name: Check whether it failed
+        if: steps.semver_checks.outcome != 'failure'
+        run: |
+          echo "Error! check-release should have failed because of the breaking change, but it has not."
+          exit 1
+
+      - name: Baseline is correct
+        run: |
+          $EXPECTED = "     Parsing ref_slice v1.2.2 (current)`r`n     Parsing ref_slice v1.2.1 (baseline)`r`n    Checking ref_slice v1.2.1 -> v1.2.2 (patch change)"
+          # Strip ANSI escapes for colors and bold text before comparing.
+          $RESULT = (cat output | grep 'ref_slice v1.' | sed "s,\x1B\[[0-9;]*[a-zA-Z],,g") | Out-String
+          compare $RESULT $EXPECTED
+
+      - name: Semver break found
+        run: |
+          $EXPECTED = "--- failure function_missing: pub fn removed or renamed ---"
+          $RESULT = (cat output | grep failure)
+          compare $RESULT $EXPECTED
+          # Ensure the following fragment (not full line!) is in the output file:
+          grep '  function ref_slice::ref_slice, previously in file' output
+
+      - name: Cleanup
+        run: rm output
 
   init-release:
     name: Run the release workflow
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -898,14 +899,81 @@ jobs:
...
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: ref-slice-fork-windows
+
+      - name: Build cargo-semver-checks
+        run: cargo build
+
+      - name: Remove the cargo registry index
+        run: Remove-Item -Recurse -Force C:\Users\runneradmin\.cargo\registry\index
```
obi1kenobi, line 946: Out of curiosity, why do we need to forcibly remove this directory to cause this problem here? Is there a situation where the directory doesn't already exist even though the user has a working Rust environment installed on their machine?

I'm just trying to figure out whether this is a bug in the GitHub Action, in `cargo-semver-checks` itself, in one of our dependency libraries, or in `cargo`.

mgr0dzicki: The directory is not created during Rust installation, but only after building some crate. Hence if we have a fresh Rust installation and use the downloaded binary of `cargo-semver-checks`, the `index` directory does not exist yet. And this is exactly the case inside GitHub Action.

Here the tool needs to be built first, as we don't want to use the latest release, but the source code which is currently in the repository. But then `index` directory is created during the built, so I remove it afterwards to simulate the environment before building any crate. 

The issue itself is that creating a subdirectory of `index` (for example `index\github.com-1ecc6299db9ec823`) fails, because `index` does not exist. It seems that for some reason it is automatically created along with the desired subdirectory on Ubuntu and MacOS, but not on Windows.

obi1kenobi: Very interesting, nice catch!

.github/workflows/ci.yml
```diff
@@ -898,14 +899,81 @@ jobs:
           cd semver
           rm output
           rm -f unexpectedly_did_not_fail
-
+
```
obi1kenobi, line 902: Assuming you use VSCode, here are a couple of useful settings to put in there:
```
```

.github/workflows/ci.yml
obi1kenobi: Opened https://github.com/frewsxcv/rust-crates-index/issues/97 for this. We should also merge this PR, just to be extra safe.

.github/workflows/ci.yml
```diff
@@ -898,14 +899,81 @@ jobs:
...
+          $RESULT = (cat output | grep 'ref_slice v1.' | sed "s,\x1B\[[0-9;]*[a-zA-Z],,g") | Out-String
+          compare $RESULT $EXPECTED
+
+      - name: Semver break found
+        run: |
+          $EXPECTED = "--- failure function_missing: pub fn removed or renamed ---"
+          $RESULT = (cat output | grep failure)
+          compare $RESULT $EXPECTED
+          # Ensure the following fragment (not full line!) is in the output file:
+          grep '  function ref_slice::ref_slice, previously in file' output
```
obi1kenobi, line 972: I added caching for baseline rustdoc JSON files in the other jobs -- would you mind adding it here too? You'll need to amend the top `EXPECTED` variable since right now it effectively asserts that caching isn't used (`Parsing ref_slice v1.2.1 (baseline)`, not `(baseline, cached)`), and add the new caching steps as in the other jobs.

.github/workflows/ci.yml
obi1kenobi, line 946: Let's add a short explanation with a link to the upstream issue here, to satisfy the curiosity of the next person to read this code. Otherwise this is a pretty strange step to encounter in an action.

obi1kenobi (commented):
Nice find! I opened an issue upstream, and we can work on merging this PR as well.

.github/workflows/ci.yml
mgr0dzicki: Yees, I use VSCode, thank you so much for these!

.github/workflows/ci.yml
mgr0dzicki: Hope it is more clear now! ;)

.github/workflows/ci.yml
```diff
@@ -1125,6 +1160,103 @@ jobs:
...
+        uses: actions/cache/restore@v3
+        with:
+          key: ${{ runner.os }}-${{ hashFiles('.rustc-version', 'semver/**/Cargo.lock') }}-${{ github.job }}-rustdoc
+          path: subject/target/semver-checks/cache
+
+      - name: Restore cargo index and rustdoc target dir
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: |
+            subject/target/semver-checks/local-ref_slice-1_2_2/
```
mgr0dzicki, line 1216: Should it be `subject` or `subject-current`? I think that the first one, but I see that in the above `run-on-ref-slice-fork` job it is actually `subject-current` and I don't really understand why.

obi1kenobi: Good catch! Fixed in #404.

obi1kenobi, line 1216: Any idea if all these paths here and below need to use `\` instead of `/` as a path separator character? This is Windows after all, and I'm not sure whether the various scripts will accommodate the Unix-style paths.

obi1kenobi (commented):
Minor concern about the path separators since I don't know how GitHub Actions works on Windows. But overall this looks great and I'm happy to merge it once you've had a look at the comment and fixed up the CI failure.

obi1kenobi:
Upgrading to `crates-index v0.19.6` should fix the problem: https://github.com/frewsxcv/rust-crates-index/compare/0.19.5...0.19.6

.github/workflows/ci.yml
mgr0dzicki: I changed all `\` to `/` to be extra sure ;)

mgr0dzicki:
I think it's ready to go now! :)

.github/workflows/ci.yml
```diff
@@ -32,6 +32,7 @@ jobs:
       - run-on-ref-slice-fork
       - run-on-tokio-explicit
       - run-on-tokio-implicit
+      - run-on-ref-slice-fork-windows
     steps:
       - run: exit 0
```
obi1kenobi, line 37: I tweaked this bit on `main` last night because it was allowing failing CI jobs to auto-merge. You'll need to merge `main` and add the new job in the new conditional around here as well.

.github/workflows/ci.yml
obi1kenobi: The explanation here is great, but once this PR merges, we probably won't remember why the jobs are separate. We might even merge them together into a matrix, even though you clearly point out how that isn't the right thing to do.

Consider adding this as a comment in the CI file, where it will be harder to miss?

.github/workflows/ci.yml
obi1kenobi: Ah darn, my PR didn't work. Never mind! Let's merge this (though consider adding the comment I suggested in a separate PR) and I'll deal with editing this job in my PR.

.github/workflows/ci.yml
mgr0dzicki: Good idea! Added in https://github.com/obi1kenobi/cargo-semver-checks/pull/407

Issue #396, opened 2023-02-23
The library should return the failed lints


MarcoIeni:
### Describe your use case

I want to use cargo-semver-checks as a library.
Right now the library only tells if the check was successful or not.
I want to know which checks failed and where.

### Describe the solution you'd like

Add failed checks to this struct:
https://github.com/obi1kenobi/cargo-semver-checks/blob/789f130bf7950b6a082260e5895a2004819fe37c/src/lib.rs#L418

### Additional Context

I started working on this.

PR #398, opened 2023-02-23
library: return required update for each crate

MarcoIeni:
First step to add more info to the `Report` struct.
This is a small improvement towards https://github.com/obi1kenobi/cargo-semver-checks/issues/396

Probably in the future you don't even want to store the `required_bump` of each crate, because you could deduce it by iterating on the crate report errors. However, I decided to store it because it seemed like a good first iteration.


src/lib.rs
```diff
@@ -392,35 +393,82 @@ impl Check {
...
             }
         };
-        let success = all_outcomes
-            .into_iter()
-            .fold_ok(true, std::ops::BitAnd::bitand)?;
 
-        Ok(Report { success })
+        Ok(Report {
+            crate_reports: all_outcomes?,
+        })
```
obi1kenobi, line 414: This has a subtle behavior change: the old code runs all the checks *even if some of them error out* because it merges the `Result` values late, whereas the new code relies on `.collect()` which will bail at the first `Err`.

Since the check output is user-visible, this seems net-negative to me? But I could probably be persuaded otherwise.

src/lib.rs
```diff
@@ -392,35 +393,82 @@ impl Check {
...
-    success: bool,
+    /// Collection containing the name and the report of each crate checked.
+    crate_reports: BTreeMap<String, CrateReport>,
 }
 
 impl Report {
     pub fn success(&self) -> bool {
-        self.success
+        let failure = self.crate_reports.values().any(|report| !report.success());
+        !failure
```
obi1kenobi, line 453:
```suggestion
        self.crate_reports.values().all(|report| report.success())
```
Perhaps simpler?

obi1kenobi (commented):
Seems mostly okay, just a couple of minor points.

src/lib.rs
MarcoIeni: Yes, you are right. I addressed this in [1445fb6](https://github.com/obi1kenobi/cargo-semver-checks/pull/398/commits/1445fb644e5f8d29ea26064d902fc3539332a05b)

Unfortunately using a BTreeMap instead of a Vec resulted in uglier code :/

MarcoIeni:
I suggest to view the changes from VSCode GitHub plugin or another tool. GitHub diff is confused by the different indentation I think.

src/lib.rs
```diff
@@ -321,106 +320,143 @@ impl Check {
...
-                            config.verbose(|config| {
-                                config.shell_status(
-                                    "Skipping",
-                                    format_args!("{crate_name} v{version} (current)"),
-                                )
-                            })?;
-                            Ok(true)
-                        } else {
+        // Create a report for each crate.
+        let all_outcomes: anyhow::Result<BTreeMap<String, anyhow::Result<CrateReport>>> =
```
obi1kenobi, line 324: Two concerns here:
- I feel like this code has gotten a bit too complex and intricate. It took me probably 10-15 minutes to figure out what exactly is going on, and that's when I was already familiar with what this code was doing. We're adding significant new complexity here — do you think there's a reasonable refactor we could do to improve the readability in the process?
- The `Result<Map<String, Result<CrateReport>>>` type is quite confusing, especially because it isn't documented what becomes an inner error vs an outer error. It seems that failure to generate the rustdoc becomes an outer error? Is that right? And if so, should that be the case? I'm genuinely not sure.

MarcoIeni: > We're adding significant new complexity here 

I agree. But trying to address [this](https://github.com/obi1kenobi/cargo-semver-checks/pull/398#discussion_r1117872436) is not easy, because you can't simply `?` everywhere when you have a BTreeMap.

> do you think there's a reasonable refactor we could do to improve the readability in the process?

I will try to think about it and change that big Result type.

obi1kenobi (commented):
Overall looks quite good! Sorry for the delay, I just got home from a family vacation.

I just have a small concern around naming / docs in the new `CrateReport` type. Otherwise this should be easy to merge and I can release a new version as soon as that happens.

src/lib.rs
```diff
@@ -392,35 +393,69 @@ impl Check {
...
+
+        Ok(Report { crate_reports })
+    }
+}
+
+/// A report for a single crate.
+#[non_exhaustive]
+#[derive(Debug)]
+pub struct CrateReport {
+    required_bump: Option<RequiredSemverUpdate>,
```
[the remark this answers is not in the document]
MarcoIeni: I took the "required" word of required_bump from the stdout of the binary.
image
So the naming should be ok, I guess 🤔 but I'll add more comments :)

MarcoIeni:
Can you help me understand why CI is failing?

obi1kenobi:
Sorry about the failing CI, @mgr0dzicki was tweaking something in the target project that corresponds to the failing job. The test case here was written quite defensively (in a good way!) but that just means it "noticed" the change. It just needs a small update and I'll merge the fix right away.

obi1kenobi:
It's actually quite concerning that the Windows equivalent of the job _passed_ while the regular Linux one failed. They are testing the same thing, and it makes me feel that the Windows one isn't actually testing it correctly...

MarcoIeni:
all green now, nice, thanks! 😄

obi1kenobi (commented):
Any chance you'd be open to chatting over Zulip / Discord / voice call about this PR, or just pair-programming a bit of it together sometime?

I'd love to add this functionality to the API, and I really appreciate your help with it! I still have concerns about the design, though, and I'm worried that proceeding one iteration at a time via fully-async GitHub comments isn't going to be effective and we'll both just end up frustrated with the process.

MarcoIeni:
Some notes from the call with Predrag:

We should report back two bumps:
- the detected one (the bump between your local version and the published one)
- the required one (the bump required by semver check, for example if there are breaking changes or not)

These bumps are around [here](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/src/check_release.rs#L102-L111):
- version_change is the detected one
- release_type is the user override. For example, they can use it to exclude minor changes.

The problem with this pr is that at the moment `successful` means that there are no semver violations. However, if you bumped your version from 1.0.0 to 2.0.0, the semver check is successful even if there are semver violations.

`successfull = detected bump > required bump`.

src/lib.rs
```diff
@@ -321,7 +322,10 @@ impl Check {
 
         // Create a report for each crate.
         // We want to run all the checks, even if one returns `Err`.
-        let all_outcomes: Vec<anyhow::Result<(String, CrateReport)>> = match &self.current.source {
+        let all_outcomes: Vec<anyhow::Result<(String, Option<CrateReport>)>> = match &self
+            .current
+            .source
+        {
```
obi1kenobi, line 328: Might your local setup have a custom rustfmt config or something? I'm surprised by this unrelated change.

src/lib.rs
```diff
@@ -427,28 +433,45 @@ impl Check {
...
     required_bump: ReleaseType,
 }
 
 impl CrateReport {
     /// Construct a new `CrateReport` with a successful outcome.
-    pub(crate) fn new_successful() -> Self {
+    pub(crate) fn new(detected_bump: ActualSemverUpdate) -> Self {
         Self {
+            detected_bump,
             required_bump: ReleaseType::Patch,
```
obi1kenobi: You're probably still in the process of tweaking things, so this note is more for me: we should remember to change `required_bump` here. Just because the run was successful doesn't mean `required_bump: ReleaseType::Patch` is correct. Either having it as an `Option` type or having it match `ActualSemverUpdate` might be best. We don't know the exact minimum required bump -- if we detected a minor bump, we don't run any of the minor-requiring checks and don't know whether minor was overkill or the minimal appropriate one.

obi1kenobi (commented):
I know you're probably still in the middle of refactoring here, so I won't review again until you hit the "request review" button. But a couple of things caught my eye and I wanted to make a note of them before I forgot.

src/lib.rs
MarcoIeni: > having it match ActualSemverUpdate might be best.

we can't do it, because there's no equivalent of `ActualSemverUpdate::NotChanged`.
I turned it into an Option 👍 

Btw, thanks for your early review. Really helpful.

src/lib.rs
MarcoIeni: I don't think I have a custom setup

obi1kenobi: Ah I missed that, my bad.

src/lib.rs
```diff
@@ -394,35 +399,94 @@ impl Check {
...
+
+/// Report of semver check of one crate.
+#[non_exhaustive]
+#[derive(Debug)]
+pub struct CrateReport {
+    /// Bump between the current version and the baseline one.
+    detected_bump: ActualSemverUpdate,
+    /// Minimum bump required to respect semver.
+    /// For example, if the crate contains breaking changes, this is [`ReleaseType::Major`].
+    /// If no bump is required, this is [`Option::None`].
```
obi1kenobi, line 440: Is this accurate?

Say we ran a semver-check over a crate where the detected bump is `ActualSemverUpdate::Major`. Would `required_bump` actually be `Some(ReleaseType::Major)`?

If no bump is required, the value will be `None`, but is that the only time? I think there are other situations where the value might be `None`.

Let's back this up with some tests? Consider making a new integration test file in the `tests/` directory to pin down the behavior here.

test_crates/trait_missing_with_major_bump/new/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+publish = false
+name = "trait_missing"
```
obi1kenobi, line 3:
```suggestion
name = "trait_missing_with_major_bump"
```

test_crates/trait_missing_with_major_bump/old/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+publish = false
+name = "trait_missing"
```
obi1kenobi, line 3:
```suggestion
name = "trait_missing_with_major_bump"
```

MarcoIeni:
Tests fail for me locally, with this weird error:
```

---- query::tests_lints::trait_unsafe_removed stdout ----
thread 'query::tests_lints::trait_unsafe_removed' panicked at 'Once instance has previously been poisoned', /Users/marcoieni/.cargo/registry/src/index.crates.io-6f17d22bba15001f/lazy_static-1.4.0/src/inline_lazy.rs:30:16

---- query::tests_lints::trait_unsafe_added stdout ----
thread 'query::tests_lints::trait_unsafe_added' panicked at 'Once instance has previously been poisoned', /Users/marcoieni/.cargo/registry/src/index.crates.io-6f17d22bba15001f/lazy_static-1.4.0/src/inline_lazy.rs:30:16

---- query::tests_lints::unit_struct_changed_kind stdout ----
thread 'query::tests_lints::unit_struct_changed_kind' panicked at 'Once instance has previously been poisoned', /Users/marcoieni/.cargo/registry/src/index.crates.io-6f17d22bba15001f/lazy_static-1.4.0/src/inline_lazy.rs:30:16

---- query::tests_lints::variant_marked_non_exhaustive stdout ----
thread 'query::tests_lints::variant_marked_non_exhaustive' panicked at 'Once instance has previously been poisoned', /Users/marcoieni/.cargo/registry/src/index.crates.io-6f17d22bba15001f/lazy_static-1.4.0/src/inline_lazy.rs:30:16

---- query::tests_lints::type_marked_deprecated stdout ----
thread 'query::tests_lints::type_marked_deprecated' panicked at 'Once instance has previously been poisoned', /Users/marcoieni/.cargo/registry/src/index.crates.io-6f17d22bba15001f/lazy_static-1.4.0/src/inline_lazy.rs:30:16

---- query::tests_lints::tuple_struct_to_plain_struct stdout ----
thread 'query::tests_lints::tuple_struct_to_plain_struct' panicked at 'Once instance has previously been poisoned', /Users/marcoieni/.cargo/registry/src/index.crates.io-6f17d22bba15001f/lazy_static-1.4.0/src/inline_lazy.rs:30:16


failures:
    query::tests::all_queries_are_valid
    query::tests::pub_use_handling
    query::tests_lints::auto_trait_impl_removed
    query::tests_lints::constructible_struct_adds_field
    query::tests_lints::constructible_struct_adds_private_field
    query::tests_lints::constructible_struct_changed_type
    query::tests_lints::derive_trait_impl_removed
    query::tests_lints::enum_marked_non_exhaustive
    query::tests_lints::enum_missing
    query::tests_lints::enum_must_use_added
    query::tests_lints::enum_repr_c_removed
    query::tests_lints::enum_repr_int_changed
    query::tests_lints::enum_repr_int_removed
    query::tests_lints::enum_struct_variant_field_added
    query::tests_lints::enum_struct_variant_field_missing
    query::tests_lints::enum_tuple_variant_field_added
    query::tests_lints::enum_tuple_variant_field_missing
    query::tests_lints::enum_variant_added
    query::tests_lints::enum_variant_missing
    query::tests_lints::function_const_removed
    query::tests_lints::function_missing
    query::tests_lints::function_must_use_added
    query::tests_lints::function_parameter_count_changed
    query::tests_lints::function_unsafe_added
    query::tests_lints::inherent_method_const_removed
    query::tests_lints::inherent_method_missing
    query::tests_lints::inherent_method_must_use_added
    query::tests_lints::inherent_method_unsafe_added
    query::tests_lints::method_parameter_count_changed
    query::tests_lints::sized_impl_removed
    query::tests_lints::struct_marked_non_exhaustive
    query::tests_lints::struct_missing
    query::tests_lints::struct_must_use_added
    query::tests_lints::struct_pub_field_missing
    query::tests_lints::struct_repr_c_removed
    query::tests_lints::struct_repr_transparent_removed
    query::tests_lints::struct_with_pub_fields_changed_type
    query::tests_lints::trait_missing
    query::tests_lints::trait_must_use_added
    query::tests_lints::trait_unsafe_added
    query::tests_lints::trait_unsafe_removed
    query::tests_lints::tuple_struct_to_plain_struct
    query::tests_lints::type_marked_deprecated
    query::tests_lints::unit_struct_changed_kind
    query::tests_lints::variant_marked_non_exhaustive

test result: FAILED. 19 passed; 45 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

MarcoIeni:
Deleting the `localdata` folder helped. I was seeing weird errors like:
```

---- query::tests_lints::function_const_removed stdout ----
thread 'query::tests_lints::function_const_removed' panicked at 'Once instance has previously been poisoned', /Users/marcoieni/.cargo/registry/src/index.crates.io-6f17d22bba15001f/lazy_static-1.4.0/src/inline_lazy.rs:30:16
```
Now I see the same error locally 👍 
Consider adding instructions to delete the `localdata` folder to this error: `Also, remember about running ./scripts/regenerate_test_rustdocs.sh when needed.`

MarcoIeni:
can you help me fixing this last error? 😅

obi1kenobi:
I'm not near a computer at the moment, but look for an error near the top
of the stack of messages -- or just rerun a single test.

Something panicked and poisoned the mutex, and that then caused the errors
you showed here.

MarcoIeni:
I fixed that error thanks to your previous help 👍 
now there's another [error](https://github.com/obi1kenobi/cargo-semver-checks/actions/runs/4865060452/jobs/8674882462):

```
---- query::tests_lints::trait_unsafe_removed stdout ----
thread 'query::tests_lints::trait_unsafe_removed' panicked at '
Query trait_unsafe_removed produced incorrect output (./src/lints/trait_unsafe_removed.ron).

Expected output (./test_outputs/trait_unsafe_removed.output.ron):
{
    "./test_crates/trait_missing/": [
        {
            "name": String("TraitStopsBeingUnsafe"),
            "path": List([
                String("trait_missing"),
                String("TraitStopsBeingUnsafe"),
            ]),
            "span_begin_line": Uint64(6),
            "span_filename": String("src/lib.rs"),
            "visibility_limit": String("public"),
        },
    ],
    "./test_crates/trait_unsafe_removed/": [
        {
            "name": String("TraitBecomesSafe"),
            "path": List([
                String("trait_unsafe_removed"),
                String("TraitBecomesSafe"),
            ]),
            "span_begin_line": Uint64(2),
            "span_filename": String("src/lib.rs"),
            "visibility_limit": String("public"),
        },
    ],
}

Actual output:
{
    "./test_crates/trait_missing/": [
        {
            "name": String("TraitStopsBeingUnsafe"),
            "path": List([
                String("trait_missing"),
                String("TraitStopsBeingUnsafe"),
            ]),
            "span_begin_line": Uint64(6),
            "span_filename": String("src/lib.rs"),
            "visibility_limit": String("public"),
        },
    ],
    "./test_crates/trait_missing_with_major_bump/": [
        {
            "name": String("TraitStopsBeingUnsafe"),
            "path": List([
                String("trait_missing_with_major_bump"),
                String("TraitStopsBeingUnsafe"),
            ]),
            "span_begin_line": Uint64(6),
            "span_filename": String("src/lib.rs"),
            "visibility_limit": String("public"),
        },
    ],
    "./test_crates/trait_unsafe_removed/": [
        {
            "name": String("TraitBecomesSafe"),
            "path": List([
                String("trait_unsafe_removed"),
                String("TraitBecomesSafe"),
            ]),
            "span_begin_line": Uint64(2),
            "span_filename": String("src/lib.rs"),
            "visibility_limit": String("public"),
        },
    ],
}

Note that the individual outputs might have been deliberately reordered.

Also, remember about running ./scripts/regenerate_test_rustdocs.sh when needed.
', src/query.rs:400:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/071f14baae931e17a5a99366bf216e76384cc4f6/library/std/src/panicking.rs:[578](https://github.com/obi1kenobi/cargo-semver-checks/actions/runs/4865060452/jobs/8674882462#step:9:579):5
   1: core::panicking::panic_fmt
             at /rustc/071f14baae931e17a5a99366bf216e76384cc4f6/library/core/src/panicking.rs:67:14
   2: cargo_semver_checks::query::tests::check_query_execution
             at ./src/query.rs:400:13
   3: cargo_semver_checks::query::tests_lints::trait_unsafe_removed
             at ./src/query.rs:442:21
   4: cargo_semver_checks::query::tests_lints::trait_unsafe_removed::{{closure}}
             at ./src/query.rs:441:28
   5: core::ops::function::FnOnce::call_once
             at /rustc/071f14baae931e17a5a99366bf216e76384cc4f6/library/core/src/ops/function.rs:250:5
   6: core::ops::function::FnOnce::call_once
             at /rustc/071f14baae931e17a5a99366bf216e7[638](https://github.com/obi1kenobi/cargo-semver-checks/actions/runs/4865060452/jobs/8674882462#step:9:639)4cc4f6/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    query::tests_lints::trait_missing
    query::tests_lints::trait_unsafe_added
    query::tests_lints::trait_unsafe_removed

test result: FAILED. 61 passed; 3 failed; 0 ignored; 0 measured; 0 filtered out; finished in 8.02s
```

obi1kenobi:
All lints run on all test crates, and it seems that your new test crate
causes a few more lints to flag issues. Just review the issues they raise
and if everything looks good, then update the expected output files :)

MarcoIeni:
Thanks, I think I have done it 👍 
Please, consider using something like [insta](https://github.com/mitsuhiko/insta) or [expect_test](https://github.com/rust-analyzer/expect-test) to improve the developer experience.

obi1kenobi:
I'd happily look at a PR for either `insta` or `expect_test`, whichever you think is better! Last I looked, neither was a clear improvement since our tests aren't just snapshot but also require running the rustdoc generation script first, so it seemed like any effort dedicated to making the experience better would be equally (or better) spent on making the existing approach better.

Would love to be proven wrong on this!

src/lib.rs
```diff
@@ -394,35 +398,98 @@ impl Check {
...
 
+/// Report of semver check of one crate.
+#[non_exhaustive]
+#[derive(Debug)]
+pub struct CrateReport {
+    /// Bump between the current version and the baseline one.
+    detected_bump: ActualSemverUpdate,
+    /// Minimum bump required to respect semver.
+    /// For example, if the crate contains breaking changes, this is [`ReleaseType::Major`].
+    /// If no bump is required, this is [`Option::None`].
```
obi1kenobi, line 439: Perhaps
```suggestion
    /// If no additional bump beyond the already-detected one is required, this is [`Option::None`].
```
or something similar?

src/lib.rs
```diff
@@ -394,35 +398,98 @@ impl Check {
...
 }
 
+/// Report of semver check of one crate.
+#[non_exhaustive]
+#[derive(Debug)]
+pub struct CrateReport {
+    /// Bump between the current version and the baseline one.
+    detected_bump: ActualSemverUpdate,
+    /// Minimum bump required to respect semver.
+    /// For example, if the crate contains breaking changes, this is [`ReleaseType::Major`].
```
obi1kenobi, line 438: Small suggested clarification + `Some` to match type:
```suggestion
    /// Minimum additional bump (on top of `detected_bump`) required to respect semver.
    /// For example, if the crate contains breaking changes, this is [`Some(ReleaseType::Major)`].
```

src/lib.rs
```diff
@@ -394,35 +398,98 @@ impl Check {
...
+            Some(required_bump) => {
+                match self.detected_bump {
+                    // If user bumped the major version, any breaking change is accepted.
+                    ActualSemverUpdate::Major => true,
+                    ActualSemverUpdate::Minor => {
+                        matches!(required_bump, ReleaseType::Minor | ReleaseType::Patch)
+                    }
+                    ActualSemverUpdate::Patch | ActualSemverUpdate::NotChanged => {
+                        required_bump == ReleaseType::Patch
+                    }
```
obi1kenobi, line 458: Does this need updating for the changes to the semantics of the struct's fields? It seems to me like some of these could be `assert` instead of comparisons.

src/lib.rs
```diff
@@ -394,35 +398,98 @@ impl Check {
...
+                }
+            }
+        }
+    }
+
+    /// Minimum bump required to respect semver.
+    /// It's [`Option::None`] if no bump is required.
```
obi1kenobi, line 465:
```suggestion
    /// It's [`Option::None`] if no bump is required beyond the already-detected bump.
```

src/lib.rs
```diff
@@ -394,35 +398,98 @@ impl Check {
...
 }
 
 impl Report {
+    /// `true` if none of the crates violate semver.
     pub fn success(&self) -> bool {
-        self.success
+        self.crate_reports.values().all(|report| report.success())
+    }
+
+    pub fn crate_reports(&self) -> &BTreeMap<String, CrateReport> {
```
obi1kenobi, line 491: Please add a doc comment here and in general on all new `pub` items.

src/query.rs
```diff
@@ -21,8 +21,17 @@ impl RequiredSemverUpdate {
...
+        match value {
+            RequiredSemverUpdate::Major => Self::Major,
+            RequiredSemverUpdate::Minor => Self::Minor,
+        }
+    }
+}
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub(crate) enum ActualSemverUpdate {
+pub enum ActualSemverUpdate {
```
obi1kenobi, line 34: Would you mind adding a doc comment here, since this type is new to the public API?

test_outputs/trait_missing_with_major_bump.output.ron
obi1kenobi, line 1: Are you sure this file is necessary? I don't recall the exact setup of the test suite, but I think the output files are per-lint, not per-test-crate, and `trait_missing_with_major_bump` is a test crate, not a lint. So that means its data can appear in lints' output files, but I don't think anything will read this file since it doesn't correspond to any lint.

obi1kenobi (commented):
The design looks good! Just a few tweaks to doc comments and this is ready to merge 🎉

test_outputs/trait_missing_with_major_bump.output.ron
MarcoIeni: removed in 54042b5

src/lib.rs
MarcoIeni: done in cf47884

src/lib.rs
MarcoIeni: Yes, you are right. I think it should never happen that the update of `detected_bump` is the same of `required_bump`, right?
Maybe we can add some warnings in this case. Or do you prefer to write an assert? Sounds scary 😅

obi1kenobi: Let's assert. If our design behaves unexpectedly, it should fail loudly and quickly so we can figure out the failure mode and fix it.

I think `required_bump` should never be equal to or lesser than `detected_bump` in this design. If we're wrong about that, there's some serious issue we've overlooked and the outcome shouldn't be trusted anyway.

MarcoIeni: I added the assert in [9cf72ee](https://github.com/obi1kenobi/cargo-semver-checks/pull/398/commits/9cf72ee870d0545448cc95e9806bc591c7b41813)

obi1kenobi (approved):
Looks great, let's ship it! Thanks for all the work on this, and sorry reviews took a while. I'm hoping my schedule calms down a bit going forward.

Also -- this found an ICE in nightly Rust. @MarcoIeni would you care to report it?

obi1kenobi (approved):
Looks great, let's ship it! Thanks for all the work on this, and sorry reviews took a while. I'm hoping my schedule calms down a bit going forward.

Also -- this found an ICE in nightly Rust. @MarcoIeni would you care to report it?

MarcoIeni:
> would you care to report it?

somebody already reported it in https://github.com/rust-lang/rust/issues/111320

> I'm hoping my schedule calms down a bit going forward.

No problem at all, thanks for the very detailed review :)

PR #399, opened 2023-02-23
Use synthetic mtime so the rustdoc build cache gets used.

obi1kenobi:


obi1kenobi:
This didn't seem to work -- the "current" rustdoc generation recompiled everything instead of printing out `Fresh` lines, even though the mtimes of the built artifacts were set to be in the future. I'm not sure why it didn't work, and I don't think I can continue digging into it right now.

Issue #401, opened 2023-02-24
Caching-friendly generation of "current" rustdoc JSON


obi1kenobi:
### Describe your use case

For workspaces that semver-check many crates, generating the "current" rustdoc JSON takes a significant portion of time: [almost an hour across all of `rust-libp2p`](https://github.com/libp2p/rust-libp2p/pull/3469#discussion_r1116373767), for example.

To ensure rustdoc JSON stability and correctness, "current" (also "baseline") rustdoc is generated by:
- making a synthetic project with a path dependency on the target crate
- generating the rustdoc for that path dependency

This synthetic project is generated at `cargo-semver-checks` runtime. Because of https://github.com/rust-lang/cargo/issues/6529, cached build artifacts for that project can't be used: cargo will ignore them since their mtime is older than the project. The hack of applying a synthetic mtime in the future to the build artifacts didn't seem to work, for unknown reasons: #399

It would appear that build artifact caching for the current rustdoc JSON is just not possible at the moment. It would be nice to make that a supported use case.

### Describe the solution you'd like

- A mechanism that allows caching of build artifacts of "current" rustdoc JSON builds
- Use of that mechanism by default in the official `cargo-semver-checks` [GitHub Action](https://github.com/obi1kenobi/cargo-semver-checks-action)

### Alternatives, if applicable

It may be possible to do this as a step-by-step process:
- Generate the current rustdoc separately, e.g. via a separate subcommand or tool:
  - Generate the synthetic project with the path dependency, e.g. via the tool.
  - Then, say via a manual script, update the mtimes of the generated project.
  - Then, either via the tool or via a manual script, generate the needed rustdoc from the synthetic project.
- Then use this rustdoc in `cargo-semver-checks` via the `--current-rustdoc <path>` CLI option.

This is certainly not as elegant, but may be substantially faster for large repos.

thomaseizinger:
What is the rationale for the synthetic project?

obi1kenobi:
It's due to issues like this one: #167

TL;DR: The user's copy of their project is not guaranteed to be in a state amenable to valid rustdoc generation. To get it there we need a fresh lockfile -- but we don't want to overwrite the user's own lockfile, so we make a synthetic project and lock there instead.

It's a really unfortunate but really serious problem, because it can trivially lead to all the flavors of semver-major false-positives one could imagine.

thomaseizinger:
[...]

PR #402, opened 2023-02-24
Reuse existing target directory

thomaseizinger:
Previously, each synthetic project got its own target directory. That results in a lot of recompilation. Because the entire workspace of `cargo semver-checks` is within the original projects `target` directory, we can simply point the `cargo doc` process to that one.

This allows us to reuse a lot of compilation artifacts up to the point where an invocation of `cargo semver-checks` only takes a couple of seconds if the user already has a debug build locally or a cache in CI.

Originally, I thought that walking up two directories is a bit of a hack but it is actually extremely stable in my testing. It works with custom target dirs (`CARGO_TARGET_DIR`) because our entire workspace is within the parent project's `target` dir.

Thoughts?


obi1kenobi:
Thanks for the ideas and PR!

Let's keep the discussion here focused on the specific implementation of the idea, and discuss the idea itself in #401. I'll switch to the issue so we can sort that out first.

Issue #412, opened 2023-03-02
Feature compatibility verification


HeroicKatora:
### Describe your use case

Feature flags can be used in rustc to enable and disable arbitrary chunks of code via cfg. Due to the _additive_ nature of features during the dependency resolution of cargo, however, in many cases they are supposed to behave in a compatible manner. 

For instance, consider a crate with an API for no-std and some utilities for `std` use cases. (As implied by this construction, there is usually a feature `std` and _not_ a feature `no_std` since adding the latter in a dependency can't be undone by a sibling dependency.) Concretely the crate may:

* Have an error type which implements `std::error::Error` gated behind `cfg(feature = "std")`.
* Offer a method `parse_str(_: &str)` and additional a method `parse<R: std::io::Read>` gated behind the feature flag.

Now, a consumer may access the crate `dep-cfg` via two unrelated sibling dependencies, `dep-a` and `dep-b`. A common example may be `serde_json` as a wide-spread format. Then dependency `dep-a` may require the `std` feature while the other does not. Also `dep-b` is usually only tested against the crate _without_ the feature. The expectation of additive features is now that `dep-cfg`'s interface is only _augmented_ by adding the `std` feature but it is not broken. Note that the sibling dependency `dep-a: dep-cfg` propagates the `std` feature flag to the crate compiled for `dep-b`, i.e. `dep-cfg` is only compiled _once_ with the aggregate of all enabled features. For a stable ecosystem, `dep-b` should be able to rely on its small test suite even in the presence of unknown downstream users such as `dep-a`.

This is the same problem as compatibility checks between compatible versions.

### Describe the solution you'd like

Add support for comparing explicit feature configurations.

* A new subcommand `check-features` with flags `--baseline-features` and `--features`

Then the above check may be implemented in CI as follows:

```base
cargo-semver-checks semver-checks check-features --baseline-features "" --features "std"
```

The semantics would be similar to using `--baseline-root .` in the existing subcommand check-release. It may be implemented in terms of this subcommand.

obi1kenobi:
Thanks for the detailed writeup, much appreciated!

This is something I've also considered, and that would be awesome to have. I think the majority of lints would be usable mostly as-is, perhaps just with tweaked error messages.

[...]

PR #426, opened 2023-04-10
Add ability to specify features in CLI

### LINKED ISSUE #181: Enable specifying features for current and baseline
### Describe your use case

I have a crate which has a feature-flag `unsupported` which does not guarantee semver compatibility. This might not be the best contract but there's no way to make `cargo semver-checks check-release` not enable `--all-features`.

```bash
cargo semver-checks check-release --exclude-feature unsupported
```

this would enable all features, except `unsupported`

```bash
cargo semver-checks check-release --features foo,bar,new_feature --baseline-features foo,bar
```

could also work, but that would cause problems in CI as you wouldn't be able to easily specify what features are new or not. E.g in the example, `new_feature` is not in the baseline, but is part of the api

### Additional Context

_No response_

### LINKED ISSUE #321: Generating rustdoc with `--all-features` exposes APIs in an unstable feature

Running on https://github.com/sdroege/ebur128 gives the following error

### Actual Behaviour

```console
$ cargo semver-checks check-release
    Updating index
     Parsing ebur128 v0.1.7 (current)
     Parsing ebur128 v0.1.7 (baseline)
    Checking ebur128 v0.1.7 -> v0.1.7 (no change)
   Completed [   0.117s] 40 checks; 39 passed, 1 failed, 0 unnecessary

--- failure struct_missing: pub struct removed or renamed ---

Description:
A publicly-visible struct cannot be imported by its prior path. A `pub use` may have been removed, or the struct itself may have been renamed or removed entirely.
       impl: https://github.com/obi1kenobi/cargo-semver-check/tree/v0.16.1/src/queries/struct_missing.ron

Failed in:
  struct ebur128::interp::Interp2F, previously in file ~/.cargo/registry/src/github.com-1ecc6299db9ec823/ebur128-0.1.7/src/interp.rs:161
  struct ebur128::interp::Interp4F, previously in file ~/.cargo/registry/src/github.com-1ecc6299db9ec823/ebur128-0.1.7/src/interp.rs:162
       Final [   0.118s] semver requires new major version: 1 major and 0 minor checks failed
```

### Expected Behaviour

This should pass because the type is not exported from the crate

#### Software version

cargo-semver-checks 0.16.1

#### Operating system

Linux 6.0.18-300.fc37.x86_64

#### Command-line

```bash
~/.cargo/bin/cargo-semver-checks semver-checks --bugreport 
```

#### cargo version

```
> cargo -V 
cargo 1.66.0 (d65d197ad 2022-11-15)
```

#### Compile time information

- Profile: release
- Target triple: x86_64-unknown-linux-gnu
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2
- Host: x86_64-unknown-linux-gnu

### Build Configuration

_No response_

### Additional Context

_No response_

staniewzki:
Resolves #321.
Resolves #181.


tonowak:
:fire:

src/rustdoc_gen.rs
```diff
@@ -97,12 +97,33 @@ impl<'a> CrateSource<'a> {
         all_crate_features.extend(self.regular_features());
         all_crate_features.into_iter().collect()
     }
+
+    fn features_not_on_blacklist(&self) -> Vec<String> {
```
obi1kenobi, line 101: "Blacklist" as a term has a bit of a problematic history and unfortunate connotations, and also isn't sufficiently descriptive here since it doesn't specify how features may end up "blacklisted."

Try `heuristically_included_features()` or something to that effect, plus a generous doc comment on the function that describes how those features are selected?

src/rustdoc_gen.rs
```diff
@@ -189,11 +210,30 @@ pub(crate) enum CrateType<'a> {
     },
 }
 
+#[derive(Debug, Clone)]
+pub(crate) struct FeatureConfig {
+    pub(crate) default_features: bool,
+    pub(crate) only_explicit_features: bool,
+    pub(crate) feature: Vec<String>,
+    pub(crate) all_features: bool,
+}
```
obi1kenobi, line 219: Please document what these fields mean and provide a few examples that clarify how these values combine. For example, what does it mean to have `default_features = true` together with `feature` set to a non-empty list?

Also, since the clap interface says a few of these conflict with each other, might it be possible that an enum would be a better representation here?

obi1kenobi (commented):
Neat! Excited to see this ship!

obi1kenobi:
Another two things to consider:
- What happens if a new feature is added in "current" but the same feature doesn't exist in "baseline"? Attempting to generate baseline rustdoc with that feature enabled would fail, since the feature isn't defined there.
- Is there a possible and easy tweak of the CLI that would let us support the use case in #412?

For both of these, please don't implement anything yet and let's just discuss possible options and designs first. It'll be harder to think of good solutions if you've already gone ahead and implemented a particular approach.

With that in mind, what do you think?

src/main.rs
```diff
@@ -192,6 +193,30 @@ struct CheckRelease {
     )]
     release_type: Option<ReleaseType>,
 
+    /// Use only the default features.
+    #[arg(
+        long,
+        help_heading = "Features",
+        conflicts_with_all = ["feature", "only_explicit_features"]
```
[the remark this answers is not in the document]
staniewzki: Okay, this makes sense. Also adjusting the info comment then.

staniewzki:
If a feature doesn't exist in "baseline", it could just be ignored. This seems quite easy to implement.

For the mentioned use case, maybe we could add another flag, that adds features to the "current" version, something like --feature-current? I think it makes more sense this way, as the ususal use case is to test the crates on the same feature set. Also, you never want "baseline" feature set to be a superset of "current" feature set, right?

obi1kenobi:
> If a feature doesn't exist in "baseline", it could just be ignored. This seems quite easy to implement.

This seems reasonable to me, though we should probably print a warning message about it to make sure the user is aware.

> For the mentioned use case, maybe we could add another flag, that adds features to the "current" version, something like --feature-current? I think it makes more sense this way, as the ususal use case is to test the crates on the same feature set. Also, you never want "baseline" feature set to be a superset of "current" feature set, right?

"Never" is a very strong word, I wouldn't say never. For example, maybe a library had an optional feature in v1 that is included by default in v1.1, and wants to check that the API is unchanged between v1+feature and v1.1 without the feature.

If we're adding a flag to affect the features in only one of current/baseline, we should add the symmetrical flag for the other one too.

A few other, smaller things:
- `cargo` uses the plural "features" for its feature-related flags (e.g. in `cargo add`), so for consistency let's use `features` as well and make it work the same way as in `cargo add`. Since we're planning on merging into cargo, it's important that we minimize friction like this.
- Our other flags that are split between current/baseline use the word "current" or "baseline" to start, so let's call the flags `--baseline-features` and `--current-features` instead of using "current" or "baseline" as a suffix.
- For the feature-related flags that affect both current and baseline, let's make sure to explicitly document this fact in their help text.

staniewzki:
@obi1kenobi how do I deal with `clippy::too_many_argument` lint on `generate_versioned_crate` function? I know that having that many arguments is not ideal, but I don't think that creating a struct to pack all the parameters there makes much more sense.

obi1kenobi:
> @obi1kenobi how do I deal with `clippy::too_many_argument` lint on `generate_versioned_crate` function? I know that having that many arguments is not ideal, but I don't think that creating a struct to pack all the parameters there makes much more sense.

Making a single struct to pack everything doesn't make sense, but perhaps we can group the arguments somehow? For example, consider whether any arguments are always used together with each other: if the function takes an `a` and `b`, and then whenever it passes `a` somewhere it also passes `b`, then maybe `a` and `b` should be combined.

Then see if it makes sense to move one inside the other and if so, in which direction. Or if not, see if it makes sense to make a struct that holds both of them.

This is a skill that takes practice to develop, so I don't want to do you the disservice of just giving you the answer and robbing you of an excellent opportunity to practice the skill in a safe space.

src/main.rs
```diff
@@ -261,6 +315,22 @@ impl From<CheckRelease> for cargo_semver_checks::Check {
...
+        } else if value.only_explicit_features {
+            check.with_only_explicit_features();
+        } else {
+            check.with_heuristically_included_features();
+        }
+        let mut mutual_features = value.features;
+        let mut current_features = value.current_features;
+        let mut baseline_features = value.baseline_features;
+        current_features.append(&mut mutual_features.clone());
+        baseline_features.append(&mut mutual_features);
```
obi1kenobi, line 331: `.append()` taking a `&mut` seems very strange. What's the reasoning for choosing it, versus taking the argument by value or as read-only ref?

src/rustdoc_gen.rs
```diff
@@ -110,6 +110,65 @@ impl<'a> CrateSource<'a> {
         all_crate_features.extend(self.regular_features());
         all_crate_features.into_iter().collect()
     }
+
+    /// Some feature names usually mean, that the feature is breaking semver by design.
+    /// This function filters them out.
```
obi1kenobi, line 115: Probably a good idea to reference the following cargo issues here:
- unstable/nightly-only features: https://github.com/rust-lang/cargo/issues/10881
- private/hidden features: https://github.com/rust-lang/cargo/issues/10882
and explain how this heuristic relates to those. Most users probably haven't needed or used features that don't adhere to semver, so we should explain a bit of the context here so they understand what's going on.

For example:
```
Sometimes crates include features that are not meant for public use or otherwise don't adhere to semver:
- private features, like `bench` used for internal benchmarking only
- nightly-only features, like `nightly`
- unstable features containing experimental code, like `unstable`

To ensure the best possible out-of-the-box user experience, this function attempts to heuristically exclude feature names like the above.

This function uses heuristics based on the name since cargo does not currently include a mechanism for marking features as private/hidden/unstable. When such mechanisms are available in cargo, we'll update this functionality to make use of them. Relevant cargo issues:
- unstable/nightly-only features: https://github.com/rust-lang/cargo/issues/10881
- private/hidden features: https://github.com/rust-lang/cargo/issues/10882
```

Then include the description of the heuristic used etc.

src/rustdoc_gen.rs
```diff
@@ -110,6 +110,65 @@ impl<'a> CrateSource<'a> {
...
+        ]);
+
+        let determine = |feature_name: &String| {
+            !features_ignored_by_default.contains(feature_name) && !feature_name.starts_with("__")
+        };
+
+        self.all_features().into_iter().filter(determine).collect()
+    }
+
+    pub(crate) fn feature_list_from_config(
```
obi1kenobi, line 138: Can you document what this function does? It has some fairly complex logic inside and I don't really understand what's going on or what it returns in each of the edge cases inside.

src/rustdoc_gen.rs
```diff
@@ -110,6 +110,65 @@ impl<'a> CrateSource<'a> {
...
+        ]
+        .concat();
+
+        result
+            .into_iter()
+            .filter(|feature_name| {
+                if !all_features.contains(feature_name) && feature_config.ignore_non_existing {
+                    global_config
+                        .shell_warn(format!(
+                            "Feature {feature_name} is not present in the baseline."
```
obi1kenobi, line 162:
```suggestion
                            "Feature `{feature_name}` is not present in the baseline."
```

src/rustdoc_gen.rs
```diff
@@ -110,6 +110,65 @@ impl<'a> CrateSource<'a> {
...
+                        ))
+                        .expect("print failed");
+                    false
```
obi1kenobi, line 165: How does this code know that it's being called in a "baseline" context? This doesn't seem clear or obvious to me.

src/rustdoc_gen.rs
```diff
@@ -123,11 +182,48 @@ pub(crate) enum CrateType<'a> {
     },
 }
 
+#[derive(Debug, Clone, Hash)]
+pub(crate) struct FeatureConfig {
+    pub(crate) base_features: BaseFeatures,
+    pub(crate) extra_features: Vec<String>,
+    pub(crate) ignore_non_existing: bool,
+}
```
obi1kenobi, line 190: Can you document what this type represents and what each of its fields means? It isn't clear to me what the difference between a "base" feature and an "extra" feature is, or when `ignore_non_existing` applies.

src/rustdoc_gen.rs
```diff
@@ -123,11 +182,48 @@ pub(crate) enum CrateType<'a> {
...
+
+#[derive(Debug, Clone, Hash)]
+pub(crate) enum BaseFeatures {
+    All,
+    Default,
+    Heuristic,
+    None,
+}
```
obi1kenobi, line 198: Does this "base" here refer to "baseline", or is it completely unrelated? Can we come up with a term that won't be confusing?

src/rustdoc_gen.rs
```diff
@@ -123,11 +182,48 @@ pub(crate) enum CrateType<'a> {
...
+
+impl FeatureConfig {
+    pub(crate) fn default(is_baseline: bool) -> Self {
```
obi1kenobi, line 201: This is a very unfortunate name for a function. Users seeing `Foo::default()` will assume it's `Default::default()` which is definitely not the case here. Also consider how `FeatureConfig::default(true)` looks — it very much isn't obvious what that `true` means, and also doesn't seem very default if it's already being configured.

Consider splitting this into separate methods for the `true/false` cases of this input argument.

src/rustdoc_gen.rs
```diff
@@ -123,11 +182,48 @@ pub(crate) enum CrateType<'a> {
...
+    }
+
+    /// Unique identifier, used to mark generated rustdoc files.
+    fn make_identifier(&self) -> String {
+        if matches!(self.base_features, BaseFeatures::Heuristic) && self.extra_features.is_empty() {
+            // If using the default config, return empty identifier
+            return String::new();
+        }
+        let mut s = std::collections::hash_map::DefaultHasher::new();
+        std::hash::Hash::hash(&self, &mut s);
```
obi1kenobi, line 217: I don't think this will work. The default hasher is randomized, and even if it weren't, it's intentionally unstable and its outputs may change between Rust versions. It isn't suitable for use as part of the cache key.

Separately, consider adding a test case that ensures that a specific feature configuration produces a specific hash value. Such a test case would have caught the problem I'm pointing out here.

test_crates/features_no_default/new/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+publish = false
+name = "features_no_default"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+
+[features]
+A = []
+B = []
```
obi1kenobi, line 11: Please make your editor add newlines at the end of all files.

test_outputs/function_missing.output.ron
```diff
@@ -1,4 +1,16 @@
 {
+    "./test_crates/features_simple/": [
+        {
+            "name": String("feature_dependant_function"),
```
obi1kenobi, line 4: typo: `dependant` -> `dependent` -- can you rename the directory and update all of these accordingly?

tests/feature_config.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    fn add_arg(&mut self, arg: &str) -> &mut Self {
+        self.args.push(String::from(arg));
+        self
+    }
+
+    fn run(&mut self) -> Assert {
+        self.cmd.args(&self.args).assert()
+    }
+}
```
obi1kenobi, line 29: Love this test tooling! 💯

obi1kenobi (commented):
I love the thought and care you've put into testing this thoroughly, it's awesome to see! It gives me great confidence in the code even though we're going to make a significant change to the `cargo-semver-checks` behavior 💯 

I have one major concern with respect to how we're doing hashing for caching, and otherwise just maintainability / internal implementation design questions and suggestions. I have confidence that the feature works overall, but would like to also make it as maintainable and easy to understand as possible, for the sake of every subsequent piece of functionality we're going to add near this code.

obi1kenobi:
Ignore the nightly failure for now, there's a new rustdoc format version (v25) and I'm waiting on `rustdoc-types` to release a corresponding new version before I can release the matching new adapter version for it.

src/main.rs
```diff
@@ -192,6 +208,44 @@ struct CheckRelease {
...
+    #[arg(long, value_name = "NAME", help_heading = "Features")]
+    features: Vec<String>,
+
+    /// Use the named features only in the baseline version.
+    #[arg(long, value_name = "NAME", help_heading = "Features")]
+    baseline_features: Vec<String>,
+
+    /// Use the named features only in the current version.
+    #[arg(long, value_name = "NAME", help_heading = "Features")]
+    current_features: Vec<String>,
```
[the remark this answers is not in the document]
staniewzki: Ah, I wanted to make it clear the the named features are used in specific version only.
What do you think about: `Use the named features in the baseline version only`?

src/main.rs
staniewzki: I agree, but don't see a method that makes this more readable.

src/rustdoc_gen.rs
staniewzki: No, I meant "foundation". I have trouble coming up with a better name. Maybe `StartingFeatures`? That seems awkward as well.

src/rustdoc_gen.rs
staniewzki: Yup, I am renaming the field to `is_baseline`

src/rustdoc_gen.rs
staniewzki: Two separate methods seems like a cool idea.

The function name is a victim of a refactor, when I named it, it didn't have arguments.

test_outputs/function_missing.output.ron
staniewzki: You meant the function, right?

staniewzki:
I really like the changes I made thanks to the review!

test_outputs/function_missing.output.ron
obi1kenobi: I was worried it might also be a crate name, but if it's just one function then even better :)

src/main.rs
```diff
@@ -192,6 +208,44 @@ struct CheckRelease {
...
+
+    /// Use all the features.
```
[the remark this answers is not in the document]
obi1kenobi: This is the text that users see when they run `--help`, right?

Probably both explain it briefly and describe how it's different than not providing the flag.

src/rustdoc_gen.rs
obi1kenobi: How about `FeatureGroup`?

src/main.rs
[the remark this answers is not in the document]
obi1kenobi: This is good!

.github/workflows/ci.yml
```diff
@@ -353,6 +357,96 @@ jobs:
...
+          workspaces: |
+            subject/target/semver-checks/local-serde-1_0_162
+
+      # Run even if there was a cache hit, to make sure the check passes.
+      - name: Run semver-checks
+        run: |
+          cd semver
+          ../bins/cargo-semver-checks semver-checks check-release --manifest-path="../subject/serde/Cargo.toml" --verbose
+
+      # Run cargo-semver-checks, enabling feature `unstable` this time.
```
obi1kenobi, line 412:
```suggestion
      # Run cargo-semver-checks, enabling feature `unstable` this time.
      # We expect this to fail.
      # This run must not use the `serde` rustdoc without this feature as a baseline.
      # If that happens, this check will unexpectedly pass -- and the test will fail.
```

A bit more explanation like this would be good. Right now it's obvious to both of us what this does and why, but in 6 months it won't be so obvious anymore.

Also, I assume you've already checked that checking the `unstable` feature in current vs not checking `unstable` in baseline passes? Since that's what would happen if we accidentally incorrectly reused the cache, and we want to know that won't _also_ fail for some unrelated reason and thereby prevent this test from catching the problem.

src/main.rs
```diff
@@ -192,6 +208,46 @@ struct CheckRelease {
...
+        help_heading = "Features",
+        conflicts_with = "only_explicit_features"
+    )]
+    default_features: bool,
+
+    /// Use no features except the explicitly mentioned ones.
+    #[arg(long, help_heading = "Features")]
+    only_explicit_features: bool,
+
+    /// Use the named features.
```
obi1kenobi, line 223: Perhaps mention that these features are added to the selected feature group (heuristically-selected, default features, or none)?

src/main.rs
```diff
@@ -192,6 +208,46 @@ struct CheckRelease {
...
+    #[arg(long, value_name = "NAME", help_heading = "Features")]
+    baseline_features: Vec<String>,
+
+    /// Use the named features in the current version only.
+    #[arg(long, value_name = "NAME", help_heading = "Features")]
+    current_features: Vec<String>,
+
+    /// Use all the features.
+    /// By default, features with names:
+    /// `unstable`, `nightly`, `bench`, `no_std` or starting with `__` are not used.
```
obi1kenobi, line 237: To me this conveys the exact opposite of the intended meaning: it seems to say "when you use `--all-features`, by default we omit the following list of features anyway: `unstable`, `nightly`, ..."

src/rustdoc_gen.rs
```diff
@@ -110,6 +111,85 @@ impl<'a> CrateSource<'a> {
         all_crate_features.extend(self.regular_features());
         all_crate_features.into_iter().collect()
     }
+
+    /// Sometimes creates include features that are not meant for public use
```
obi1kenobi, line 115:
```suggestion
    /// Sometimes crates include features that are not meant for public use
```

src/rustdoc_gen.rs
```diff
@@ -110,6 +111,85 @@ impl<'a> CrateSource<'a> {
...
+    /// or otherwise don't adhere to semver:
+    ///  - private features, like `bench` used for internal benchmarking only
+    ///  - nightly-only features, like `nightly`
+    ///  - unstable features containing experimental code, like `unstable`
+    ///
+    /// To ensure the best possible out=of-the-box user experience,
```
obi1kenobi, line 121:
```suggestion
    /// To ensure the best possible out-of-the-box user experience,
```

src/rustdoc_gen.rs
```diff
@@ -110,6 +111,85 @@ impl<'a> CrateSource<'a> {
...
+    /// this function attempts to heuristically exclude feature names like above.
+    ///
+    /// The heuristics is based on the name since cargo does not currently include
```
obi1kenobi, line 124:
```suggestion
    /// The heuristics are based on the name since cargo does not currently include
```

src/rustdoc_gen.rs
```diff
@@ -110,6 +111,85 @@ impl<'a> CrateSource<'a> {
...
+    /// a mechanism for marking features as private/hidden/unstable. When such
+    /// mechanisms are available in cargo, we'll update this functionality to make
+    /// use of them. Relevant cargo issues:
+    /// - unstable/nightly-only features: <https://github.com/rust-lang/cargo/issues/10881>
+    /// - private/hidden features:        <https://github.com/rust-lang/cargo/issues/10882>
+    ///
+    /// This function tries to filter mentioned features out.
+    /// Feature names not included by this function:
```
obi1kenobi, line 132: This can also very easily be misinterpreted as the exact opposite of what it's trying to say, as if it's saying "this function tries to filter such features out, but explicitly does not do so for the following feature names: unstable, nightly, bench, ..."

src/rustdoc_gen.rs
```diff
@@ -110,6 +111,85 @@ impl<'a> CrateSource<'a> {
...
+    pub(crate) fn feature_list_from_config(
+        &self,
+        global_config: &mut GlobalConfig,
+        feature_config: &FeatureConfig,
+    ) -> Vec<String> {
+        let all_features: std::collections::HashSet<String> =
+            self.all_features().into_iter().collect();
+
+        let result = [
+            match feature_config.base_features {
```
obi1kenobi, line 168: I think `feature_group` is a good fit instead of `base_features`.

src/rustdoc_gen.rs
```diff
@@ -123,11 +203,82 @@ pub(crate) enum CrateType<'a> {
...
+
+        // Calculate a rolling hash for the identifier, as it could end up too long.
+        let mut hash = Wrapping(0_u64);
+        let base = Wrapping(263_u64); // a hand-picked prime number
+        for b in unique_identifier.as_bytes() {
+            hash = hash * base + Wrapping(*b as u64);
+        }
+
+        format!("{:x}", hash)
+    }
```
obi1kenobi, line 274: This sort of thing is where a design doc would have been useful again. It's okay to stop and discuss approaches instead of always trying to produce new code. I promise it's going to be faster and less frustrating to discuss possibly-tricky design decisions up front, instead of implementing them first and discussing them second. It also shows your maturity as an engineer when you flag possible tricky situations, instead of relying on me as a code reviewer to flag them for you.

Why are we building our own hash implementation inside `rustdoc_gen`? What is the benefit, what is the risk? What alternatives did you consider? Is this a good tradeoff?

This is a blocker.

obi1kenobi (commented):
The code looks in pretty good shape, and I'm excited to merge and release this soon!

This PR can also be a good learning opportunity: 90% if not 100% of the suggestions I made here were things you could have caught too by carefully re-reading your code in the PR review interface. They were things like typos, user-facing text that could be misinterpreted, etc. I'm not in any way special at being able to catch them — but every time I catch them, we have to go through another fix-review loop which due to time zones can take 24h or more. Then you get frustrated that we haven't merged this after a month of work, and I'm frustrated that I'm reading 800 lines of code yet another time. It ends up being a lose-lose for both of us.

One of the most important skills you'll need in industry is the ability to structure your work so that it's easy to merge and meets the quality bar on the first review so that reviewers can just hit the "merge" button after one read. In retrospect, you can probably see that this PR was too large of a slice of work to do at once, and could have been split into "internal engine improvements" (3-4 PRs) + plugging in the CLI and adding end-to-end tests (1 more PR) instead. Even though this is more PRs, you could develop them in parallel and because they are smaller, they'll merge faster. Time-to-merge is exponential in PR size!

If you'd like to chat in more depth about how to structure PRs for more efficiency for both you and the code reviewer, I'd be happy to talk. Ping me on Zulip and we can set up a time.

.github/workflows/ci.yml
staniewzki: Yes, I checked that!

src/main.rs
staniewzki: Okay, I see

src/main.rs
obi1kenobi: Users are "in a hurry" and in practice won't cross-reference multiple places to determine how things work. Instead, they'll read maybe one place and make a guess about how it works. In this case they'll probably guess wrong — that's what we're trying to avoid here.

Just by looking around on our issue tracker, you can see a number of issues opened about things that already at the time were explained in `--help` text, in the repo FAQ, etc. The users that opened them are perfectly capable of reading, but just took the "path of least resistance" and made a guess based on looking at one thing. We can't change that, so we have to meet users where they are and give them all the contextually-relevant information in the place where they expect to find it i.e. the one place they'll read before making their guess about how things work.

staniewzki: Okay. So, something like: `Use the named feature. It is added to the selected feature group (heuristically-selected, default features, or none)`?

My issue with this is that 'feature group' is not defined anywhere, along with the `heuristically-selected` group name.

staniewzki: Also: how do I go about describing `--baseline-features` and `--current-features`? Making this description lengthy and repeating it three times seems awkward.

src/main.rs
```diff
@@ -232,9 +232,9 @@ struct CheckRelease {
     #[arg(long, value_name = "NAME", help_heading = "Features")]
     current_features: Vec<String>,
 
-    /// Use all the features.
-    /// By default, features with names:
-    /// `unstable`, `nightly`, `bench`, `no_std` or starting with `__` are not used.
+    /// Use all the features, including features named
+    /// `unstable`, `nightly`, `bench`, `no_std` or starting with `__`,
+    /// that are disabled by default.
```
obi1kenobi, line 237:
```suggestion
    /// that are otherwise disabled by default.
```
This is good, just a small tweak.

src/rustdoc_gen.rs
```diff
@@ -248,29 +249,25 @@ impl FeatureConfig {
...
-
-        // Calculate a rolling hash for the identifier, as it could end up too long.
-        let mut hash = Wrapping(0_u64);
-        let base = Wrapping(263_u64); // a hand-picked prime number
-        for b in unique_identifier.as_bytes() {
-            hash = hash * base + Wrapping(*b as u64);
-        }
-
-        format!("{:x}", hash)
+        // Sort the features, because the order is not revelant.
```
obi1kenobi, line 252:
```suggestion
        // Sort the features to ensure a deterministic order.
```
"Not relevant" is at risk of communicating the exact opposite of the intended meaning. The order isn't relevant for rustdoc, but is really relevant for our hashing or else we might needlessly miss the cache a lot.

src/rustdoc_gen.rs
```diff
@@ -248,29 +249,25 @@ impl FeatureConfig {
...
+        let mut to_serialize = self.clone();
+        to_serialize.extra_features.sort();
+
+        // Serialize the whole struct. `The is_baseline` field is also serialized,
```
obi1kenobi, line 256:
```suggestion
        // Serialize the whole struct. The `is_baseline` field is also serialized,
```

src/main.rs
```diff
@@ -192,6 +208,46 @@ struct CheckRelease {
     )]
     release_type: Option<ReleaseType>,
 
+    /// Use only the default and explicitly added features.
```
obi1kenobi, line 211:
```suggestion
    /// Use only the crate-defined default features, as well as any features 
    /// added explicitly via other flags.
    ///
    /// Using this flag disables the heuristic that enables all features
    /// except `unstable`, `nightly`, `bench`, `no_std`, and ones starting with `__`.
```

src/main.rs
```diff
@@ -192,6 +208,46 @@ struct CheckRelease {
...
+    #[arg(
+        long,
+        help_heading = "Features",
+        conflicts_with = "only_explicit_features"
+    )]
+    default_features: bool,
+
+    /// Use no features except the explicitly mentioned ones.
```
obi1kenobi, line 219:
```suggestion
    /// Use no features except ones explicitly added by other flags.
    ///
    /// Using this flag disables the heuristic that enables all features
    /// except `unstable`, `nightly`, `bench`, `no_std`, and ones starting with `__`.
```
Repetition in docs isn't bad, so long as it's brief, informative, and something the user needs to know but wouldn't have found out otherwise.

src/main.rs
obi1kenobi: I think the actual answer was to tweak the other help text instead, so I've added some suggestions there.

Here's what I'd do here:
```suggestion
    /// Add a feature to the set of features being checked.
    /// The feature will be used in both the baseline and the current version of the crate.
```
and then we can tweak the second sentence for `--baseline-features` and `--current-features` as needed.

obi1kenobi (commented):
I added some suggestions for the CLI. Everything else looks good, let's get this one home and release it!

obi1kenobi:
I merged `main` to make nightly start passing again.

obi1kenobi:
Awesome job 🎉

In the next PR, could you add a few examples of the new functionality to the README? That way they'll show up on crates.io in the next version and I'll also use them for the release notes and for the release thread on Twitter / Mastodon.

PR #427, opened 2023-04-12
feat: Add lint for missing trait methods

### LINKED ISSUE #368: New lint: trait item removed
Going from
```rs
pub trait Foo {
  fn foo();
}
```
to
```rs
pub trait Foo {
}
```
is a breaking change. There should be a lint for that.

zurawiki:
- chore: Bump `trustfall_rustdoc` version to `0.10.0`
- test: Add trait_method_removed test crate
- chore: cargo update
- feat: Add lint for missing trait methods

Closes #368


src/lints/trait_method_missing.ron
```diff
@@ -0,0 +1,58 @@
...
+
+                            span_: span @optional {
+                                filename @output
+                                begin_line @output
+                            }
+                        }
+                    }
+                }
+            }
+            current @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
```
obi1kenobi, line 32: Recall that we're aiming for only a single lint to be triggered for any given semver violation.

With this `@fold @transform(op: "count") @filter(op: "=", value: ["$zero"])` here, I believe it's possible to trigger this lint in a case that _doesn't involve_ removing or renaming a trait method.

Can you think of any such example situations? They'd be great to include in the test suite even after updating the query here.

src/lints/trait_method_missing.ron
```diff
@@ -0,0 +1,58 @@
...
+                item {
+                    ... on Trait {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                        }
+
+                        impl @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
```
obi1kenobi, line 41: I think this might be a bit of copy-pasta from another query, since `Trait` has a `method` edge and doesn't need to go through `impl`.

src/lints/trait_method_missing.ron
```diff
@@ -0,0 +1,58 @@
...
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "zero": 0,
+    },
+    error_message: "A publicly-visible trait method is no longer visible, or cannot be called by its prior path. A member, method, or item of the trait may have been renamed or removed entirely.",
+    per_result_error_template: Some("trait {{join \"::\" path}}, previously in file {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 57: Let's name the problematic method here as well as its trait. Right now we only name the trait, so if more than one method gets removed from the same trait, the user would see a duplicate error message and be quite confused.

obi1kenobi (commented):
This is an excellent start!

The query is almost there, and as I think I mentioned, most of the work is going to be in coming up with good wording for the user-facing error messages and in making good test cases :)

src/lints/trait_method_missing.ron
zurawiki: ack

src/lints/trait_method_missing.ron
```diff
@@ -0,0 +1,58 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "zero": 0,
+    },
+    error_message: "A publicly-visible trait method is no longer visible, or cannot be called by its prior path. A member, method, or item of the trait may have been renamed or removed entirely.",
```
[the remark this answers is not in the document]
zurawiki: ack, still working on it

src/lints/trait_method_missing.ron
zurawiki: ack

zurawiki:
I'm still trying to setup the crate so it can build but it seems that there are conflicting types coming from the fact that multiple `trustfall_core` crate versions are imported. This seems to be the source of the failure with the linter. I can reproduce this locally, and this issue prevents me from building `cargo-semver-checks`.


```
error[E0308]: mismatched types
   --> src/query.rs:299:13
    |
299 | /             results_iter
300 | |                 .map(|res| res.into_iter().map(|(k, v)| (k.to_string(), v)).collect())
301 | |                 .collect::<Vec<BTreeMap<_, _>>>(),
    | |_________________________________________________^ expected enum `trustfall::FieldValue`, found enum `trustfall_core::ir::value::FieldValue`
    |
    = note: enum `trustfall_core::ir::value::FieldValue` and enum `trustfall::FieldValue` have similar names, but are actually distinct types
note: enum `trustfall_core::ir::value::FieldValue` is defined in crate `trustfall_core`
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/trustfall_core-0.4.0-beta.1/src/ir/value.rs:12:1
    |
12  | pub enum FieldValue {
    | ^^^^^^^^^^^^^^^^^^^
note: enum `trustfall::FieldValue` is defined in crate `trustfall_core`
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba1[500](https://github.com/obi1kenobi/cargo-semver-checks/actions/runs/4674097773/jobs/8277971388?pr=427#step:6:501)1f/trustfall_core-0.3.3/src/ir/value.rs:10:1
    |
10  | pub enum FieldValue {
    | ^^^^^^^^^^^^^^^^^^^
    = note: perhaps two different versions of crate `trustfall_core` are being used?
```

What's the best way to update all dependencies on trustfall code?

obi1kenobi:
> I'm still trying to setup the crate so it can build but it seems that there are conflicting types coming from the fact that multiple `trustfall_core` crate versions are imported. This seems to be the source of the failure with the linter. I can reproduce this locally, and this issue prevents me from building `cargo-semver-checks`.

It just needed a matching version update for the `trustfall` dependency together with `trustfall_rustdoc`. I added it in #428 which should merge shortly, and you'll be able to just merge `main` to pick it up in a bit.

zurawiki:
Thanks for following up. I'd like to see this through but I unfortunately won't have bandwidth this coming week. I'm happy to let someone else takeover the task if there is interest.

obi1kenobi:
There's no rush, so if you think you'll be able to come back to it in the coming weeks, you're welcome to do that. But also no pressure, it's not a big deal to close this and let someone else take a crack at it.

src/lints/trait_method_missing.ron
```diff
@@ -0,0 +1,59 @@
...
+                        name @output
+
+                        importable_path {
+                            path @output @tag
+                        }
+
+                        span_: span @optional {
+                                filename @output
+                                begin_line @output
+                            }
```
obi1kenobi, line 23:
```suggestion
                            filename @output
                            begin_line @output
                        }
```

obi1kenobi: Just wanted to help with a quick fix. You're headed in the right direction in general.

zurawiki:
Updated the PR with the feedback. It is now ready for review. Thanks @obi1kenobi for all your guidance here.

obi1kenobi:
As soon as #455 merges, this should start passing tests. Thanks for putting it together!

src/lints/trait_method_missing.ron
```diff
@@ -0,0 +1,54 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "zero": 0,
+    },
+    error_message: "A publicly-visible trait method is no longer callable. A method of the trait may have been renamed or removed entirely.",
```
obi1kenobi, line 52:
```suggestion
    error_message: "A trait method is no longer callable, and may have been renamed or removed entirely.",
```
Just a bit more succinct. Unlike regular methods, trait methods are always public if the trait itself is public, so we probably don't have to emphasize the public-ness.

obi1kenobi (approved):

obi1kenobi:
Thanks for putting this together! The nightly failure is on me, we just need to pull in the adapter for the new format and I'll take care of it a bit later.

PR #434, opened 2023-04-20
Reproduce the `perf-event-open-sys2` crash in CI.

obi1kenobi:
Expecting this to fail and therefore confirm that it's a valid repro. Then I will merge it into #433 to verify the fix there.


obi1kenobi:
It works!

PR #449, opened 2023-05-05
Minimum rustc version error

dianegolfouse:
Add an error when the user is running cargo-semver-checks with a rustc version `< 1.65`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,7 @@ jobs:
       - run-on-tokio-explicit
       - run-on-tokio-implicit
       - run-on-perf-event-open-sys2
+      - fails-on-1-64-rustc-rust-fnv
     if: ${{ success() || failure() }}  # Run this job even if a dependency has failed.
     steps:
       - name: Job outcomes
@@ -60,6 +61,7 @@ jobs:
           echo "run-on-tokio-explicit: ${{ needs.run-on-tokio-explicit.result }}"
           echo "run-on-tokio-implicit: ${{ needs.run-on-tokio-implicit.result }}"
           echo "run-on-perf-event-open-sys2: ${{ needs.run-on-perf-event-open-sys2.result }}"
+          echo "fails-on-1-64-rustc-rust-fnv: ${{ needs.fails-on-1-64-rustc-rust-fnv.result }}"
 
       # Fail this required job if any of its dependent jobs have failed.
       #
@@ -99,6 +101,8 @@ jobs:
       - if: ${{ needs.run-on-tokio-implicit.result != 'success' }}
         run: exit 1
       - if: ${{ needs.run-on-perf-event-open-sys2.result != 'success' }}
+        run: exit 1
+      - if: ${{ needs.fails-on-1-64-rustc-rust-fnv.result != 'success' }}
         run: exit 1
 
   lint:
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -363,12 +363,24 @@ impl Check {
...
                         let version = &selected.version;
 
+                        let rust_version = rustc_version::version()
+                            .expect("the rust version could not be determined !");
+                        if rust_version < rustc_version_1_65 {
+                            anyhow::bail!(
+                                "rustc version is not high enough: >=1.65.0 needed, got {}\n\nHELP: to run with the latest rustc version, use `rustup update stable && cargo +stable semver-checks ...`",
+                                rust_version
+                            );
+                        }
```
obi1kenobi, line 382: I think this logic should be outside of this `match` so that it applies to all cases, not just this most-common one.

And then we'll probably want to explicitly skip the check if both current and baseline rustdoc files are given explicitly instead of being generated through the tool, since then the installed Rust version doesn't matter.

test_crates/minimum_rustc/fails/rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "1.64"
```
obi1kenobi, line 2: Unfortunately I think the way our tests are set up will explicitly ignore this.

It might be less work to write this test as a GitHub Action job. That way we don't have to change how the test setup works, and people don't have to keep Rust 1.64 installed just to run tests locally.

arnaudgolfouse: So, I managed to have the test actually use the `1.64` channel automatically. However I agree that forcing people to have Rust 1.64 installed is pretty annoying.
My original idea was that it would catch bumps of the minimal version, so that the error detection + message might be updated. But looking at it now, I guess such a bump would fail CI anyway, and it would make it clear that the check `rust_version < rustc_version_1_65` must be updated... I think ?

So, should I remove the tests entirely ?

src/lib.rs
```diff
@@ -363,12 +363,24 @@ impl Check {
...
+                let rustc_version_1_65 = rustc_version::Version::new(1, 65, 0);
+
                 selected
                     .iter()
                     .map(|selected| {
                         let crate_name = &selected.name;
                         let version = &selected.version;
 
+                        let rust_version = rustc_version::version()
+                            .expect("the rust version could not be determined !");
```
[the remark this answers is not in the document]
arnaudgolfouse: I see that the default log level ignores warning, is that an issue ?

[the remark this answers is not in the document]
arnaudgolfouse: Ah nevermind, I suppose I should use `config.shell_warn`

src/lib.rs
arnaudgolfouse: Done!

src/lib.rs
```diff
@@ -317,6 +317,26 @@ impl Check {
...
+                    let rustc_version_needed = rustc_version::Version::new(1, 65, 0);
+                    if rustc_version < rustc_version_needed {
+                        anyhow::bail!("rustc version is not high enough: >={} needed, got {}\n\nHELP: to run with the latest rustc version, use `rustup update stable && cargo +stable semver-checks ...`", rustc_version_needed, rustc_version);
+                    }
+                }
+                Err(error) => {
+                    config.shell_warn(format_args!(
+                        "while trying to query the current rustc version: {}",
+                        error
+                    ))?;
```
obi1kenobi, line 335: This error message text is "implementation-oriented." The user that would see it wouldn't know we were querying the rustc version (or why) so they'll just be confused and won't know what to do. Let's try to make this a user-oriented, actionable error message:
```suggestion
                    config.shell_warn(format_args!(
                        "failed to determine the current rustc version: {}\n\nHELP: to avoid errors please ensure Rust 1.65+ is used",
                        error
                    ))?;
```
or something to that effect. Bonus points if you could pull the `rustc_version_needed` into `config` itself and then somehow use its value to generate the "Rust 1.65+" portion of the error message string.

src/lib.rs
```diff
@@ -317,6 +317,26 @@ impl Check {
...
 
+        // If both the current and baseline rustdoc are given explicitely as a file path, we don't need to use the installed rustc, and this check can be skipped.
+        if !(matches!(self.current.source, RustdocSource::Rustdoc(_))
+            && matches!(self.baseline.source, RustdocSource::Rustdoc(_)))
+        {
+            match rustc_version::version() {
+                Ok(rustc_version) => {
+                    let rustc_version_needed = rustc_version::Version::new(1, 65, 0);
+                    if rustc_version < rustc_version_needed {
+                        anyhow::bail!("rustc version is not high enough: >={} needed, got {}\n\nHELP: to run with the latest rustc version, use `rustup update stable && cargo +stable semver-checks ...`", rustc_version_needed, rustc_version);
```
obi1kenobi, line 328:
```suggestion
                        anyhow::bail!("rustc version is not high enough: >={} needed, got {}\n\nHELP: to run with the latest rustc version, use `rustup update stable && cargo +stable semver-checks <arguments-here>`", rustc_version_needed, rustc_version);
```
Just out of concern someone might copy-paste this command without noticing that the `...` isn't an actual argument to `cargo-semver-checks`.

test_crates/minimum_rustc/fails/rust-toolchain.toml
obi1kenobi: I would remove the tests here but then set up an equivalent test in GitHub Actions, where it explicitly uses Rust 1.64 to semver-check a crate, captures the output, and asserts the exit code was non-zero and contained the warning about using an old Rust.

There are some tests like that in GitHub Actions already, and you should be able to do what they do as well.

obi1kenobi (commented):
Functionality-wise this looks good. Just a couple of wording suggestions in the error messages and the test tweaks.

obi1kenobi (commented):
Functionality-wise this looks good. Just a couple of wording suggestions in the error messages and the test tweaks.

src/lib.rs
arnaudgolfouse: Does pulling the `rustc_version_needed` into `config` bring other benefits? Right now the message is generated by the local `let rustc_version_needed = rustc_version::Version::new(1, 65, 0)`, and it works well.

obi1kenobi: This local `let` is unfortunately not authoritative on which version is actually supported — it's totally possible that we bump the underlying `trustfall_rustdoc` dependency (which is authoritative on the supported rustdoc versions) while forgetting to change this value here.

It's somewhat less likely that we forget it if it's either a global static, or in the config, than a random local variable in the middle of the file. So it's just about improving the probability of us remembering to change it, nothing else :)

src/lib.rs
```diff
@@ -316,6 +316,26 @@ impl Check {
             .deps(false)
             .silence(!config.is_verbose());
 
+        // If both the current and baseline rustdoc are given explicitely as a file path, we don't need to use the installed rustc, and this check can be skipped.
```
obi1kenobi, line 319:
```suggestion
        // If both the current and baseline rustdoc are given explicitly as a file path, we don't need to use the installed rustc, and this check can be skipped.
```
Typo fix + let's try to keep comments and string literals to below 100 chars both here and below

src/lib.rs
```diff
@@ -316,6 +316,26 @@ impl Check {
...
+        if !(matches!(self.current.source, RustdocSource::Rustdoc(_))
+            && matches!(self.baseline.source, RustdocSource::Rustdoc(_)))
+        {
+            let rustc_version_needed = rustc_version::Version::new(1, 65, 0);
```
obi1kenobi, line 323: If we can lift this to a global static (ideally) or into the global config (next best), that'd be great.

obi1kenobi (commented):
The new test looks great. Just a couple of minor nits left.

I'll fix the nightly problem tomorrow or on Monday, don't worry about it for now. It isn't required to pass to merge PRs.

obi1kenobi (commented):
The new test looks great. Just a couple of minor nits left.

I'll fix the nightly problem tomorrow or on Monday, don't worry about it for now. It isn't required to pass to merge PRs.

src/lib.rs
arnaudgolfouse: Got it!

obi1kenobi (approved):
Looks great, thanks for all the help!

obi1kenobi:
If you might be interested in continuing to contribute to the project, we could certainly use the help! #451 in particular is an unfortunate (and kind of embarrassing) issue that shouldn't be too hard to fix. The full list of contributor-friendly issues is here: https://github.com/obi1kenobi/cargo-semver-checks/issues?q=is%3Aissue+is%3Aopen+label%3AE-help-wanted

.github/workflows/ci.yml
```diff
@@ -60,6 +60,7 @@ jobs:
           echo "run-on-tokio-explicit: ${{ needs.run-on-tokio-explicit.result }}"
           echo "run-on-tokio-implicit: ${{ needs.run-on-tokio-implicit.result }}"
           echo "run-on-perf-event-open-sys2: ${{ needs.run-on-perf-event-open-sys2.result }}"
+          echo "fails-on-1-64-rustc-rust-fnv: ${{ needs.fails-on-1-64-rustc-rust-fnv.result }}"
```
obi1kenobi, line 63: You'll need to add this to the `needs:` list above, that's why this job is failing right now.

After that this is ready to merge.

obi1kenobi (approved):

Issue #183, opened 2022-11-15
Improved error message when no baseline version found on crates.io


obi1kenobi:
Right now, if attempting to semver-check a crate that doesn't have a version published on `crates.io`, the error message is terse and not helpful:
```
    Updating index
     Parsing foo_bar v0.1.0 (current)
Error: foo_bar not found in registry
```

Possible improvements:
- State that the registry is crates.io specifically. Consider linking to the tracking issue for private registries (#166) in case that's where the crate is published.
- Advise the user that they may still semver-check this crate by using one of the other ways to get a baseline, for example via a git ref.

Perhaps related to #146 as well.

era:
Hi @obi1kenobi, I would like to give it a try. If I understood correctly the error you are mentioning comes from https://github.com/obi1kenobi/cargo-semver-checks/blob/main/src/rustdoc_gen.rs#L589, given `need_retry` I imagine the error is `err.class() == git2::ErrorClass::Index && err.code() == git2::ErrorCode::NotFound`.

And for what I see on the project, the errors bubble up to the main function and force the program to exit with the error message printed. If that is the case, would something like this be ok or you have something else in mind?

```rust
while need_retry(index.update()).map_err(human_readable_message)? {
            config.shell_status("Blocking", "waiting for lock on registry index")?;
            std::thread::sleep(REGISTRY_BACKOFF);
        }
```

obi1kenobi:
I believe that's all correct and the approach seems sound. Go for it! Thanks for giving it a try 🙌

era:
In the end, it was more simple because there is already a check if the crate is at crates.io, so I just modified the error message. If this is not enough I can go back to the previous solution we talked about.

Issue #431, opened 2023-04-18
Missing `path` in `enum_variant_added` lint's `per_result_error_message`.


tonowak:
There is only 'enum_name' and no 'path' in the message, which makes parsing the output harder.

Issue #430, opened 2023-04-17
Confusing `method_parameter_count_changed` error message (`per_result_error_template`)


tonowak:
The message https://github.com/obi1kenobi/cargo-semver-checks/blob/a53106c4f384ed3409b6da033743f038ce0fb43f/src/lints/method_parameter_count_changed.ron#L104 points to the line number of the `baseline`, but the context makes it sound like it is in `current`.

New message suggestion from @obi1kenobi: `method X used to take 2 parameters in <span> but now takes 1 parameter in <spans>`

Issue #424, opened 2023-03-30
cargo semver-checks will exit 0 even if it skips all crates


rcoh:
1. Create a cargo workspace where all crates are set to `publish = false`
2. `cargo semver-checks check-release  -v`
Without `-v` it seems to imply that all is well—it will silently skip all crates. Checking 0 crates is unlikely to be what the user expected, perhaps it should exist non-zero?

obi1kenobi:
I didn't add a test case specifically for workspaces full of `publish = false` crates, so I'm not going to close this right now, but #477 does add an explicit error if the selections contain no crates with library targets, so it's _possible_ this case would be handled too.

PR #456, opened 2023-05-28
Improved error message when no baseline version found on crates.io

era:
Fixes https://github.com/obi1kenobi/cargo-semver-checks/issues/183. 

The new error message now is:

```
Error: tinylang not found in registry. The registry is crates.io. Crates published on a custom registry cannot be checked using default settings, see https://github.com/obi1kenobi/cargo-semver-checks/issues/166
``` 

Where `tinylang` is the crate I tested the program. 

Let me know how you want to change the error message :)

### DIFF
--- a/src/rustdoc_gen.rs
+++ b/src/rustdoc_gen.rs
@@ -586,7 +586,7 @@ impl RustdocFromRegistry {
         let mut index = crates_index::Index::new_cargo_default()?;
 
         config.shell_status("Updating", "index")?;
-        while need_retry(index.update())? {
+        while need_retry(index.update()).map_err(human_readable_message)? {
             config.shell_status("Blocking", "waiting for lock on registry index")?;
             std::thread::sleep(REGISTRY_BACKOFF);
         }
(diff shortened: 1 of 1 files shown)


obi1kenobi:
Neat, thanks for the PR! The new message is certainly a lot better, and merging it would already be an improvement.

In case you're interested in doing one more round of polish, I wanted to suggest a couple of options for further improvements if you'd be open to it:
- The repetition of "registry" in the beginning is a bit verbose. Concise errors are best. Can we shorten it to `Error: tinylang not found in registry (crates.io)` or similar?
- Do you think linking to [the FAQ section on checking non-crates.io crates](https://github.com/obi1kenobi/cargo-semver-checks#does-the-crate-im-checking-have-to-be-published-on-cratesio) might be a better option?
- It's possible at this point in the code that the user didn't actually use default settings, and in that case the reference to default settings would be confusing. The underlying problem is that we're trying to look up a crate version in the registry, and failing since the registry doesn't know about the crate. Perhaps we can reuse some language from the FAQ entry and nudge the user toward the workaround of supplying a baseline themselves (and not just a baseline version number).

Don't worry about the nightly failure, that's a "me" problem and not anything you did. I'm on it.

Also, might you be interested in tackling some more issues in the project? If you'd like, I'd be happy to act as concierge and suggest issues if you let me know what sorts of things you find interesting.

era:
Hi @obi1kenobi , thank you for the quick review.

I tried to reduce the message in this new revision, I hope I was able to still pass the right message. Since the problem is not necessarily the usage of a custom registry, but also the possibility of the user not even publishing on one I tried to follow your advice and be a bit more generic towards the root cause.

I think the FAQ makes sense, the person facing the error will probably go to the GitHub issue and follow the link to the FAQ anyway, so maybe this way we can optimize it a bit.

Let me know if you want me to rephrase something on the error message.

> Also, might you be interested in tackling some more issues in the project? If you'd like, I'd be happy to act as concierge and suggest issues if you let me know what sorts of things you find interesting.

I would love to :), I just don't want to delay any important feature, so any task that is not high-priority and I can take my time to work on it would be perfect.

src/rustdoc_gen.rs
```diff
@@ -658,10 +658,14 @@ impl RustdocGenerator for RustdocFromRegistry {
...
         crate_data: CrateDataForRustdoc,
     ) -> anyhow::Result<PathBuf> {
-        let crate_ = self
-            .index
-            .crate_(crate_data.name)
-            .with_context(|| anyhow::format_err!("{} not found in registry", crate_data.name))?;
+        let crate_ = self.index.crate_(crate_data.name).with_context(|| {
+            anyhow::format_err!(
+                "{} not found in registry (crates.io). \
+       For workarounds check \
```
obi1kenobi, line 664:
```suggestion
        For workarounds check \
```
just an alignment fix

obi1kenobi (approved):

obi1kenobi:
Awesome, thanks for the help! Depending on your interests, there are at least two flavors of issues to choose from.

One set is issues that directly improve the CLI and ergonomics of using the tool. A few example issues:
- #35
- #424
- #414
- #325
- #298

Another direction is improving our checking capabilities, like adding more lints and improving existing ones. For example:
- #430
- #431
- #366, or any other subpart of #241
- #232

Sometimes, improving our checking capabilities requires making changes in [the `trustfall-rustdoc-adapter` crate](https://github.com/obi1kenobi/trustfall-rustdoc-adapter) as well. I'd be happy to help you get onboarded — it isn't nearly as hard as it looks — and you could make a seriously valuable contribution by writing up onboarding docs we could use for future contributors.

None of these are particularly time-sensitive, so don't worry about taking your time. We're all volunteers here so things get done when they get done :)

PR #458, opened 2023-05-30
Update README with feature flags examples

staniewzki:
I am not sure if adding this to FAQ is the best approach, especially as this section is pretty big relative to the others.

There's also a slight change in a comment that I didn't commit in the last PR on accident.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,35 +77,17 @@ registries other than crates.io should use one of the other approaches of genera
 
 By default, crates are built with all features excluding heuristically selected features that usually break semver. More precisely, features named `unstable`, `nightly`, `bench`, `no_std` or starting with `__` are disabled.
 
-The flags below can be used to select the checked features set:
-```
---default-features
-    Use only the crate-defined default features, as well as any features added explicitly via other flags.
+This behaviour can be overriden. Checked feature set can be changed to:
+- all the features, selected with `--all-features`,
+- only the crate's default features, selected with `--default-features`,
+- empty set, selected with `--only-explicit-features`.
 
-    Using this flag disables the heuristic that enables all features except `unstable`, `nightly`, `bench`, `no_std`, and ones starting with `__`.
-
---only-explicit-features
-    Use no features except ones explicitly added by other flags.
-
-    Using this flag disables the heuristic that enables all features except `unstable`, `nightly`, `bench`, `no_std`, and ones starting with `__`.
-
---features <NAME>
-    Add a feature to the set of features being checked. The feature will be used in both the baseline and the current version of the crate
-
---baseline-features <NAME>
-    Add a feature to the set of features being checked. The feature will be used in the baseline version of the crate only
-
---current-features <NAME>
-    Add a feature to the set of features being checked. The feature will be used in the current version of the crate only
-
---all-features
-    Use all the features, including features named `unstable`, `nightly`, `bench`, `no_std` or starting with `__`, that are otherwise disabled by default
-```
+Additionally, features can be enabled one-by-one, using flags `--features`, `--baseline-features` and `--current-features`.
 
 For example, consider crate [serde](https://github.com/serde-rs/serde), with the following features (per v1.0.163):
 - `std` - the only crate's default feature,
-- `alloc`, `derive`, `rc` - some optional features,
-- `unstable`- a feature that possibly breaks semver.
+- `alloc`, `derive`, `rc` - optional features,
+- `unstable` - a feature that possibly breaks semver.
 
 | used flags | selected feature set |
 |--|--|
--- a/README.md
+++ b/README.md
@@ -102,18 +102,18 @@ The flags below can be used to select the checked features set:
     Use all the features, including features named `unstable`, `nightly`, `bench`, `no_std` or starting with `__`, that are otherwise disabled by default
 ```
 
-Let's say we have a [crate](https://github.com/serde-rs/serde) with the following features:
+For example, consider crate [serde](https://github.com/serde-rs/serde), with the following features (per v1.0.163):
 - `std` - the only crate's default feature,
 - `alloc`, `derive`, `rc` - some optional features,
 - `unstable`- a feature that possibly breaks semver.
 
 | used flags | selected feature set |
 |--|--|
-| none | `std`, `non-default`  |
-| `--features unstable` | `std`, `non-default`, `unstable` |
-| `--all-features` | `std`, `non-default`, `unstable` |
+| none | `std`, `alloc`, `derive`, `rc`  |
+| `--features unstable` | `std`, `alloc`, `derive`, `rc`, `unstable` |
+| `--all-features` | `std`, `alloc`, `derive`, `rc`, `unstable` |
 | `--default-features` | `std` |
-| `--default-features --features non-default` | `std`, `non-default` |
+| `--default-features --features derive` | `std`, `derive` |
 | `--only-explicit-features --features unstable` | `unstable` |
 
 ### Does `cargo-semver-checks` have false positives?
--- a/README.md
+++ b/README.md
@@ -72,6 +72,49 @@ Custom registries are not currently supported
 ([#160](https://github.com/obi1kenobi/cargo-semver-check/issues/160)), so crates published on
 registries other than crates.io should use one of the other approaches of generating the baseline.
 
+### What features does `cargo-semver-checks` enable in the tested crates?
+
+By default, crates are built with all features excluding heuristically selected features that usually break semver. More precisely, features named `unstable`, `nightly`, `bench`, `no_std` or starting with `__` are disabled.
+
+The flags below can be used to select the checked features set:
+```
+--default-features
+    Use only the crate-defined default features, as well as any features added explicitly via other flags.
+
+    Using this flag disables the heuristic that enables all features except `unstable`, `nightly`, `bench`, `no_std`, and ones starting with `__`.
+
+--only-explicit-features
+    Use no features except ones explicitly added by other flags.
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -72,6 +72,49 @@ Custom registries are not currently supported
...
+
+--baseline-features <NAME>
+    Add a feature to the set of features being checked. The feature will be used in the baseline version of the crate only
+
+--current-features <NAME>
+    Add a feature to the set of features being checked. The feature will be used in the current version of the crate only
+
+--all-features
+    Use all the features, including features named `unstable`, `nightly`, `bench`, `no_std` or starting with `__`, that are otherwise disabled by default
+```
```
obi1kenobi, line 102: This code block doesn't look good when rendered:
https://github.com/staniewzki/cargo-semver-check/tree/cli-specify-features#what-features-does-cargo-semver-checks-enable-in-the-tested-crates

README.md
```diff
@@ -72,6 +72,49 @@ Custom registries are not currently supported
 ([#160](https://github.com/obi1kenobi/cargo-semver-check/issues/160)), so crates published on
 registries other than crates.io should use one of the other approaches of generating the baseline.
 
+### What features does `cargo-semver-checks` enable in the tested crates?
```
obi1kenobi, line 75: There's a thought exercise that I think is worth trying for this, if you're up for it.

Pretend you're a user of this tool, not a contributor.

When might you come looking for this section? What questions might you have about it?

Were you able to find the answers quickly and efficiently? If not, what changes would help with that?

Did reading the section leave you with more unanswered questions? If so, can we either rephrase to avoid introducing the confusion that raises those questions, or answer them immediately?

Ideally, the section should _succinctly_ answer the user's questions, avoid ambiguity, and leave the user with an impression of "this feature is well-designed and does what I want."

README.md
```diff
@@ -72,6 +72,49 @@ Custom registries are not currently supported
...
+
+By default, crates are built with all features excluding heuristically selected features that usually break semver. More precisely, features named `unstable`, `nightly`, `bench`, `no_std` or starting with `__` are disabled.
```
obi1kenobi, line 77: Try to avoid linker phrases like "more precisely." I know we're all taught in school that we should connect our sentences together — I'm still working on getting rid of "therefore" and "also" in my own writing. But these are all just fluff, and make the writing less concise without helping the reader get the point.

Better to just get straight to the point.

README.md
```diff
@@ -72,6 +72,49 @@ Custom registries are not currently supported
...
+- `non-default`,
+- `unstable`- a feature that possibly breaks semver.
+
+| used flags | selected feature set |
+|--|--|
+| none | `default`, `non-default`  |
+| `--features unstable` | `default`, `non-default`, `unstable` |
+| `--all-features` | `default`, `non-default`, `unstable` |
+| `--default-features` | `default` |
+|`--default-features --features non-default` | `default`, `non-default` |
```
obi1kenobi, line 115:
```suggestion
| `--default-features --features non-default` | `default`, `non-default` |
```

README.md
```diff
@@ -72,6 +72,49 @@ Custom registries are not currently supported
...
+--current-features <NAME>
+    Add a feature to the set of features being checked. The feature will be used in the current version of the crate only
+
+--all-features
+    Use all the features, including features named `unstable`, `nightly`, `bench`, `no_std` or starting with `__`, that are otherwise disabled by default
+```
+
+Let's say we have a crate with the following features:
+- `default` - the only crate's default feature,
+- `non-default`,
```
obi1kenobi, line 106: Let's rename this feature to something else. I know it's `non-default` but the "default" substring there really throws me off in the examples. It's also a possible point of confusion around whether `--default-features` means "features named `default`" or something else.

Consider using real feature names crates would use and users might already be used to. For example, `std` as a default feature, `serde` as an optional feature, etc. That way we don't risk confusing the reader between flags and feature names.

obi1kenobi (commented):
Thanks for putting this together! Writing user-facing docs is really hard and takes a lot of practice and iterations to get to the highly-polished state that is often seen in the Rust ecosystem. I've made some suggestions to move closer to that state.

README.md
staniewzki: I guess it's not exactly what you meant, but `serde` seems like an excellent crate to showcase the functionality.

README.md
staniewzki: Yes, it is little bit awkward. Some solutions I see:
1. Try to wrap the lines to make them shorter. This still results in a big code block.
2. Move it out of a codeblock, and briefly explain all the flags related to features.
3. Scrap it altogether, and explain the functionality in a shorter way.

README.md
staniewzki: I will try to come up with a short list of possible questions tomorrow, I am too tried already today 😵‍💫

staniewzki: Putting it here for documentation's sake.

Some scenarios (put in order of how probable they seem to me, from most to least):
- You are considering adding cargo-semver-checks to your project's CI and you are curious if it would work for your usecase. 
  - does cargo-semver-checks detect API changes for every feature configuration?
  - what features are enabled in the checks?
  - how much can the feature set be customized?
- You are trying to use adapt cargo-semver-checks in your project's CI,
  - what features exactly are enabled by default?
  - how to enable a specific set of features?
  - how does a flag work?
 - Running cargo-semver-checks results in a false-positive, because of the feature configuration. 
   - is this the expected behaviour?
   - how to fix this issue?
 - You add (or possibly remove) a feature into your project that already uses cargo-semver-checks
   -  how does cargo-semver-checks interacts with that?
   - how to make sure the new feature is checked?
   - how to run checks when removing a feature?

staniewzki:
I have notices that in https://github.com/obi1kenobi/cargo-semver-checks/pull/463 the help messages were not updated, so I am adding that here.

README.md
staniewzki: I implemented the 2. solution, and I think I like it.

README.md
staniewzki: I looked through this list, and I think that answers to almost all the questions can be found reading the README, possibly requiring to also look at the tool's help message. 

I am not sure about the last point:
> * You add (or possibly remove) a feature into your project that already uses cargo-semver-checks
>   * how does cargo-semver-checks interacts with that?
>   * how to make sure the new feature is checked?
>   * how to run checks when removing a feature?

README.md
obi1kenobi: How about something like:
```suggestion
By default, checking is done on all features except features named `unstable`, `nightly`, `bench`, `no_std`, or ones with prefix `_`, `unstable-`, or `unstable_`, as such names are commonly used for private or unstable features.
```
It's roughly the same length. I've also updated it for the heuristic tweaks suggested by Ed Page.

README.md
obi1kenobi: Great catch! This last point is worth making a new FAQ entry for — let's do a separate PR for that so we can merge this one faster.

If you'd prefer not to make that PR, could you instead just open an issue about it so we don't forget?

README.md
```diff
@@ -72,6 +73,31 @@ Custom registries are not currently supported
...
+
+This behaviour can be overriden. Checked feature set can be changed to:
+- *all* the features, selected with `--all-features`,
+- only the crate's default features, selected with `--default-features`,
+- empty set, selected with `--only-explicit-features`.
+
+Additionally, features can be enabled one-by-one, using flags `--features`, `--baseline-features` and `--current-features`.
+
+For example, consider crate [serde](https://github.com/serde-rs/serde), with the following features (per v1.0.163):
+- `std` - the only crate's default feature,
```
obi1kenobi, line 88:
```suggestion
- `std` - the crate's only default feature,
```
This way `only` refers to "default feature" and not to "crate".

README.md
```diff
@@ -72,6 +73,31 @@ Custom registries are not currently supported
...
+- `alloc`, `derive`, `rc` - optional features,
+- `unstable` - a feature that possibly breaks semver.
+
+| used flags | selected feature set |
+|--|--|
+| none | `std`, `alloc`, `derive`, `rc`  |
+| `--features unstable` | `std`, `alloc`, `derive`, `rc`, `unstable` |
+| `--all-features` | `std`, `alloc`, `derive`, `rc`, `unstable` |
+| `--default-features` | `std` |
+| `--default-features --features derive` | `std`, `derive` |
```
obi1kenobi, line 98:
```suggestion
| `--default-features --features derive` | `std`, `derive` |
| `--only-explicit-features` | none |
```
Seems worth documenting.

README.md
```diff
@@ -72,6 +73,31 @@ Custom registries are not currently supported
...
+| `--only-explicit-features --features unstable` | `unstable` |
```
obi1kenobi, line 99: This is much clearer already.

Have you thought about possibly adding an "explanation" column to the right?

For example:
- In the `none` row it could say: "Feature `unstable` is excluded by the default heuristic."
- In the next row it could say: "The flag explicitly adds `unstable` to the heuristic's selections."
etc.

README.md
```diff
@@ -1,3 +1,4 @@
+
```
obi1kenobi, line 1:
```suggestion
```
Seems like an unintentional change, you probably could have caught this in a self-review here.

obi1kenobi (commented):
I like this! Just a few very minor comments and this is good to go.

README.md
staniewzki: Looks good!

README.md
staniewzki: I think this is a good idea, unfortunately now there are line breaks in the flags, but I don't think that is a big issue.

obi1kenobi: yeah, it's unfortunate but definitely the lesser problem, this is fine as-is

obi1kenobi:
Nicely done! I'm going to prep a release shortly.

Issue #414, opened 2023-03-08
Running `cargo semver-checks` with no other arguments should default to checking


Lokathor:
### Describe your use case

I ran `cargo semver-checks` and it said i needed to pick a command or argument.

### Describe the solution you'd like

Instead it should "do the right thing" and just do the checks if no command or argument is provided.

obi1kenobi:
100% agreed. Originally, I thought we might have multiple subcommands, but that turned out to not be the case and I've even considered deprecating `check-release` as a subcommand instead of just running it by default. I'm still not sure if we should go *that* far, but certainly running `check-release` by default would be an improvement.

If you'd be open to putting together a PR for this, I'd be thrilled to merge it.

obi1kenobi:
Related to: https://github.com/obi1kenobi/cargo-semver-checks/issues/86#issuecomment-1374672344

era:
@obi1kenobi I can send a PR changing the behaviour to run check-release by default.

obi1kenobi:
Sounds great, thank you!

obi1kenobi:
Fixed in #459 and shipped already, just forgot to close this issue!

PR #459, opened 2023-05-30
Running cargo semver-checks with no other arguments should default to checking

era:
Addresses #414.

It allows users to run cargo semver-checks without using the subcommand `check-release`.

I basically moved `CheckRelease` to be a field at `SemverChecks`. 

I tried to also remove it from the `SemverChecksCommands::CheckRelease` enum by using `global = true` on all fields from `CheckRelease`, but I could not add that property to `manifest` and `workspace` fields.

I reused the tests from `feature_config` to make sure the same use cases work when user specifies the subcommand or not.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -81,7 +81,7 @@ fn main() -> anyhow::Result<()> {
     };
     let report = check.check_release()?;
     if report.success() {
-        std::process::exit(0)
+        std::process::exit(0);
     } else {
         std::process::exit(1);
     }
--- a/src/main.rs
+++ b/src/main.rs
@@ -48,7 +48,7 @@ fn main() -> anyhow::Result<()> {
             )?;
         }
 
-        let mut config = GlobalConfig::new().set_level(args.verbosity.log_level());
+        let mut config = GlobalConfig::new().set_level(args.check_release.verbosity.log_level());
         config.shell_note("Use `--explain <id>` to see more details")?;
         std::process::exit(0);
     } else if let Some(id) = args.explain.as_deref() {
@@ -86,7 +86,13 @@ fn main() -> anyhow::Result<()> {
             }
         }
         None => {
-            anyhow::bail!("subcommand required");
+            let check: cargo_semver_checks::Check = args.check_release.into();
+            let report = check.check_release()?;
+            if report.success() {
+                std::process::exit(0)
+            } else {
+                std::process::exit(1);
+            }
         }
     }
 }
@@ -111,8 +117,8 @@ struct SemverChecks {
     #[arg(long, global = true, exclusive = true)]
     list: bool,
 
-    #[command(flatten)]
-    verbosity: clap_verbosity_flag::Verbosity<clap_verbosity_flag::InfoLevel>,
+    #[clap(flatten)]
+    check_release: CheckRelease,
 
     #[command(subcommand)]
     command: Option<SemverChecksCommands>,
--- a/src/main.rs
+++ b/src/main.rs
@@ -75,25 +75,15 @@ fn main() -> anyhow::Result<()> {
         std::process::exit(0);
     }
 
-    match args.command {
-        Some(SemverChecksCommands::CheckRelease(args)) => {
-            let check: cargo_semver_checks::Check = args.into();
-            let report = check.check_release()?;
-            if report.success() {
-                std::process::exit(0)
-            } else {
-                std::process::exit(1);
-            }
-        }
-        None => {
-            let check: cargo_semver_checks::Check = args.check_release.into();
-            let report = check.check_release()?;
-            if report.success() {
-                std::process::exit(0)
-            } else {
-                std::process::exit(1);
-            }
-        }
+    let check: cargo_semver_checks::Check = match args.command {
+        Some(SemverChecksCommands::CheckRelease(args)) => args.into(),
+        None => args.check_release.into(),
+    };
+    let report = check.check_release()?;
+    if report.success() {
+        std::process::exit(0)
+    } else {
+        std::process::exit(1);
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -86,7 +86,13 @@ fn main() -> anyhow::Result<()> {
...
         }
         None => {
-            anyhow::bail!("subcommand required");
+            let check: cargo_semver_checks::Check = args.check_release.into();
+            let report = check.check_release()?;
+            if report.success() {
+                std::process::exit(0)
+            } else {
+                std::process::exit(1);
+            }
```
obi1kenobi, line 95: Could we refactor the code here a bit to avoid the duplicated logic?

Maybe something like:
```rust
let check: cargo_semver_checks::Check = args.command.unwrap_or_else(|| args.check_release.into());

let report = check.check_release()?;
// ...
```

src/main.rs
```diff
@@ -111,8 +117,8 @@ struct SemverChecks {
     #[arg(long, global = true, exclusive = true)]
     list: bool,
 
-    #[command(flatten)]
-    verbosity: clap_verbosity_flag::Verbosity<clap_verbosity_flag::InfoLevel>,
```
obi1kenobi, line 115: I assume we don't need `verbosity` anymore because `CheckRelease` already handles it?

tests/feature_config.rs
```diff
@@ -17,6 +17,16 @@ impl CargoSemverChecks {
             ],
         }
     }
+    fn new(current_path: &str, baseline_path: &str) -> Self {
```
obi1kenobi, line 20:
```suggestion

    fn new(current_path: &str, baseline_path: &str) -> Self {
```

obi1kenobi: Just to match the surrounding style.

tests/feature_config.rs
```diff
@@ -30,6 +40,14 @@ impl CargoSemverChecks {
 
 #[test]
 fn simple_only_explicit_feature() {
+    CargoSemverChecks::new_with_check_release_subcommand(
+        "test_crates/features_simple/new/",
+        "test_crates/features_simple/old/Cargo.toml",
+    )
+    .add_arg("--only-explicit-features")
+    .run()
+    .success();
```
obi1kenobi, line 49: This duplication is a bit unfortunate: over time as we edit these tests and add new ones, we might update one version of the check but fail to update the other. This is a maintainability risk.

Could you think of another way to set up the test where we don't need to duplicate all the test code like this? Perhaps by doing some more refactoring of the `CargoSemverChecks` type?

obi1kenobi (commented):
Love the tests, great idea to use the existing test setup to ensure this functionality works too!

Just a few minor suggestions for further improvement.

src/main.rs
era: Correct, since we are flattening CheckRelease, we have the same verbosity flag available from the same places (with subcommand and not subcommand). Clap was actually the thing that highlighted it thanks to the "duplicated option" error.

era:
Let me know if the tests are better now, I thought of adding an `assert(fn(Assert) -> ()) -> ()` function to `CargoSemverChecks`, but I was not sure if exposing the vector wouldn't be the best move. 

Also, not sure about the way I remove the subcommand from args. Let me know if you think it could be better.

src/main.rs
```diff
@@ -75,19 +75,15 @@ fn main() -> anyhow::Result<()> {
...
-        None => {
-            anyhow::bail!("subcommand required");
-        }
+    let check: cargo_semver_checks::Check = match args.command {
+        Some(SemverChecksCommands::CheckRelease(args)) => args.into(),
+        None => args.check_release.into(),
+    };
+    let report = check.check_release()?;
+    if report.success() {
+        std::process::exit(0)
```
obi1kenobi, line 84:
```suggestion
        std::process::exit(0);
```
I just noticed the other branch has a `;` and this one didn't.

This was already present before you made any changes so you did the right thing by leaving it as-is. I'm familiar with the context of this code and the `;` vs not doesn't make a difference, but it's nice to be consistent :)

tests/feature_config.rs
```diff
@@ -18,13 +18,27 @@ impl CargoSemverChecks {
...
+    }
+
+    fn run_without_subcommand(&mut self) -> Assert {
+        let mut args = self.args.clone();
+        args.remove(SUBCOMMAND_ARGS_INDEX);
+        self.command().args(&args).assert()
+    }
+
+    fn run_with_subcommand(&mut self) -> Assert {
+        self.command().args(&self.args).assert()
```
obi1kenobi, line 41: Do we need to take `&mut self` on all of these? I'm not sure we're mutating anything on any of them.

tests/feature_config.rs
```diff
@@ -1,14 +1,14 @@
 use assert_cmd::{assert::Assert, Command};
 
+static SUBCOMMAND_ARGS_INDEX: usize = 1;
```
obi1kenobi, line 3: This is an implementation detail on `CargoSemverChecks`, so consider moving it into `impl CargoSemverChecks` instead. The invariant it corresponds to is the way the `args` value is constructed inside `CargoSemverChecks::new()`, so the constant should be near there as well.

Also, perhaps `const` instead of `static`? The difference is subtle and it's not a big deal either way.

obi1kenobi (commented):
Nicely done! A couple of tiny tweaks and this is good to go 🚀

era:
Thank you again @obi1kenobi for all the reviews!

tests/feature_config.rs
era: I think I cannot use a static inside the `impl` block, but changed it to const and moved it to the `impl CargoSemverChecks` :)

obi1kenobi (approved):
Awesome, let's ship it 🚀

PR #470, opened 2023-06-12
New lint: trait removed supertrait

### LINKED ISSUE #232: New lint: trait removed supertrait
Example of why this is breaking: say we removed the `DerefMut: Deref` supertrait relationship. Then, this code would break:
```rust
fn foo(x: impl Deref) {}

fn bar(x: impl DerefMut) {
    foo(x)
}
```

h/t https://twitter.com/i2talics/status/1559607755975057408

era:
Resolves #232

Besides the new lint, this PR also includes a newer version of trustfall-rustdoc-adapter because we need the `supertrait` field.

### DIFF
--- a/src/lints/trait_removed_supertrait.ron
+++ b/src/lints/trait_removed_supertrait.ron
@@ -1,7 +1,7 @@
 SemverQuery(
     id: "trait_removed_supertrait",
     human_readable_name: "supertrait removed or renamed",
-    description: "It is a breaking change to loosen generic bounds on a type since this can break users expecting the tightened bounds.",
+    description: "Removing a supertrait bound is a breaking change, since users of the trait can no longer assume it can also be used like its supertrait.",
     required_update: Major,
     reference_link: None,
     query: r#"
@@ -55,6 +55,6 @@ SemverQuery(
         "public": "public",
         "zero": 0,
     },
-    error_message: "A supertrait was removed from a trait.",
+    error_message: "A supertrait was removed from a trait, users of the trait can no longer assume it can also be used like its supertrait.",
     per_result_error_template: Some("supertrait {{join \"::\" trait_path}} of trait {{name}} in file {{span_filename}}:{{span_begin_line}}"),
 )
--- a/src/lints/trait_removed_supertrait.ron
+++ b/src/lints/trait_removed_supertrait.ron
@@ -1,0 +1,62 @@
+SemverQuery(
+    id: "trait_removed_supertrait",
+    human_readable_name: "supertrait removed or renamed",
+    description: "It is a breaking change to loosen generic bounds on a type since this can break users expecting the tightened bounds.",
+    required_update: Major,
+    reference_link: None,
+    query: r#"{
+        CrateDiff {
+            baseline {
+                item {
+                    ... on Trait {
+                        visibility_limit @filter(op: "=", value: ["$public"]) @output
+                        importable_path {
+                            path @output @tag
+                        }
+
+
+                        supertrait {
+                            trait {
+                                importable_path {
+                                    trait_path: path @output @tag
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            current {
+                item {
+                    ... on Trait {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        name @output
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                        }
+
+                        supertrait @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
+                            trait {
+                                importable_path {
+                                    path @filter(op: "=", value: ["%trait_path"])
+                                }
+                            }
+                        }
+
+                        span_: span @optional {
+                            filename @output
+                            begin_line @output
+                        }
+
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "zero": 0,
+    },
+    error_message: "A supertrait was removed from a trait.",
+    per_result_error_template: Some("supertrait {{trait_path}} for trait {{name}} in file {{span_filename}}:{{span_begin_line}}"),
+)
--- a/src/lints/trait_removed_supertrait.ron
+++ b/src/lints/trait_removed_supertrait.ron
@@ -3,7 +3,7 @@ SemverQuery(
     human_readable_name: "supertrait removed or renamed",
     description: "Removing a supertrait bound is a breaking change, since users of the trait can no longer assume it can also be used like its supertrait.",
     required_update: Major,
-    reference_link: None,
+    reference_link: Some("https://doc.rust-lang.org/reference/items/traits.html#supertraits"),
     query: r#"
     {
         CrateDiff {
--- a/src/lints/trait_removed_supertrait.ron
+++ b/src/lints/trait_removed_supertrait.ron
@@ -58,5 +58,5 @@ SemverQuery(
         "zero": 0,
     },
     error_message: "A supertrait was removed from a trait.",
-    per_result_error_template: Some("supertrait {{trait_path}} for trait {{name}} in file {{span_filename}}:{{span_begin_line}}"),
+    per_result_error_template: Some("supertrait {{join \"::\" trait_path}} of trait {{name}} in file {{span_filename}}:{{span_begin_line}}"),
 )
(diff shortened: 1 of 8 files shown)


obi1kenobi (commented):
This looks really good so far. The lint query is correct and the new test crate is nicely set up. We're 90% of the way there and I'm excited to merge this very soon!

Two remaining things to consider: user ergonomics when this lint reports a problem, and expanded test coverage to prevent future debugging.

Users of `cargo-semver-checks` might not know that removing a supertrait is a breaking change. Let's channel the helpfulness of rustc and tailor the error message and description in the lint to make them maximally helpful. This is one of the major challenges of writing software aimed at a very large audience like "everyone writing Rust" 😅 But that also makes it really rewarding to get it right!

Consider expanding the test suite a bit with more true-positive cases as well as guard test cases against possible false-positives.
- Example true-positive cases: removing one of multiple supertraits, removing supertrait when the subtrait is generic, or removing a generic supertrait.
- Example false-positive guard cases: removing a lifetime trait bound (`pub trait Foo<'a> : Bar + 'a` then lose the `'a`), changing the generic trait bound in a breaking way (to a different type, to a different lifetime, etc.) or in a non-breaking way (e.g. from `i64` to `Into<i64>`).

The guard cases are as important as the true-positive test cases, since false-positives are very harmful: they make users have a bad experience and make them not trust the tool, and they cause support load for us when issues get opened and we have to triage and fix them. An ounce of prevention is worth a pound of cure :)

To come up with good guard cases, consider:
- similar code changes that are non-breaking, and shouldn't be caught
- similar changes that are breaking but should be caught by _another_ lint — we have a policy that any breakage should be reported by one lint and one lint only, for the sake of user ergonomics.

Ping me with any questions! I'm super excited by all the awesome work you've been doing, and I'm very happy to support you along the way! 🚀

src/lints/trait_removed_supertrait.ron
```diff
@@ -0,0 +1,62 @@
+SemverQuery(
+    id: "trait_removed_supertrait",
+    human_readable_name: "supertrait removed or renamed",
+    description: "It is a breaking change to loosen generic bounds on a type since this can break users expecting the tightened bounds.",
```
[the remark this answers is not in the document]
era: Do you mind if we refine it together? It's a bit hard to be technically precise and simple in a few words. Trying to address the first part of your comment. We are talking about a very specific part of the "generic bounds", so:
"It is a breaking change to remove supertrait on a trait since this can break users expecting trait to also behave like the supertrait."

^ This is my OOP brain trying to translate the supertrait as inherence. Maybe there is a better way to put it?

About why it breaks, I'm not sure how to explain it without showing an example.  I tried to express that with the "users expecting trait to also behave like the supertrait" but not sure if it's enough.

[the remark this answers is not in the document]
obi1kenobi: Happy to help! I have a ton of stuff going on today so I probably won't be able to circle back to this until tomorrow, just wanted to let you know so you don't worry about it :)

src/lints/trait_removed_supertrait.ron
```diff
@@ -0,0 +1,60 @@
...
+    required_update: Major,
+    reference_link: None,
```
obi1kenobi, line 6: Any chance we can find something to cite here?

src/lints/trait_removed_supertrait.ron
```diff
@@ -0,0 +1,60 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "zero": 0,
+    },
+    error_message: "A supertrait was removed from a trait.",
```
obi1kenobi, line 58: Perhaps tweak this to include a hint of why this might be a problem. Feel free to use some of the language from the description field.

test_outputs/trait_removed_supertrait.output.ron
```diff
@@ -1,18 +1,102 @@
...
+                String("trait_removed_supertrait"),
+                String("SuperTrait2"),
+            ]),
+            "visibility_limit": String("public"),
+        },
+        {
+            "name": String("RemovingMultiple"),
+            "path": List([
+            String("trait_removed_supertrait"),
+            String("RemovingMultiple"),
```
obi1kenobi, line 77: something strange has happened with the indentation in a few places here
```suggestion
                String("trait_removed_supertrait"),
                String("RemovingMultiple"),
```

test_outputs/trait_removed_supertrait.output.ron
```diff
@@ -1,18 +1,102 @@
...
+            ]),
+            "span_begin_line": Uint64(10),
+            "span_filename": String("src/lib.rs"),
+            "trait_path": List([
+            String("trait_removed_supertrait"),
+            String("SuperTrait"),
```
obi1kenobi, line 83:
```suggestion
                String("trait_removed_supertrait"),
                String("SuperTrait"),
```

test_outputs/trait_removed_supertrait.output.ron
```diff
@@ -1,18 +1,102 @@
...
+            ]),
+            "visibility_limit": String("public"),
+        },
+        {
+            "name": String("RemovingMultiple"),
+            "path": List([
+            String("trait_removed_supertrait"),
+            String("RemovingMultiple"),
```
obi1kenobi, line 91:
```suggestion
                String("trait_removed_supertrait"),
                String("RemovingMultiple"),
```

test_outputs/trait_removed_supertrait.output.ron
```diff
@@ -1,18 +1,102 @@
...
+            ]),
+            "span_begin_line": Uint64(10),
+            "span_filename": String("src/lib.rs"),
+            "trait_path": List([
+            String("trait_removed_supertrait"),
+            String("SuperTrait2"),
```
obi1kenobi, line 97:
```suggestion
                String("trait_removed_supertrait"),
                String("SuperTrait2"),
```

src/lints/trait_removed_supertrait.ron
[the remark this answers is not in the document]
era: Ah, thanks! That looks perfect and simple!

src/lints/trait_removed_supertrait.ron
era: I should have asked about this earlier, on the `semver` reference does not seem to be an entry for this lint.

There is [this](https://doc.rust-lang.org/cargo/reference/semver.html#minor-loosening-generic-bounds) that is somewhat similar but it's definitely not the exact case nor does the explanation help (IMHO).

The [documentation](https://doc.rust-lang.org/reference/items/traits.html#supertraits) about supertraits helps to understand what is breaking indirectly.

Is it worth adding any of those links?

obi1kenobi: Yes, the docs link for supertraits is the best option. It's not perfect, but it's certainly better than nothing.

src/lints/trait_removed_supertrait.ron
```diff
@@ -55,6 +55,6 @@ SemverQuery(
         "public": "public",
         "zero": 0,
     },
-    error_message: "A supertrait was removed from a trait.",
+    error_message: "A supertrait was removed from a trait, users of the trait can no longer assume it can also be used like its supertrait.",
```
obi1kenobi, line 58:
```suggestion
    error_message: "A supertrait was removed from a trait. Users of the trait can no longer assume it can also be used like its supertrait.",
```
this seemed to read a bit better to me

obi1kenobi (commented):
This is great. Last couple of minor things and this is ready to merge and release!

src/lints/trait_removed_supertrait.ron
era: Thank you, I will add it :)

era:
Thank you again for all the mentoring during the PR, I'm super grateful for all the time you invested in it and all the comments you add.

Super thank you!

obi1kenobi (approved):

obi1kenobi:
No worries, I'm happy I could help. Thank you for all the hard work you've been putting in to make `cargo-semver-checks` better!

Now that you've seen a few different sides of `cargo-semver-checks`, do you have a preference between writing new lints versus improving the user-facing side of the tool?

How can I best support you and make sure you're set up for success?

era:
No preferences :), I liked the process of both. I will review the list you gave me and pick something interesting from there.

And at the moment I don't need anything specific, I think now I have a better macro picture of the whole components. Thank you again for all the reviews!

PR #488, opened 2023-07-11
POC: Sparse protocol support

inflation:
Quick and dirty way to support sparse protocol and set it by default.

The proper way is to support registry selection and protocol type detection.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,6 +37,7 @@ pub struct Check {
     release_type: Option<ReleaseType>,
     current_feature_config: rustdoc_gen::FeatureConfig,
     baseline_feature_config: rustdoc_gen::FeatureConfig,
+    git: bool,
 }
 
 /// The kind of release we're making.
@@ -226,7 +227,7 @@ impl Scope {
 }
 
 impl Check {
-    pub fn new(current: Rustdoc) -> Self {
+    pub fn new(current: Rustdoc, git: bool) -> Self {
         Self {
             scope: Scope::default(),
             current,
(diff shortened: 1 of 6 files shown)


obi1kenobi:
Very interesting, thanks for putting this POC together!

Regarding the protocol type detection, is there a good way for us to use `crates_index` to do that? Or is that something that `crates_index` will have to build and expose to us?

Also, by using sparse protocol by default, are we likely to expose Windows users to this bug? https://github.com/frewsxcv/rust-crates-index/issues/126

inflation:
I guess it's one of the internal things that `cargo` should "cratelize" for reusing. It need to parse through the hierarchy of `config.toml` to figure out the index URL of the specified registry. Then we can determine if it's sparse or not by checking the `sparse+` prefix. 

Windows should be fine since `reqwest` handles it anyway.

src/lib.rs
```diff
@@ -37,6 +37,7 @@ pub struct Check {
     release_type: Option<ReleaseType>,
     current_feature_config: rustdoc_gen::FeatureConfig,
     baseline_feature_config: rustdoc_gen::FeatureConfig,
+    git: bool,
```
inflation, line 40: Need to store the HTTP client

obi1kenobi:
Since this PR, the `crates_index` became unmaintained but then thanks to the author of `tame-index`, we switched to that crate and gained sparse index support in #506.

The next release of `cargo-semver-checks` will have sparse index support 🎉 Thanks for the help and for your patience on this!

PR #490, opened 2023-07-13
Ensure missing span_filename and span_begin_line shows user friendly error

gliu20:
1. Shows user-friendly error when span_filename and span_begin_line are missing
2. Updates contributing guide to guide people when they encounter the error while also making a note to prevent people from making this mistake

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -146,6 +146,14 @@ Now it's time to fill in these files!
   removal of public fields were to report that a struct was removed.
   Struct removal has its own lint, so there's no reason to also report that
   the removed struct also had its fields removed.
+- Note that your lint must output `span_filename` and `span_begin_line` for it
+  to be a valid lint. The pattern we commonly use is:
+  ```
+  span_: span @optional {
+    filename @output
+    begin_line @output
+  }
+  ```
 - Re-run [`./scripts/regenerate_test_rustdocs.sh`](https://github.com/obi1kenobi/cargo-semver-checks/tree/main/scripts/regenerate_test_rustdocs.sh)
   to generate rustdoc JSON files for your new test crates.
 
@@ -184,6 +192,22 @@ Actual output:
     ],
 }
 ```
+
+If your lint fails with the following error:
+```
+---- query::tests_lints::struct_missing stdout ----
+thread 'query::tests_lints::struct_missing' panicked at 'a valid query must have span_filename', src/query.rs:390:22
+note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
+```
+
+It likely means that your lint does not specify the `span_filename` and `span_begin_line` of where the error occurs. To fix this, add the following to the part of query that catches the error:
+```
+span_: span @optional {
+  filename @output
+  begin_line @output
+}
+```
+
 
 Inspect the "actual" output:
 - Does it report the semver issue your lint was supposed to catch? If not, the lint query
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -212,7 +212,7 @@ Congrats on the new lint!
 If your lint fails with an error similar to the following:
 ```
 ---- query::tests_lints::enum_missing stdout ----
-thread 'query::tests_lints::enum_missing' panicked at 'A valid query must output both span_filename and span_begin_line. See https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md for how to do this.', src/query.rs:395:26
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -185,6 +193,22 @@ Actual output:
...
+
+It likely means that your lint does not specify the `span_filename` and `span_begin_line` of where the error occurs. To fix this, add the following to the part of query that catches the error:
+```
+span_: span @optional {
+  filename @output
+  begin_line @output
+}
+```
+
+
```
obi1kenobi, line 211:
```suggestion
```

CONTRIBUTING.md
```diff
@@ -146,6 +146,14 @@ Now it's time to fill in these files!
...
   Struct removal has its own lint, so there's no reason to also report that
   the removed struct also had its fields removed.
+- Note that your lint must output `span_filename` and `span_begin_line` for it
+  to be a valid lint. The pattern we commonly use is:
+  ```
+  span_: span @optional {
+    filename @output
+    begin_line @output
+  }
+  ```
```
obi1kenobi, line 156: Do you think this note should be here, or next to the "define the lint" part? Is this where it seems more intuitive? I have the contributing steps in muscle memory so I genuinely don't know, happy to leave it here if this is where you think it's best.

src/query.rs
```diff
@@ -385,10 +385,15 @@ mod tests {
...
-                    elem["span_filename"].as_str().unwrap().to_owned(),
-                    elem["span_begin_line"].as_usize().unwrap(),
-                )
+                let filename = elem
+                    .get("span_filename")
+                    .and_then(|value| value.as_str())
+                    .expect("a valid query must have span_filename");
+                let line = elem
+                    .get("span_begin_line")
+                    .expect("a valid query must have span_begin_line");
```
obi1kenobi, line 394: Two minor suggestions:
- Can we check if _both_ values are missing, and if so then report them both missing at once?
- Can we make the `.expect()` text directly point the user to the relevant portion of the `CONTRIBUTING.md` document (perhaps even linking to it on GitHub?) so that they don't have to be confused about what the error means?

obi1kenobi (commented):
Thanks for putting this together, I'm happy to merge it since it's a clear improvement. Just wanted to bring up a few additional options to consider, in case you think they are further improvements over what this PR already brings.

src/query.rs
gliu20: Yup will do! About the linking to the relevant section, do we need to make some sort of heading somewhere so we can do `#heading_name` to link it?

CONTRIBUTING.md
gliu20: Yup, this should be moved

src/query.rs
obi1kenobi: Certainly open to it if you think it's best, I trust your judgment. It could also be as simple as just linking to the file and mentioning that `span_filename` and `span_begin_line` should be output like that file describes, and we could improve it further from there if needed in the future.

src/query.rs
```diff
@@ -385,10 +385,15 @@ mod tests {
...
-                    elem["span_begin_line"].as_usize().unwrap(),
-                )
+                let filename = elem.get("span_filename").and_then(|value| value.as_str());
+                let line = elem.get("span_begin_line");
+
+                match (filename, line) {
+                    (Some(filename), Some(line)) => (filename.to_owned(), line.as_usize()),
+                    (Some(_filename), _) => panic!("A valid query must have span_filename. See https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md for details."),
+                    (_, Some(_line)) => panic!("A valid query must have span_begin_line. See https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md for details."),
+                    _ => panic!("A valid query must have both span_filename and span_begin_line. See https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md for details."),
```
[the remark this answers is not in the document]
gliu20: Will do! 

> Also perhaps let's delimit the column names with backticks to make it clear they are (effectively) code?
I'm not sure what you mean. What columns are you referring to?

src/query.rs
gliu20: Sounds good

obi1kenobi (approved):
Nice! Ship it 🚀

src/query.rs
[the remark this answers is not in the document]
obi1kenobi: I meant the output names `span_filename` and `span_begin_line`.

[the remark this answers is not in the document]
gliu20: Done!

obi1kenobi (approved):

Issue #366, opened 2023-02-10
New schema and lints for const and static values


obi1kenobi:
Add schema to represent `const` and `static` items. This will require new types for each in our query schema.

For now, ignore the type of the item. Adding type info to the schema is blocked on #149 which is a very hard issue (generics, lifetimes, HRTBs, effects, oh my!).

New lints enabled by this:
- pub module-level const missing
- pub module-level static missing
- const inside type's impl block missing
- mutable static became immutable static

All semver-major changes, because the item cannot be imported (it must have been renamed or deleted) and downstream is broken.

EDIT: Some of these have already been implemented, please check the rest of the thread to see what's left 👇

obi1kenobi:
Almost completely implemented, only remaining missing piece is associated constants. Waiting to hear back in rust-lang Zulip for that: https://rust-lang.zulipchat.com/#narrow/stream/266220-rustdoc/topic/Rustdoc.20JSON.3A.20How.20.60const.60.20items.20are.20represented

obi1kenobi:
Remaining lints here:
- [ ] removed const in trait 
- [ ] removed pub const inside impl
- [ ] mutable static became immutable static

eugenesvk:
Should a change in the **value** of a pub constant be tracked in a lint in addition to a change in name/visibility?

obi1kenobi:
I wouldn't make it a hard error like the other kinds of breaking changes, but definitely could be a very reasonable warn-level lint.

We don't currently have the means to allow users to control error/warn/ignore levels, or enable/disable lints, so I wouldn't add it right away. There are also situations where the value of the pub constant might not have changed but its rustdoc value might change from a literal value to a `_` which means something like "undetermined," so there's some nuance to worry about.

If this is something you might be interested in working toward, I'd be happy to mentor for PRs and point you in the right direction!

eugenesvk:
Thanks for the offer, but I was only interested in this crate as a way to extract a list of constants and their types/values, and this seems to be based in your lower-level trustfall ecosystem, so to that end I've tried to patch it like https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/280, maybe that could be a helpful base for the semver checks if you could take at look there?

PR #491, opened 2023-07-13
Add lint for missing pub module-level const

gliu20:
This addresses part of #366

### DIFF
--- a/test_crates/pub_module_level_const_missing/new/src/lib.rs
+++ b/test_crates/pub_module_level_const_missing/new/src/lib.rs
@@ -1,37 +1,32 @@
-
-pub const pub_const_in_global: i32 = 0;
-pub const pub_const_in_global_is_renamed: i32 = 0;
-pub static pub_const_in_global_will_be_static: i32 = 0;
-const pub_const_in_global_will_be_private_const: i32 = 0;
-static pub_const_in_global_will_be_private_static: i32 = 0;
+pub const PUB_CONST_IN_GLOBAL: i32 = 0;
+pub const PUB_CONST_IN_GLOBAL_IS_RENAMED: i32 = 0;
+pub static PUB_CONST_IN_GLOBAL_WILL_BE_STATIC: i32 = 0;
+const PUB_CONST_IN_GLOBAL_WILL_BE_PRIVATE_CONST: i32 = 0;
+static PUB_CONST_IN_GLOBAL_WILL_BE_PRIVATE_STATIC: i32 = 0;
 
 mod re_exporter {
-    pub const special_const: i32 = 0;
-    pub static special_static: i32 = 0;
+    pub const SPECIAL_CONST: i32 = 0;
+    pub static SPECIAL_STATIC: i32 = 0;
 }
 
 pub mod my_module {
-    pub const pub_const_in_module: i32 = 0;
-    pub const pub_const_in_module_is_renamed: i32 = 0;
-    pub static pub_const_in_module_will_be_static: i32 = 0;
-    const pub_const_in_module_will_be_private_const: i32 = 0;
-    static pub_const_in_module_will_be_private_static: i32 = 0;
+    pub const PUB_CONST_IN_MODULE: i32 = 0;
+    pub const PUB_CONST_IN_MODULE_IS_RENAMED: i32 = 0;
+    pub static PUB_CONST_IN_MODULE_WILL_BE_STATIC: i32 = 0;
+    const PUB_CONST_IN_MODULE_WILL_BE_PRIVATE_CONST: i32 = 0;
+    static PUB_CONST_IN_MODULE_WILL_BE_PRIVATE_STATIC: i32 = 0;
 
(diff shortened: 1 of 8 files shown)


test_outputs/pub_module_level_const_missing.output.ron
```diff
@@ -0,0 +1,118 @@
...
+            "visibility_limit": String("public"),
+        },
+        {
+            "name": String("pub_const_in_module_will_be_private_static"),
+            "span_begin_line": Uint64(15),
+            "span_filename": String("src/lib.rs"),
+            "visibility_limit": String("public"),
+        },
+        {
+            "name": String("pub_const_in_module_will_re_export"),
```
gliu20, line 64: This being marked as a missing pub const is unexpected (since my understanding is that we are re-exporting). For reference:

**Current**
```rust
mod re_exporter {
    pub const special_const: i32 = 0;
    pub static special_static: i32 = 0;
}
pub mod my_module {    
    pub use crate::re_exporter::special_const as 
        pub_const_in_module_will_re_export;
}
```
**Baseline**
```rust
pub mod my_module {
    pub const pub_const_in_module_will_re_export: i32 = 0;
}
```

test_outputs/pub_module_level_const_missing.output.ron
```diff
@@ -0,0 +1,118 @@
...
+            "visibility_limit": String("public"),
+        },
+        {
+            "name": String("pub_const_in_nested_module_will_be_private_static"),
+            "span_begin_line": Uint64(25),
+            "span_filename": String("src/lib.rs"),
+            "visibility_limit": String("public"),
+        },
+        {
+            "name": String("pub_const_in_nested_module_will_re_export"),
```
gliu20, line 106: Same issue here

test_outputs/pub_module_level_const_missing.output.ron
gliu20: After some more digging, it appears that rust exports don't inherit the static/const-ness of the original variable and must be declared separately... will update tests to reflect that

obi1kenobi: Oh, interesting! Would love to learn more about the experiment you ran to find this, because this seems unexpected to me and I'm a bit worried it might be a bug in my code.

test_crates/pub_module_level_const_missing/new/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+        pub const pub_const_in_nested_module_is_renamed: i32 = 0;
+        pub static pub_const_in_nested_module_will_be_static: i32 = 0;
+        const pub_const_in_nested_module_will_be_private: i32 = 0;
+
+        pub use crate::re_exporter::special_const as 
+            pub_const_in_nested_module_will_re_export;
+        pub use crate::re_exporter::special_static as 
+            pub_const_in_nested_module_will_re_export_static;
+    }
+}
```
obi1kenobi, line 37: This file is missing a trailing newline. Your editor can be configured to add those automatically, and I'd highly recommend looking into that.

obi1kenobi (commented):
This looks headed in a good direction, good job so far!

I think usual Rust style is to use `ALL_CAPS` for const and static items, and I think `clippy` will complain about the naming in the test crates. Would you mind double-checking this and then if needed uppercasing the const and static names in the test crates?

test_outputs/pub_module_level_const_missing.output.ron
gliu20: Sorry, I think I was mistaken. I was under the impression that `pub const use` or `pub static use` was something that existed in rust after reading some stackoverflow answers. After some testing, it seems that rust does inherit the static/const-ness of the original variable, though some confirmation would be nice since I can't get the compiler to spit out whether something is a static or const and trying to assign to a re-exported const doesn't tell me much since doing so is invalid in the first place.

test_crates/pub_module_level_const_missing/new/src/lib.rs
gliu20: Fixed! Installed rust-analyzer and formatOnSave is on so shouldn't happen again

obi1kenobi (commented):
I'm going to look into the re-export issue you're seeing, it's possible the bug is in my code.

So we can parallelize work, could you try in the meantime to find a good reference link for why this is a major breaking change, and also work on the remaining TODOs in the description and user-facing error text? Aim for correct but concise text in all those — we really want them as short as feasible while conveying all the important info.

test_outputs/pub_module_level_const_missing.output.ron
obi1kenobi: Based on this test case I just added, it seems that `pub use` works correctly and normally over both `const` and `static` items :+1:

I'm not sure why the test cases here are failing, but I'm happy to take a deeper look if you'd like me to.

obi1kenobi (approved):
Excellent work! 🚀

PR #494, opened 2023-07-13
Prettier test diff output via `similar-asserts`.

obi1kenobi:


obi1kenobi:
Example of what the test output is after this PR:
image

tonowak:
That's soo good :fire:

PR #496, opened 2023-07-15
New lint: Pub module level static missing

gliu20:
Implements the equivalent lint for `public static`s as was done for `public const`s. Note that per our discussion, we explicitly allow `static` to convert to `const` as there are instances where this is a non-breaking change.

edit: Addresses bullet point 2 of #366


gliu20:
FYI I'm not sure if it's an issue on my side or with how github displays files, but VS Code (in Github Codespaces) shows that a trailing newline is present but the online viewer doesn't.

src/lints/pub_module_level_static_missing.ron
```diff
@@ -0,0 +1,47 @@
...
+                        span_: span @optional {
+                            filename @output
+                            begin_line @output
+                        }
+                    }
+                }
+            }
+            current {
+                item @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
+                    # A GlobalValue is a supertype of Static and Const
```
obi1kenobi, line 29:
```suggestion
                    # A GlobalValue is a supertype of Static and Constant
```
to match the type's name in the schema

src/lints/pub_module_level_static_missing.ron
```diff
@@ -0,0 +1,47 @@
+SemverQuery(
+    id: "pub_module_level_static_missing",
+    human_readable_name: "pub module-level static is missing",
```
obi1kenobi, line 3: Unlike consts, statics are not allowed to be associated with a type:
image

After looking into this, this limitation seems very unlikely to be changed anytime soon, since the semantics of such a feature would be quite a knot to untangle.

So in name of brevity let's remove `module level` from all the names and explanations, since all statics for the foreseeable future are always going to be module-level?

obi1kenobi (commented):
This looks great! I have two tiny nits that should be fixable in just one minute or so, and this is good to merge!

obi1kenobi:
> FYI I'm not sure if it's an issue on my side or with how github displays files, but VS Code (in Github Codespaces) shows that a trailing newline is present but the online viewer doesn't.

At the moment I'm seeing all files ending with newlines. What does your editor show?

I'd assume the online viewer is authoritative since it's looking at the file state in the GitHub repo. But who knows! Weird bug, consider reporting it if you can reproduce it easily!

src/lints/pub_static_missing.ron
```diff
@@ -1,5 +1,5 @@
 SemverQuery(
-    id: "pub_module_level_static_missing",
+    id: "pub_static_missing",
     human_readable_name: "pub module-level static is missing",
```
obi1kenobi, line 3:
```suggestion
    human_readable_name: "pub static is missing",
```
I think this is the last little thing and this is ready to merge.

obi1kenobi (approved):
Applied the last rename in the description and queued up for merging! 🚀

gliu20:
Sweet!!!! Thanks again for the review and help!!! 🎉

obi1kenobi:
Thanks to you for all the work you've put in, especially with all the thorough tests you've added. Teamwork makes the dream work! 🚀

Issue #487, opened 2023-07-10
Can't use sparse registry mirror 


inflation:
### Steps to reproduce the bug with the above code

I'm having this config for the mirror of crates.io:

```toml
[source.crates-io]
replace-with = 'tuna'

[source.tuna]
registry = "sparse+https://mirrors.tuna.tsinghua.edu.cn/crates.io-index/"

[registries.tuna]
index = "sparse+https://mirrors.tuna.tsinghua.edu.cn/crates.io-index/"
```

### Actual Behaviour

`cargo semver-checks check-release` failed with:
```
Error: The repo at path /Users/inflation/.cargo/registry/index/mirrors.tuna.tsinghua.edu.cn-2eab394af869c8a2 is unusable due to having an invalid HEAD reference: reference 'refs/heads/master' not found; class=Reference (4); code=NotFound (-3)
```

### Expected Behaviour

Checks normally

#### Software version

cargo-semver-checks 0.22.1

#### Operating system

macOS 13.4.1 (Darwin 22.5.0)

#### Command-line

```bash
/Users/inflation/.cargo/bin/cargo-semver-checks semver-checks --bugreport
```

#### cargo version

```
> cargo -V
cargo 1.70.0 (ec8a8a0ca 2023-04-25)
```

#### Compile time information

- Profile: release
- Target triple: aarch64-apple-darwin
- Family: unix
- OS: macos
- Architecture: aarch64
- Pointer width: 64
- Endian: little
- CPU features: aes,crc,dit,dotprod,dpb,dpb2,fcma,fhm,flagm,fp16,frintts,jsconv,lor,lse,neon,paca,pacg,pan,pmuv3,ras,rcpc,rcpc2,rdm,sb,sha2,sha3,ssbs,v8.1a,v8.2a,v8.3a,v8.4a,vh
- Host: aarch64-apple-darwin

obi1kenobi:
Interesting, thanks for the report! I've never used registry mirrors, is there somewhere I can read about what benefits they provide and in what circumstances they get used?

`cargo-semver-checks` uses the `crates_index` crate to interact with the crate index and the local cargo cache. So we support whatever that crate supports, and like I mentioned, I haven't looked at mirror functionality so I'm not sure whether this is supported or not.

The error message seems to indicate that, for this configuration, the underlying `crates_index` crate expected to find a git repo but did not. Is it possible this is due to a configuration error?

The underlying code in question is here:
[...]

Issue #160, opened 2022-10-26
Custom registry does not seem supported


nbigaouette:
### Describe your use case

I'd like to integrate `cargo-semver-check` workflow into CI at work where we use a custom crate registry to (internally) publish our crates.

Unfortunately, `cargo-semver-check` fails to run on a crate that is published on the custom registry:

```
❯ cargo semver-checks check-release --verbose
    Updating index
     Parsing mycrate v0.1.0 (current)
 Documenting mycrate v0.1.0 (~/mycrate.git)
    Finished dev [unoptimized + debuginfo] target(s) in 0.96s
Error: mycrate not found in registry
```

I suspect this is because it is not aware of the crate registry when trying to download the crate.

### Describe the solution you'd like

`cargo-semver-check` should fetch the published package from the crate registry it has been published previously.

In my case, this information is located in the repo's `.cargo/config`, but I think that info can also be located elsewhere.

obi1kenobi:
Thanks for the detailed report!

I think we both want to add custom registry support, and in the meantime also provide a better, more actionable error message than the current one.

While the registry-based previous version lookup won't work with a custom registry, you should be able to use one of the other ways to tell `cargo-semver-checks` about the previous version to compare against. Excerpt from `cargo semver-checks check-release --help`:
```
      --baseline-rev <REV>
          Git revision to lookup for a baseline

      --baseline-root <MANIFEST_ROOT>
          Directory containing baseline crate source

      --baseline-rustdoc <JSON_PATH>
          The rustdoc json file to use as a semver baseline
```

I think all of those should be unaffected by the custom registry setup, and should able to unblock your current use case today. Please let us know if that works!

obi1kenobi:
I'm adding an entry to the FAQ about using the other baseline generation options for custom registries until this issue is resolved: #165

pwnorbitals:
This issue was "closed as completed", does this mean custom registries are now supported ? Otherwise, I would suggest leaving this issue open, as an indication the project could accept PRs to add support

PR #506, opened 2023-07-26
Replace `crates-index` with `tame-index`

Jake-Shadle:
This PR replaced `crates-index` with `tame-index` which gives more friendly access to the crates.io index, whether it is git, sparse, or source replaced with another git or sparse registry.

Resolves: #487

### DIFF
--- a/src/rustdoc_gen.rs
+++ b/src/rustdoc_gen.rs
@@ -1,10 +1,10 @@
 use std::path::PathBuf;
 
-use anyhow::Context;
-use crates_index::Crate;
+use anyhow::Context as _;
 use itertools::Itertools;
 use serde::Serialize;
 use sha2::{Digest, Sha256};
+use tame_index::IndexKrate;
 
 use crate::manifest::Manifest;
 use crate::rustdoc_cmd::RustdocCommand;
@@ -13,21 +13,26 @@ use crate::GlobalConfig;
 
 #[derive(Debug, Clone)]
 pub(crate) enum CrateSource<'a> {
-    Registry { crate_: &'a crates_index::Version },
-    ManifestPath { manifest: &'a Manifest },
+    Registry {
+        crate_: &'a tame_index::IndexVersion,
+        version: String,
+    },
+    ManifestPath {
+        manifest: &'a Manifest,
+    },
 }
 
 impl<'a> CrateSource<'a> {
     pub(crate) fn name(&self) -> anyhow::Result<&str> {
         Ok(match self {
-            Self::Registry { crate_ } => crate_.name(),
+            Self::Registry { crate_, .. } => &crate_.name,
             Self::ManifestPath { manifest } => crate::manifest::get_package_name(manifest)?,
         })
     }
 
     pub(crate) fn version(&self) -> anyhow::Result<&str> {
         Ok(match self {
-            Self::Registry { crate_ } => crate_.version(),
+            Self::Registry { version, .. } => version.as_str(),
             Self::ManifestPath { manifest } => crate::manifest::get_package_version(manifest)?,
         })
     }
@@ -49,7 +54,7 @@ impl<'a> CrateSource<'a> {
     /// <https://doc.rust-lang.org/cargo/reference/features.html#the-features-section>
     pub(crate) fn regular_features(&self) -> Vec<String> {
         match self {
-            Self::Registry { crate_ } => crate_.features().keys().cloned().collect(),
+            Self::Registry { crate_, .. } => crate_.features().map(|(k, _v)| k).cloned().collect(),
             Self::ManifestPath { manifest } => manifest.parsed.features.keys().cloned().collect(),
         }
     }
@@ -58,11 +63,10 @@ impl<'a> CrateSource<'a> {
     /// <https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies>
     pub(crate) fn implicit_features(&self) -> std::collections::BTreeSet<String> {
         let mut implicit_features: std::collections::BTreeSet<_> = match self {
-            Self::Registry { crate_ } => crate_
+            Self::Registry { crate_, .. } => crate_
                 .dependencies()
                 .iter()
-                .filter_map(|dep| dep.is_optional().then_some(dep.name()))
-                .map(|x| x.to_string())
+                .filter_map(|dep| dep.is_optional().then(|| dep.name.to_string()))
                 .collect(),
             Self::ManifestPath { manifest } => {
                 let mut dependencies = manifest.parsed.dependencies.clone();
@@ -75,14 +79,13 @@ impl<'a> CrateSource<'a> {
                 }
                 dependencies
                     .iter()
-                    .filter_map(|(name, dep)| dep.optional().then_some(name))
-                    .map(|x| x.to_string())
+                    .filter_map(|(name, dep)| dep.optional().then(|| name.clone()))
                     .collect()
             }
         };
 
         let feature_defns: Vec<&String> = match self {
-            Self::Registry { crate_ } => crate_.features().values().flatten().collect(),
+            Self::Registry { crate_, .. } => crate_.features().flat_map(|(_k, v)| v).collect(),
             Self::ManifestPath { manifest } => {
                 manifest.parsed.features.values().flatten().collect()
             }
@@ -582,7 +585,7 @@ fn bytes2str(b: &[u8]) -> &std::ffi::OsStr {
 pub(crate) struct RustdocFromRegistry {
     target_root: PathBuf,
     version: Option<semver::Version>,
-    index: crates_index::Index,
+    index: tame_index::index::ComboIndex,
 }
 
 impl core::fmt::Debug for RustdocFromRegistry {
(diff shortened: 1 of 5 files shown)


Jake-Shadle:
Oh, didn't realize this crate tested in several old versions, I'll relax it to 1.67.0.

obi1kenobi:
> This was fixed in [EmbarkStudios@a966a4e](https://github.com/EmbarkStudios/cargo-semver-checks/commit/a966a4e840dc1aa29980aeb4787b6d6340be3233), not sure why the commit isn't being picked up in this PR...

Another contributor in another repo reported the same problem yesterday. I have to assume this is a GitHub bug.

A retry like `git commit --amend --no-edit` followed by a force-push of the branch seemed to resolve it that time.

src/rustdoc_gen.rs
```diff
@@ -597,13 +600,56 @@ impl core::fmt::Debug for RustdocFromRegistry {
...
-        let mut index = crates_index::Index::new_cargo_default()?;
+        let index_url = tame_index::IndexUrl::crates_io(
+            // This is the config root, where .cargo/config.toml configuration files
+            // are crawled to determine if crates.io has been source replaced
+            // <https://doc.rust-lang.org/cargo/reference/source-replacement.html>
+            // if not specified it defaults to the current working directory,
+            // which is the same default that cargo uses, though note this can be
+            // extremely confusing if one can specify the manifest path of the
+            // crate from a different current working directory, though AFAICT
+            // this is not how this binary works
```
obi1kenobi, line 611: The `--manifest-path` flag can be used to select a crate from a different working directory, so I don't believe this is accurate. (The plan is to merge `cargo-semver-checks` into cargo itself, so our flags try as much as possible to behave as cargo's own.)

At the point where `RustdocFromRegistry` is constructed, the `--manifest-path` value (if any) is not present anymore. But as one possible solution, we could put the `index_url` value inside `cargo_semver_checks::Check` and then inside `GlobalConfig`, which will make it available throughout. Or if the better thing to carry around is a `ComboIndexCache`, we could do the same with that struct.

Just one thing to note: it's possible to use the binary and library here without an index. For example, if the user has already generated baseline and current rustdoc JSON files and just wants to semver-check them. This shouldn't fail with a network or configuration error on account of the (unused) index.

obi1kenobi (commented):
This seems very nice! I like many of the changes you've decided to make relative to `crates-index` and I'm excited to merge this.

I flagged one concern that I believe should be possible to resolve without too much additional trouble.

Also, one more question: would this PR also enable our users to use mirrors of crates.io via source replacement? That's been a commonly requested feature (#160) and I was wondering if this PR might also happen to resolve that as well.

src/rustdoc_gen.rs
Jake-Shadle: Yah I noticed that trying to debug why a test was failing, I think for now I'll keep this as it is since it will follow the same cargo logic of using configuration based on the current directory rather than the parent of the manifest, but IMO that logic is incredibly confusing if you don't know about it.

Jake-Shadle:
> Also, one more question: would this PR also enable our users to use mirrors of crates.io via source replacement? That's been a commonly requested feature (https://github.com/obi1kenobi/cargo-semver-checks/issues/160) and I was wondering if this PR might also happen to resolve that as well.

For source replacement of crates.io, yes, this PR resolves that ([IndexUrl::crates_io](https://docs.rs/tame-index/0.2.2/tame_index/index/location/enum.IndexUrl.html#method.crates_io)), however #160 could mean both source replacement of crates.io or custom registries in addition to or instead of crates.io, which this PR alone wouldn't resolve, as each unique index would need to be opened based on the set of unique source indices, similar to what [this code](https://github.com/EmbarkStudios/cargo-deny/blob/3fce2031de75932099e2cf1bbd905f38f339fc2a/src/advisories/helpers/index.rs#L13-L68) is doing in cargo-deny.

src/rustdoc_gen.rs
obi1kenobi: Indeed, I did not know about it and I am finding it confusing! But I agree, matching what cargo does is the right call here.

obi1kenobi:
The code looks good to me! If you could try to amend your latest commit and force-push it to work around this GitHub glitch we seem to be having, I'm happy to merge as soon as CI passes.

Thanks again for letting me know about your fork of `crates-index` and for making this PR. `cargo-semver-checks` is much better off for your expertise and hard work here 🚀 

If you might be open to adding support for custom registries at any point, I'd love to have your help on that as well!

Jake-Shadle:
Sorry should have been more clear, the 1.67.0 problem was because of rust-version is fixed, however there is a legitimate bug in the current code that my testing didn't catch because of some peculiarities with cargo, fixing that now before rerunning the tests here.

Jake-Shadle:
Ok, so the error in https://github.com/obi1kenobi/cargo-semver-checks/actions/runs/5671264307/job/15368054854?pr=506 is resolved at the head of the branch I patched to since I could repro the CI environment locally in a container, the issue seems to be that it is using a cached binary where the cache key isn't accounting for Cargo.lock changes, so not sure how to fix that, but I'll publish the fixed version and change this PR to use that published version.

Jake-Shadle:
Oh, and this PR broke the 1.64.0 check, I thought this crate only cared about 1.65.0 so set the rust-version to that, I'll update it to 1.64.0 as otherwise that job will succeed, but by failing on an error that bypasses the actual failure that job is intended to test.

Jake-Shadle:
Oh wait, furthest tame-index can go back is 1.65.0, which is the lowest version supported by gix https://github.com/search?q=repo%3AByron%2Fgitoxide%20rust-version&type=code

Jake-Shadle:
Ahh, so nvm looking how the caching works the cache key is correct and it is using the correct version, I'll continue taking a look tomorrow.

obi1kenobi:
> Oh, and this PR broke the 1.64.0 check, I thought this crate only cared about 1.65.0 so set the rust-version to that, I'll update it to 1.64.0 as otherwise that job will succeed, but by failing on an error that bypasses the actual failure that job is intended to test.

That check was actually buggy in itself, fixed in #507 — thanks for getting me to take a look 😅 The actual min supported version is 1.67 and we don't need support for anything older than that.

`cargo-semver-checks` has two Rust versions it cares about: the version it's compiled with, and the version found at runtime which will generate the rustdoc JSON. That check ensures that if the _runtime_ rustc version is too old, we fail with a better error than "failed to deserialize rustdoc JSON." It tests a binary compiled with a newer rustc (e.g. like a prebuilt binary) running on a system with an older rustc (previously 1.64, now 1.66 in #507).

So supporting 1.67 in `tame-index` should be totally fine.

> Ahh, so nvm looking how the caching works the cache key is correct and it is using the correct version, I'll continue taking a look tomorrow.

Yeah, as you probably already saw, we just build the binary once in the workflow (per-run) and reuse it in multiple jobs to skip needing to compile from scratch every time. The failure is worth looking into, though no particular rush -- tomorrow is fine. Thanks for digging into this!

Jake-Shadle:
Ok, I think this is ready to go now, doing this PR forced me to fix several bugs and add improvements to tame-index, eg https://github.com/EmbarkStudios/tame-index/pull/7, so it's now in a much better state.

Jake-Shadle:
If you'd like, I can also do a follow up PR to replace git2 with gix to reduce compile times/binary size.

obi1kenobi:
This looks great, thanks so much for digging into the issues and fixing them! I'm very excited to merge and release this 🚀 

Is my understanding correct that this PR switches our index behavior from "always git index" to "same as cargo" i.e. default-sparse unless configured otherwise?

If so, then I think all our CI tests are now over a sparse index and we have a gap in our test suite. Would you mind adding a test job that makes sure semver-checking with a git index still works as expected? For example, perhaps duplicate the `run-on-ref-slice-fork` and `run-on-ref-slice-fork-windows` jobs and run variants of them that first configure a git index?

Jake-Shadle:
The `cross-version-caching` still uses the git index as it is using 1.67 which hadn't yet stabilized sparse indices, much less made them the default, as seen by the [failure](https://github.com/obi1kenobi/cargo-semver-checks/actions/runs/5671567612/job/15369018687) before I resolved the issue in tame-index.

obi1kenobi:
Unfortunately, that is only temporary coverage — this next release is the last one that will support 1.67 — so that job will get updated to use two newer Rust versions that happen to emit different rustdoc JSON formats.

I'd really love an explicit test job for testing the git index, because bugs related to index configuration are always such a pain to triage and debug.

How can I make adding that test as easy as possible for you?

I've never added or modified index configs before, so I don't know how to assert that cargo / cargo-semver-checks are using a git index. But if you'd prefer, I can add two new jobs (Linux / Windows) that catch semver issues and have a fill-in-the-blank step that should add the configuration and assert it's being applied correctly.

Jake-Shadle:
Oh it's no problem for me to add them I just thought I would point out there is still at least one test using the git index, I'll add variations for the above tests you recommended.

obi1kenobi:
Thanks, I really appreciate it!

Also, switching to `gix` in a separate PR would be great as well!

obi1kenobi:
Very interesting, TIL about that env var!

Is `tame-index` guaranteed to read it? Is there a way to assert that it has?

I'm a bit worried about "test passes for the wrong reasons" situations, like accidentally ignoring the env var and falling back to the default index config.

Jake-Shadle:
Yes, the purpose of the `IndexUrl::crates_io` is to open the same index that cargo would given the same environment, including the [env var](https://github.com/EmbarkStudios/tame-index/blob/main/src/index/location.rs#L64-L69). I understand your concern though, but I think in this particular case it's not actually very important, the index information that is used is identical whether it used the git or sparse index for crates.io.

obi1kenobi (approved):
Thanks for all the awesome work you did. `cargo-semver-checks` and all its users are much better off for it 🚀

GitHub is showing a merge conflict at the moment, but as soon as that's resolved, I'm happy to merge this.

obi1kenobi:
Thanks again!

If you're still interested in porting `cargo-semver-checks` to `gix`, I'd happily merge a PR for that. But obviously, no pressure at all.

Jake-Shadle:
Next week ;)

obi1kenobi:
Enjoy your weekend!

Issue #120, opened 2022-09-01
`#[doc(hidden)]` result in false positives


mqudsi:
### Steps to reproduce the bug with the above code

It is possible to use `#[doc(hidden)]` to implement seemingly-breaking changes [without actually breaking semver compatibility](https://neosmart.net/blog/2022/prettysize-0-3-release-and-a-weakness-in-rusts-type-system/).

Since `cargo-semver-checks` only uses the json rustdoc output to form its analysis, the claim that `cargo-semver-checks` has no false positives should probably have this shortcoming noted in the readme. 

e.g. these two versions of the same code below are technically semver compatible:

```rust
pub struct Foo {}
pub struct Bar {}
```

```rust
#[doc(hidden)]
pub struct Foo {}
pub struct Bar {}
```

as merely hiding a type (or as is more often, a method) from the documentation doesn't break the semver contract.

or, for a more interesting case, the following:

```rust
pub struct Fo {}
```

```rust
pub struct Foo {}
// The previous release included a typo in the type name so the following is kept for backwards-compatibility:
#[doc(hidden)]
pub type Fo = Foo;
```

There is a lot more dark magic that can done that would break docs but preserve backwards compatibility. Relying solely on the docs is a great and efficient way of implementing straightforward semver breakage cases, but I don't think it's fair to claim that any tool using this approach has no false positives.

### Actual Behaviour

A breaking change is reported.

### Expected Behaviour

No breaking change should be reported (though I realize this is fundamentally not possible for this crate as it is written).
As such, the "no false positives" claim should be removed or at least the caveats thoroughly mentioned.

obi1kenobi:
Thanks for reporting.

Action items:
- [ ] Suitably update README wording on false positives. For cargo-semver-checks, false positives are bugs — we avoid them as best we can and squash them when we find them, but few things are truly 100% bug-free and we're all doing our best here.
- [ ] We currently run rustdoc with `--document-private-items`, and we should also run it with `--document-hidden-items` as well to get the additional data we need (ref: https://github.com/rust-lang/rust/pull/67875 and https://github.com/rust-lang/cargo/issues/7766 )
- [ ] For the various deleted / renamed / moved lints, we should only check items that *were not* hidden in the baseline, and match them regardless of whether they are hidden in the current version. Should also add test cases for these behaviors.
- [ ] For all other lints (I believe, let's double-check), we should filter only to items that were not, and continue to not be, hidden. Also need test cases.
[...]

PR #525, opened 2023-08-29
add item_removed_from_public_api lint

davidhewitt:
Accompaniment to https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/260

Relates #120 

This new lint adds a rule to make marking items as `#[doc(hidden)]` a breaking change unless they are also deprecated at the same time.

For #120 the existing lints probably need to be changed to require that the items be `public_api`, which should be a mechanical change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3178,8 +3178,6 @@ dependencies = [
 [[package]]
 name = "trustfall-rustdoc-adapter"
 version = "26.2.0"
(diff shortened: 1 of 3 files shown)


obi1kenobi:
Someone ended up implementing this separately in #594, which we merged. We'll have to pair on another lint sometime soon instead :)

PR #526, opened 2023-09-04
Fix error in link

c-git:
The current link gets you to the issue but not the comment as it appears it's missing the trailing 4.


obi1kenobi (approved):
Thank you!

PR #539, opened 2023-09-20
New lint for removed associated constant in trait

samuelrayment:
Adds a new lint for removing an associated constant in a trait.

Happy to add more examples if there are any.  I pretty much followed the pre-existing trait removed associated type lint, but I think I've got the hang of how Trustfall works now, so hopefully I can pick another soon.

Lint referenced here:
https://github.com/obi1kenobi/cargo-semver-checks/issues/366#issuecomment-1662820577

### DIFF
--- a/src/lints/trait_removed_associated_constant.ron
+++ b/src/lints/trait_removed_associated_constant.ron
@@ -1,0 +1,54 @@
+SemverQuery(
+    id: "trait_removed_associated_constant",
+    human_readable_name: "trait's associated constant was removed",
+    description: "A trait's associated constant was removed or renamed",
+    required_update: Major, 
+    reference_link: Some("https://doc.rust-lang.org/cargo/reference/semver.html#item-remove"),
+    query: r#"
+    {
+        CrateDiff {
+            baseline {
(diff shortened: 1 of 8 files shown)


obi1kenobi:
Nicely done! Thanks for putting this together, much appreciated!

Small nitpick: it seemed like some of the indentation in the Trustfall
query portion of the lint file was a bit off, could you take a quick look
at that?

I'm currently traveling so I might not get a chance to merge this for a
bit, but it's in good shape and I wanted to let you know ahead of time so
you don't worry if it doesn't merge immediately :)

If you're open to writing more lints, please do! You don't have to wait for
this one to merge, feel free to open multiple PRs concurrently. If you
aren't sure which lints to write next, take a look at the issue labels or
ping me and I'm happy to help.

Thanks again!

obi1kenobi (approved):

obi1kenobi:
I added a small tweak and I'm about to merge.

I made the span `@optional` because that information is only intended to help us emit a better diagnostic message. If for any reason it isn't present, we still want to flag the issue even without span data.

Issue #503, opened 2023-07-21
New lints: fn changed ABI


obi1kenobi:
Lints:
- [x] `pub fn` changed ABI to an ABI of a different name, regardless of unwind: e.g. `"C"` to `"Rust"`, or `"system-unwind"` to `"C"`
  - this changes the calling convention for the function, which is a breaking change for external code that depends on it
- [x] `pub fn` changed from an unwind-capable ABI to the same-named ABI without unwind: `"C-unwind"` to `"C"` (#689)
  - this makes unwinding become UB, which is a breaking change
- [x] `pub fn` changed from a non-unwind ABI to the unwind-capable equivalent of that same ABI, e.g. `"C"` to `"C-unwind"`
  - In principle, this could be combined with the lint above, but let's not do that. Some projects may wish to enable only one of these two lints and not the other, as unwind-capable ABIs were introduced relatively recently.
- [x] `fn` marked `#[no_mangle]` or `#[export_name]` with no importable names changed to ABI of different name
  - no importable names = it's either private or pub-in-priv -- this disambiguates the lint from the `pub fn` lints
  - we have no importable names to match on, but `#[no_mangle]` or `#[export_name]` means the function's export name must be globally unique, so match on that
  - we can use the `export_name: String` property on `Function` (implemented [here](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/237)), to simplify matching and account for both `#[no_mangle]` and `#[export_name]`

obi1kenobi:
The third checkbox was handled in #650.

qstommyshu:
Hi, I would like to try the remaining task on this issue. 
I tried to understand the pattern from the PR for the two completed task from this issue, but there are still things I don't understand, like how to check what are the fields available to a `baseline` and when and why do we choose to write `... on Function {}` at some point. Can you please give me some guidance on these please? Thanks!

obi1kenobi:
Thanks for looking into this.

Have you had a chance to check out [the schema](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/8244b6a76f985769b222f8c62783bf3cea87dc6b/src/rustdoc_schema.graphql) these queries are written against, or tried out some queries in [the rustdoc query playground](https://play.predr.ag/rustdoc)? If not, they are good to take a peek at. Note that the playground can only query one crate at a time — it can't compare two different versions at once.

[...]

PR #546, opened 2023-10-03
Adds lint for public function ABI change

samuelrayment:
Adds a new lint for extern function ABI changes.  

I've left the unwind cases for now (this can be the next lint).  If you'd like any more examples please let me know.  This hopefully closes the first bullet point on this issue: https://github.com/obi1kenobi/cargo-semver-checks/issues/503


src/lints/pub_function_changed_abi.ron
```diff
@@ -0,0 +1,58 @@
...
+                        visibility_limit @filter(op: "=", value: ["$public"]) @output
+
+                        importable_path {
+                            path @output @tag
+                        }
+                        
+                        span_: span @optional {
+                            filename @output
+                            begin_line @output
+                        }
```
obi1kenobi, line 23: Probably best if we grab the span info of the _current_ function (instead of the baseline) so we can point to the file+line in the current code where the user should look.

test_crates/pub_function_changed_abi/new/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+
+pub fn c_function_becomes_rust_implicitly() -> () {}
+
+pub extern "C" fn rust_function_becomes_c() -> () {}
+
+pub extern "C" fn implicit_rust_function_becomes_c() -> () {}
+
+pub extern "Rust" fn implicit_c_function_becomes_rust() -> () {}
+
+pub extern "Rust" fn rust_function_remains_rust() -> () {}
```
obi1kenobi, line 11: Suggestion for a couple more cases here:
- an explicitly-Rust ABI function remains Rust ABI but implicitly so, by removing the `extern "Rust"`
- the same thing in the opposite direction: adding explicit Rust ABI to an already implicitly-Rust ABI function

These cases are important guards against possible _false-positives_, which are very annoying for users and so are at least as important to protect against as the test cases for true-positives that make sure our functionality continues to work.

src/lints/pub_function_changed_abi.ron
```diff
@@ -0,0 +1,58 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "zero": 0,
+    },
+    error_message: "A publicly-visible function changed it's ABI type.",
```
obi1kenobi, line 56:
```suggestion
    error_message: "A publicly-visible function changed its ABI.",
```

obi1kenobi: Just consistency with the text below.

src/lints/pub_function_changed_abi.ron
```diff
@@ -0,0 +1,58 @@
...
+    per_result_error_template: Some("function {{join \"::\" path}} changed ABI type from {{abi_name}} to {{new_abi_name}} in {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 57:
```suggestion
    per_result_error_template: Some("{{join \"::\" path}} changed ABI from {{abi_name}} to {{new_abi_name}} in {{span_filename}}:{{span_begin_line}}"),
```
We really want to try to fit this on one line in the terminal, so brevity is everything.

src/lints/pub_function_changed_abi.ron
```diff
@@ -0,0 +1,58 @@
...
+                            path @filter(op: "=", value: ["%path"])
+                        }
+                        
+                        abi @fold @transform(op: "count")  @filter(op: "=", value: ["$zero"]) {
+                            name @filter(op: "=", value: ["%name"])
+                        }
+
+                        new_abi_: abi {
+                            name @output
+                        }
```
obi1kenobi, line 46: Functions are guaranteed to have exactly one ABI: we see this in the Trustfall schema where the ABI edge is defined `abi: FunctionAbi!` meaning exactly one neighboring vertex.

That makes this combination of query clauses equivalent to:
```suggestion
                        new_abi_: abi {
                            name @filter(op: "!=", value: ["%name"]) @output
                        }
```

In the original, we say a combination of "there is no matching ABI" + "get the actual ABI here." In the revised version I'm suggesting, we just say "the ABI that exists is not a match." This is the same thing when there's exactly one ABI, no more and no less. If there could be zero or more than one ABI, that wouldn't be equivalent and we'd need a structure like the original one here.

src/lints/pub_function_changed_abi.ron
```diff
@@ -0,0 +1,58 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "zero": 0,
```
obi1kenobi, line 54:
```suggestion
```
I think we won't need this argument after applying the suggested tweak above.

obi1kenobi (commented):
This is in very good shape, and I'm looking forward to merging it soon! Just a couple of minor tweaks and suggestions to further improve it and it's good to go — should only take just another minute of your time or so.

Thanks for putting this together! 🚀

src/lints/pub_function_changed_abi.ron
samuelrayment: Thanks for the detailed feedback it was really useful.  That should be all the suggested changes done now.

obi1kenobi: Happy to hear it! Thanks for working on this!

obi1kenobi:
> Actually I'm not sure about the convention for the lint name.
> 
> I went with: `pub_function_changed_abi` but I've noticed the existing function lints omit the `pub_`. `pub_static_missing` does exist though, is there a convention for the `pub_` prefix?

Oof, we don't have any official convention but we probably should. Let's omit `pub` to stay consistent with the function lints. `pub_static_missing` should probably become `static_missing` as well — PR welcome and much appreciated if you'd like to put one together.

obi1kenobi:
Heads up that I'm planning to publish a new release soon, probably on Sunday. I'd be happy to merge and ship this new lint, but it's also totally okay if we don't get it there in time. I just wanted to let you know my plans so you don't feel surprised to see it 👍 Thanks again for contributing!

samuelrayment:
`pub_` prefix now removed from the filenames and lint name.

obi1kenobi (approved):

obi1kenobi:
Awesome stuff, merging! Thanks for putting it together, much appreciated 🚀

PR #550, opened 2023-10-08
Move the `@optional` block to the end in the ABI lint.

obi1kenobi:
This should make the query marginally faster.

### DIFF
--- a/src/lints/function_changed_abi.ron
+++ b/src/lints/function_changed_abi.ron
@@ -16,7 +16,7 @@ SemverQuery(
                         importable_path {
                             path @output @tag
                         }
-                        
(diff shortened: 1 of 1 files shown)


obi1kenobi:
@samuelrayment this is another little trick — by evaluating the filter that checks for ABI mismatches before grabbing the optional span info, we probably end up pruning more data earlier in the processing steps. Eventually Trustfall will do this kind of optimization automatically, but it doesn't do it yet :)

Also, you might want to tweak your editor settings to remove trailing whitespace from lines. If you use vscode, the setting is `    "files.trimTrailingWhitespace": true,`

PR #566, opened 2023-10-22
Treat an empty features string as a no-op like Cargo does

### LINKED ISSUE #565: Consider accepting --features="" as a no-op like Cargo does
### Describe your use case

When a crate of mine has features, the associated CI commonly uses a feature matrix like this...

```yaml
    strategy:
      matrix:
        features:
          - '""'
          - 'std'

    # ... later used in a job step like this:
    
      - name: Type-check the program
        run: cargo check --all-targets --features=${{ matrix.features }}
```

Cargo is thankfully fine with me specifying `--features=""` as a no-op here, which is very convenient as it allows me not to duplicate CI job steps for the "no extra features case" with respect to the "with extra features" case.

Unfortunately, cargo semver-checks is not so forgiving, and running `cargo semver-checks --features=""` results in this sort of errors:

```text
     Parsing hwlocality v1.0.0 (current)
Error: `cargo metadata` exited with an error: error: failed to select a version for `hwlocality`.
    ... required by package `rustdoc v0.0.0 (/home/hadrien/Bureau/RustModernization/hwlocality/target/semver-checks/local-hwlocality-1_0_0-e48bd6b743e86099)`
versions that meet the requirements `*` are: 1.0.0

the package `rustdoc` depends on `hwlocality`, with features: `` but `hwlocality` does not have these features.


failed to select a version for `hwlocality` which could resolve this conflict
```

I'd like cargo semver-checks to accept `--features=""` as a no-op like most cargo commands do, keeping my CI workflow simple and clean.

Do nothing. Duplicating the cargo semver-checks CI job step is annoying, but not the end of the world.

### Additional Context

_No response_

HadrienG2:
Fixes #565 . ~~Draft because it depends on [an un-merged change in clap_verbosity_flag](https://github.com/clap-rs/clap-verbosity-flag/pull/73)~~ => Not anymore, since use of Default was considered harmful.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -121,7 +121,7 @@ enum SemverChecksCommands {
     CheckRelease(CheckRelease),
 }
 
-#[derive(Debug, Args)]
+#[derive(Debug, Args, Default)]
 struct CheckRelease {
     #[command(flatten, next_help_heading = "Current")]
     pub manifest: clap_cargo::Manifest,
@@ -336,6 +336,10 @@ impl From<CheckRelease> for cargo_semver_checks::Check {
             check.with_heuristically_included_features();
         }
         let mut mutual_features = value.features;
+        // Treat --features="" as a no-op like cargo does
+        if mutual_features == [""] {
+            mutual_features.clear();
+        }
         let mut current_features = value.current_features;
         let mut baseline_features = value.baseline_features;
         current_features.append(&mut mutual_features.clone());
@@ -350,4 +354,18 @@ impl From<CheckRelease> for cargo_semver_checks::Check {
 fn verify_cli() {
     use clap::CommandFactory;
     Cargo::command().debug_assert()
+}
+
+#[test]
+fn features_empty_string_is_no_op() {
+    use cargo_semver_checks::Check;
+    let without_features = CheckRelease::default();
+    let with_empty_features = CheckRelease {
+        features: vec!["".to_owned()],
+        ..CheckRelease::default()
+    };
+    assert_eq!(
+        Check::from(without_features),
+        Check::from(with_empty_features)
+    );
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -336,16 +336,19 @@ impl From<CheckRelease> for cargo_semver_checks::Check {
             check.with_heuristically_included_features();
         }
         let mut mutual_features = value.features;
-        // Treat --features="" as a no-op like cargo does
-        if mutual_features == [""] {
-            mutual_features.clear();
-        }
         let mut current_features = value.current_features;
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -336,6 +336,10 @@ impl From<CheckRelease> for cargo_semver_checks::Check {
             check.with_heuristically_included_features();
         }
         let mut mutual_features = value.features;
+        // Treat --features="" as a no-op like cargo does
+        if mutual_features == [""] {
+            mutual_features.clear();
+        }
```
obi1kenobi, line 342: Two questions:
- Does this also work in the case of `--features '' --features xyz`? This flag can be repeated and we presumably want the empty-string feature to always be a no-op.
- Should we also make this same behavior apply to `--current-features` and `--baseline-features` as well? Those flags are very similar (`--features` is shorthand for using both of those flags at once with the same inputs) so it feels like they should behave similarly in this aspect as well.

src/main.rs
```diff
@@ -121,7 +121,7 @@ enum SemverChecksCommands {
     CheckRelease(CheckRelease),
 }
 
-#[derive(Debug, Args)]
+#[derive(Debug, Args, Default)]
```
obi1kenobi, line 124: I don't think I'm comfortable deriving `Default` here, since the default settings for `cargo-semver-checks` might not be the default values of each of these settings individually. That sort of thing looks like a footgun, and I'd rather avoid it.

src/main.rs
HadrienG2: I didn't know you could specify features multiple times in an additive fashion. In this case, your suggestion does indeed make sense.

HadrienG2: Changed in the latest commit.

src/main.rs
HadrienG2: I'm a little confused here. You are using clap's `#[derive(Args)]`, which takes the default type value as the default argument value in the absence of user override, and you're not using any `#[arg(default = ...)]` to override this default, so at this point `Default` is correct. Do you fear you might add explicit `#[arg(default = ...)]` annotations someday?

If so, the latest push should alleviate your concern, but again, I'm not sure I fully understood it.

obi1kenobi: Yes — `clap`'s notion of what is default is authoritative, so even though currently a `Default` impl would happen to be the same, there's no guarantee that would remain the case. Best to avoid the opportunity to have a bug sneak in.

obi1kenobi (approved):

obi1kenobi:
Thanks!

obi1kenobi:
Queued up a new release, v0.24.1, in #567.

PR #569, opened 2023-10-23
Treat quoted empty string as an empty string

HadrienG2:
In #566 , I forgot that when cargo semver-checks is not invoked via a shell (as in, say, cargo-semver-checks-action), it is our job to treat empty quoted strings like empty strings.


obi1kenobi:
Interesting, I missed this as well. Would you mind adding a test case that exercises this new code path? Otherwise, I worry that we might miss it again in the future and break it without noticing.

HadrienG2:
I modified the unit test that I added last time to test both configurations.

obi1kenobi:
Queued up a release with this in #572. Thanks!

Issue #578, opened 2023-11-07
New lints: item that was previously public API is now `#[doc(hidden)]` i.e. no longer public API


obi1kenobi:
For example, uncommenting the `#[doc(hidden)]` line is a breaking change since the struct is no longer considered public API, even though it is public. This is because some non-public API may need to be accessible from outside its crate, for example in macro-heavy crates like `pyo3`.
```rust
// #[doc(hidden)]
pub struct Example;
```

We need lints for this kind of situation for:
- [ ] structs
- [ ] enums
- [ ] unions
- [ ] functions
- [ ] traits
- [ ] statics
- [ ] module-level constants
- [ ] `ImplOwner` (struct/enum/union) inherent associated functions / methods
- [ ] `ImplOwner` associated constants
- [ ] struct fields
- [ ] enum variants
- [ ] enum tuple variant fields
- [ ] enum struct variant fields
- [ ] trait methods
- [ ] trait associated functions / methods
- [ ] trait associated types
- [ ] trait associated constants

obi1kenobi:
Four lints to go here, and we can close this issue! 🚀

PR #587, opened 2023-11-25
implement struct_now_doc_hidden lint

u9g:
Implements a lint to catch structs that are public and have just had `#[doc(hidden)]` added to them.

checks the first checkmark on https://github.com/obi1kenobi/cargo-semver-checks/issues/578#issue-1982324505


src/lints/struct_now_doc_hidden.ron
```diff
@@ -0,0 +1,50 @@
+SemverQuery(
+    id: "struct_now_doc_hidden",
+    human_readable_name: "pub struct is now #[doc(hidden)]",
+    description: "A pub struct is now hidden from documentation effectively removing it from the crate's public api.",
```
obi1kenobi, line 4: Minor wording tweak suggestion:
```suggestion
    description: "A pub struct is now marked #[doc(hidden)] and is thus no longer part of the public API.",
```

I wanted to avoid the "effectively" since it's a bit of a weasel word and makes it sound like a "lesser" semver issue. `doc(hidden)` not being public API by convention is less known but is an hard fact of how Rust works, as the reference link you added shows.

src/lints/struct_now_doc_hidden.ron
```diff
@@ -0,0 +1,50 @@
...
+                        }
+                    }
+                }
+            }
+            current {
+                item {
+                    ... on Struct {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        struct_name: name @output
+                        struct_type @filter(op: "=", value: ["%struct_type"])
```
obi1kenobi, line 28: I wonder if we should care that the struct remains the same kind of struct (plain, tuple, or unit struct) across this change.

Normally we use this to make sure we don't trigger multiple lints, like "struct pub field removed" _and_ "struct changed kind." But whether the struct is public API or not seems orthogonal to whether it changed kind or has fields, so I'm having a hard time imagining a duplicate lint situation here.

What do you think?

src/lints/struct_now_doc_hidden.ron
```diff
@@ -0,0 +1,50 @@
...
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+    },
+    error_message: "A publicly visible struct is now hidden via #[doc(hidden)].",
+    per_result_error_template: Some("Struct {{struct_name}} in file {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 49:
```suggestion
    per_result_error_template: Some("struct {{struct_name}} in file {{span_filename}}:{{span_begin_line}}"),
```
for consistency with other lints

src/lints/struct_now_doc_hidden.ron
```diff
@@ -0,0 +1,50 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+    },
+    error_message: "A publicly visible struct is now hidden via #[doc(hidden)].",
```
obi1kenobi, line 48: Small tweak suggestion:
```suggestion
    error_message: "A pub struct is now #[doc(hidden)], removing it from the crate's public API.",
```
Many users might not know why `doc(hidden)` is relevant and being checked here, so we want to mention that the crate's public API has changed due to `doc(hidden)`.

test_outputs/struct_now_doc_hidden.output.ron
```diff
@@ -0,0 +1,46 @@
...
+                String("type_hidden_from_public_api"),
+                String("ExampleUnitStruct"),
+            ]),
+            "span_begin_line": Uint64(10),
+            "span_filename": String("src/lib.rs"),
+            "struct_name": String("ExampleUnitStruct"),
+            "struct_type": String("unit"),
+        },
+    ],
+}
```
obi1kenobi, line 46:
```suggestion
}

```
Please configure your editor to add these automatically.

test_crates/struct_now_doc_hidden/old/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+struct StayingKnown;
```
obi1kenobi, line 1: Did you intentionally leave this struct private in both `old` and `new`? It was never part of the public API in the first place.

Also btw we already have a test configuration where we lint `old` vs itself, and `new` vs itself, to make sure that we never raise lints when nothing changed.

test_crates/struct_now_doc_hidden/old/src/lib.rs
obi1kenobi, line 1: These test cases are fine, but please do try to think of some more edge cases to test for. Simple lints like this are great for practicing those skills!

One trick to help you think of test cases that is called "mutation testing." Take the correct lint you've implemented, and intentionally introduce a bug into it: for example, delete a filter clause, change a filter operator, or add an unnecessary filter or edge traversal somewhere. Then run the tests and see if they catch the bug.

You know you have good test coverage when no matter what bug you introduce, the tests always catch it. 

After some practice, you'll be able to run this process "in your head" without actually modifying the lint, and you'll be able to directly write sufficient test cases to catch most reasonable bugs. At the moment, I think there are several plausible ways to accidentally mis-implement this lint that I don't think our test suite as a whole will catch.

obi1kenobi (commented):
As always, good work! Will be happy to merge it.

I made a few minor suggestions and pointed out a few opportunities for improvement, if you're interested in making the code even better. Better tests don't seem that valuable when you're writing them, but they pay off massively by speeding up future development.

src/lints/struct_now_doc_hidden.ron
u9g: Makes perfect sense, thanks!

src/lints/struct_now_doc_hidden.ron
u9g: Yeah, you're right.

src/lints/struct_now_doc_hidden.ron
u9g: Good point, when is description shown then?

obi1kenobi: Currently, it isn't shown in the CLI, but the plan is to eventually create a feature analogous to `rustc --explain <error-code>` that shows the description.

In the meantime, we link to the lint file's URL when lints are raised and users can look up the description that way.

test_crates/struct_now_doc_hidden/old/src/lib.rs
u9g: > Did you intentionally leave this struct private in both old and new? It was never part of the public API in the first place.

No I didn't, it's been removed.

> Also btw we already have a test configuration where we lint old vs itself, and new vs itself, to make sure that we never raise lints when nothing changed.

Oh cool!

test_crates/struct_now_doc_hidden/old/src/lib.rs
u9g: Really good point, I went ahead and tried to get rid of some lines in the lint which actually ended up being optional entirely :)

In terms of tests, I added a bunch more tests that should make it a lot harder to mis-implement this rule.

test_crates/struct_now_doc_hidden/new/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+pub mod MyNestedPublicMod {
+    pub mod PublicInnerStruct {
+        // added #[doc(hidden)], however this struct is in a
+        // public mod, so it is part of the crate's public api
+        #[doc(hidden)]
+        pub struct MyStruct;
+    }
+}
+
+#[doc(hide)] // shouldn't flag, it should be #[doc(hidden)] not #[doc(hide)]
```
obi1kenobi, line 35: I don't think this is a legal attribute, and it might break in the future. Here's the list of things the `doc` attribute supports and we probably should stick to it: https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html

I'd probably pick one of the other existing `doc` commands instead of using one that doesn't exist.

u9g: good point, I replaced it with a few valid doc attribute values.

PR #593, opened 2023-12-04
implement enum_now_doc_hidden lint

u9g:
Implements a lint to catch enums that are public and have just had `#[doc(hidden)]` added to them.

checks the second checkmark on https://github.com/obi1kenobi/cargo-semver-checks/issues/578#issue-1982324505


test_crates/enum_now_doc_hidden/new/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+    // api
+    #[doc(hidden)]
+    pub enum MyEnum {
+        A,
+    }
+}
+
+pub mod MyPublicMod {
+    // added #[doc(hidden)], however this enum is in a
+    // public mod, so it is part of the crate's public api
```
obi1kenobi, line 13:
```suggestion
    // public mod, so it previously was part of the crate's public api
```

test_crates/enum_now_doc_hidden/new/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+        // api
+        #[doc(hidden)]
+        pub enum MyEnum {
+            A,
+        }
+    }
+}
+
+pub mod MyNestedPublicMod {
+    pub mod PublicInnerStruct {
```
obi1kenobi, line 33: copy-pasta?

test_crates/enum_now_doc_hidden/new/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+    pub enum MyEnum {
+        A,
+    }
+}
+
+mod MyNestedNonPublicMod {
+    pub mod PublicInnerStruct {
+        // despite adding #[doc(hidden)], this enum is in a
+        // private outer mod, so it isn't part of the crate's public
+        // api
```
obi1kenobi, line 24: As a variation of this test, you can try using a `pub mod` that is itself `#[doc(hidden)]` and then adding `#[doc(hidden)]` on the enum held inside. That shouldn't get flagged because the enum wasn't public API originally — you would have to had imported it from a hidden module.

obi1kenobi (commented):
Looks good! Just a couple of minor suggestions and it's ready to merge 🚀

test_crates/enum_now_doc_hidden/new/src/lib.rs
u9g: it's actually like this on the other test too, the name just plain doesn't make sense :/

obi1kenobi (approved):

PR #594, opened 2023-12-05
implement function_now_doc_hidden lint

u9g:
Implements a lint to catch functions that are public and have just had `#[doc(hidden)]` added to them.

checks the fourth checkmark on https://github.com/obi1kenobi/cargo-semver-checks/issues/578#issue-1982324505

### DIFF
--- a/test_crates/function_now_doc_hidden/new/src/lib.rs
+++ b/test_crates/function_now_doc_hidden/new/src/lib.rs
@@ -58,4 +58,14 @@ pub fn public_function_documented_with_string_hidden() {}
 
 #[doc(hidden)] // shouldn't flag under the `function_now_doc_hidden` lint,
                // this is a constant with a function value, not a real function
-const MY_FN: fn() = || {};
+pub const MY_FN: fn() = || {};
+
+fn my_private_fn() {
+    #[doc(hidden)] // shouldn't flag because functions aren't hoisted out of nested scopes
+    fn my_public_inner_fn_inside_my_private_fn() {}
+}
+
+pub fn my_public_fn() {
+    #[doc(hidden)] // shouldn't flag because functions aren't hoisted out of nested scopes
+    fn my_public_inner_fn_inside_my_public_fn() {}
+}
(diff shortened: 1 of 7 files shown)


test_crates/function_now_doc_hidden/new/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+#[doc(hidden)]
+fn public_function_that_goes_private() {}
+
+#[doc = "hidden"] // shouldn't flag, this is just documented with the string "hidden",
+                  // it's not actually #[doc(hidden)]
+pub fn public_function_documented_with_string_hidden() {}
+
+#[doc(hidden)] // shouldn't flag under the `function_now_doc_hidden` lint,
+               // this is a constant with a function value, not a real function
+const MY_FN: fn() = || {};
```
obi1kenobi, line 61: This wouldn't get flagged anyway because the constant is private.

```suggestion
#[doc(hidden)] // shouldn't flag under the `function_now_doc_hidden` lint,
               // this is a constant with a function value, not a real function
pub const MY_FN: fn() = || {};
```

test_crates/function_now_doc_hidden/old/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+
+pub fn my_function() {}
+
+pub fn my_function_with_types(a: i32, b: i32) -> i32 {}
+
+fn public_function_that_goes_private() {}
+
+pub fn public_function_documented_with_string_hidden() {}
+
+const MY_FN: fn() = || {};
```
obi1kenobi, line 36:
```suggestion
pub const MY_FN: fn() = || {};
```

obi1kenobi (commented):
Looks good!

Good idea on testing the constant-which-happens-to-be-a-function case. Unfortunately, the `const` wasn't public so the test didn't quite do what it said it would. I suggested a fix, otherwise this is ready to merge 🎉

obi1kenobi (approved):
Great test cases!

PR #600, opened 2023-12-07
Add draft of proposed new library API.

obi1kenobi:


obi1kenobi:
Note that the file name at the time this PR was opened is arbitrary and will not be merged as such. I added it so we get syntax highlighting and a proper review environment, without worrying about compile errors / imports / what is in which file.

src/proposed_new_lib_api.rs
```diff
@@ -0,0 +1,190 @@
...
+    /// Stop running checks and return an error.
+    FailStop,
+
+    /// Log an error but continue with other checks.
+    LogAndContinue,
+
+    /// Continue with other checks as if nothing happened.
+    /// Do not print anything to stderr.
+    ContinueSilently,
+}
```
thomaseizinger, line 190: An alternative to this design would be to model the internals of `cargo semver-checks` as a state machine that a binary or other consumer needs to manually advance. That is a bit more work to integrate but gives you greater flexibility. In other words, the library part would only offer building blocks that you can compose together with your own policy.

For example, if said state machine yields an event like:

```rust
struct WorkspaceCrateDiscovered {
	pub publish: bool,
	pub name: String
	// ...
}
```

We would also need APIs like:

```rust
pub fn load_crate_from_registry(name: &str) -> Result<...> { }

pub fn check_semver(base: &CrateData, new: &CrateData) -> Result<impl Iterator<Item = SemverViolation>, ...> {}
```

Upon encountering a `WorkspaceCrateDiscovered` event a user could then e.g. decide whether or not to continue checking the crate for semver-violations based on the `publish` field.

src/proposed_new_lib_api.rs
```diff
@@ -0,0 +1,190 @@
...
+    rev: String,
+
+    /// Path (relative to the repo root) to the `Cargo.toml` of the workspace
+    /// containing the crate to check, or to the crate's own `Cargo.toml`
+    /// if the crate isn't part of a workspace.
+    root_manifest: std::path::PathBuf,
+
+    /// If we failed to use the git repo and check out the revision.
+    /// For example, if the repo path is incorrect or the specified revision doesn't exist.
+    if_failed_to_check_out_rev: OnUnexpectedOutcome
```
obi1kenobi, line 103:
```suggestion
    if_failed_to_check_out_rev: OnUnexpectedOutcome,
```

Issue #454, opened 2023-05-22
Warn when attempting to use a baseline version higher than the crate's current version


obi1kenobi:
Users might accidentally command that a baseline version higher than the crate's current version is used, which is almost never what one actually wants. This can often happen when `--workspace` and `--baseline-version` are used together in a workspace that doesn't version all its crates in lockstep — for example [here](https://github.com/obi1kenobi/cargo-semver-checks/issues/453#issuecomment-1557312249).

We should detect that we're about to check a crate whose local version is higher than the baseline version, and emit a warning.

sladyn98:
@obi1kenobi I could attempt to give this a shot, can I get some code pointers.

obi1kenobi:
This is the bit where the current crate's version is known, and where the baseline (the version we're comparing against) is about to have its rustdoc data generated. An invalid baseline version would fail a few lines after that: https://github.com/obi1kenobi/cargo-semver-checks/blob/main/src/lib.rs#L600-L607

The `GlobalConfig` argument has a `shell_warn()` method that we use for printing warnings.

You may need to do some refactoring and plumbing to get all the data in the right places. Right now the method that generates the rustdoc JSON data doesn't know if an exact version was specified for the baseline. Perhaps it should? Or perhaps we can refactor some of the steps out from there and tweak the abstraction.

So just take a look around and figure out what you think is best. I'm happy to assist in whatever way is best — just ask. Happy to talk things over before you start implementing, or look at draft PRs, or whatever else you need.

PR #614, opened 2023-12-23
Fix false-positive in lint trait_method_unsafe_removed

yottalogical:
If a trait sealed, it should be okay to make its methods safe, since it's impossible for an external type to implement it.


obi1kenobi:
Nice catch!

It might be worth splitting this PR into two: one for adding `unsafe` and one for removing it. This is because we'll be able to merge the "unsafe added" one immediately, whereas the other one is blocked on checking whether the trait is sealed or not. This is something that's on the "to do" list but isn't implemented yet -- if you're interested in giving it a shot, I'd be happy to point you in the right direction!

yottalogical:
I'll give it a try. If you give me an overview of the changes that you think will need to be made, I can take it from there.

obi1kenobi:
On second thought, implementing sealed trait detection is quite hard, so I suggested to the rustdoc maintainers that they consider implementing it in rustdoc itself. We'll see what they say 🤞 You can find the details here: https://github.com/rust-lang/rust/issues/119280

If you're open to implementing other functionality, feel free to grab any `E-mentor` issue that looks interesting. For example, this one is a fairly easy one to get started with: https://github.com/obi1kenobi/cargo-semver-checks/issues/454

obi1kenobi:
👋 Hi! Just wanted to double-check on this PR.

I came up with a clever way to eliminate _the vast majority_ of the false-positives targeted here, without having to implement the complex check for whether traits are sealed or not. It's fairly simple: in the lint, check if the trait in question has any supertraits that have no importable paths.

The overwhelming majority of traits are sealed by inheriting a pub-in-priv supertrait, which have no importable paths. If the lint only reports traits that don't have any supertraits without importable paths, that will eliminate almost all false-positives. This is 0.1% the work of the full solution (trait sealing analysis) and produces 99.9% of the benefit.

Are you still interested in pursuing this? No worries if not, I can take it over.

yottalogical:
I can give it a try.

obi1kenobi:
No pressure if you're busy! I appreciate your help and I just wanted to make sure you still get first dibs on this if you're still interested in it :)

src/lints/trait_method_unsafe_removed.ron
```diff
@@ -17,6 +17,12 @@ SemverQuery(
                             public_api @filter(op: "=", value: ["$true"])
                         }
 
+                        supertrait @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
+                            trait {
+                                importable_path @fold @transform(op: "count") @filter(op: "=", value: ["$zero"])
+                            }
+                        }
```
obi1kenobi, line 24: One thing I just realized: I'm not 100% sure how this plays with supertraits that are from a foreign crate including `std` or `core`.

Could you throw in a test case where the trait under test has a std or core trait as a supertrait? Perhaps a supertrait like `std::convert::AsRef` could do the trick.

yottalogical: Are you talking about a sealed trait with a `std` supertrait, or an unsealed trait with a `std` supertrait (or both)?

yottalogical:
I don't have a lot of experience with Trustfall (or even GraphQL) yet, so if you think it could be written more idiomatically, I'd be interested in learning how.

src/lints/trait_method_unsafe_removed.ron
obi1kenobi: Because we currently don't do cross-crate analysis, some `supertrait` or `trait` edges might be missing if the supertrait is foreign. If the edges have lower-than-expected cardinality like that, it might spuriously satisfy the `count = 0` condition and cause false-positives.

So I think the case to look at is something like "sealed trait with `std` supertrait." But given how long I needed to figure out the exact case here, probably both the sealed and unsealed cases are worth having in the test suite.

obi1kenobi:
> I don't have a lot of experience with Trustfall (or even GraphQL) yet, so if you think it could be written more idiomatically, I'd be interested in learning how.

No, you did exactly the right thing with the query. At some point in the future we'll add syntax sugar (something like `@nonexistent`) to simplify writing the `count = 0` clause, but right now this is the only available way to write this query.

PR #615, opened 2023-12-24
Update dependencies to support rustdoc v28.

obi1kenobi:


phil-opp:
Thanks! Could we get a new crates.io release for this?

obi1kenobi:
I'm planning on making a new release today, after the new stable Rust ships.

Please consider either using stable Rust if your project allows it, or at least pinning your project's nightly Rust version. In general, `cargo-semver-checks` does _not_ promise to support nightly Rust versions with _any_ guarantees or SLA whatsoever. Nightly support is purely best-effort, as I'm sure you've read [in the FAQ](https://github.com/obi1kenobi/cargo-semver-checks#what-rust-versions-does-cargo-semver-checks-support). This is because tighter guarantees are difficult to maintain, and I have to optimize for avoiding maintainer burnout.

When using nightly, it's also important to know that some nightly versions may stop being supported even while older stable Rust versions are still supported. This is usually the case when a rustdoc format version was nightly-only and has been superseded without ever making it to a stable Rust release. For example, cargo-semver-checks v0.24 [supported](https://github.com/obi1kenobi/cargo-semver-checks/blob/v0.24.0/Cargo.toml#L18) rustdoc formats v24, v26, and v27, and did not support rustdoc v25 which was only used in several nightly releases.

I'm happy to sign a support and [reciprocal access](https://words.filippo.io/dispatches/reciprocal/) contract for projects that require stronger guarantees, such as an SLA on support for specific `nightly` versions. This can be as simple as discussing expectations over email and then setting up a GitHub sponsorship for an agreed-upon amount. If that sounds interesting, I'd love to chat about it!

phil-opp:
Thanks for all your work on this project! Great to hear that a new release is planned!

obi1kenobi:
The new release is building now and should be available shortly. It will get used by the GitHub Action automatically once the build completes.
https://github.com/obi1kenobi/cargo-semver-checks/actions/runs/7349612629

PR #619, opened 2023-12-30
Detect when `build.target` is configured, and allow overriding it from the command line

### LINKED ISSUE #579: Support semver-checking cross-compiled crates

```bash
cargo new --lib dummy
cd dummy/
mkdir .cargo
echo "[build]" >.cargo/config.toml
echo "target = \"thumbv7em-none-eabihf\"" >>.cargo/config.toml
echo '#![no_std]' >src/lib.rs
cargo semver-checks
```

### Actual Behaviour

```
Error: Could not find expected rustdoc output for `dummy`: C:\Users\Martin\dummy\target\semver-checks\local-dummy-0_1_0\target\semver-checks\target\doc/dummy.json
```

### Expected Behaviour

The program does not error internally.

#### Software version

cargo-semver-checks 0.24.2

#### Operating system

Windows 6.2.9200

#### Command-line

```bash
'C:\Users\Martin\.cargo\bin\cargo-semver-checks.exe' semver-checks --bugreport
```

#### cargo version

```
> cargo -V
cargo 1.73.0 (9c4383fb5 2023-08-26)
```

#### Compile time information

- Profile: release
- Target triple: x86_64-pc-windows-msvc
- Family: windows
- OS: windows
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2
- Host: x86_64-pc-windows-msvc

### Build Configuration
```toml
[build]
target = "thumbv7em-none-eabihf"
```

### Additional Context

The error indicates that the expected path is:
- `target\semver-checks\local-dummy-0_1_0\target\semver-checks\target\doc\dummy.json`

However, the real path due to the cross compilation seems to be:
- `target\semver-checks\local-dummy-0_1_0\target\semver-checks\target\thumbv7em-none-eabihf\doc\dummy.json`

### LINKED ISSUE #586: Failure to find json file when using `build.target` config

```console
> CARGO_BUILD_TARGET=x86_64-unknown-linux-gnu cargo semver-checks check-release
     Parsing ureq v2.9.0 (current)
Error: Could not find expected rustdoc output for `ureq`: /run/user/1000/cargo-home/target/shared/semver-checks/target/doc/ureq.json

> ls $CARGO_HOME/target/shared/semver-checks/target/x86_64-unknown-linux-gnu/doc/ureq.json
/run/user/1000/cargo-home/target/shared/semver-checks/target/x86_64-unknown-linux-gnu/doc/ureq.json
```

### Actual Behaviour

failure

### Expected Behaviour

it looks in the correct location

#### Software version

cargo-semver-checks 0.24.0

#### Operating system

Linux 6.1.58

#### Command-line

```bash
/nix/store/90q5v5asbra0mrx9hl9ixmk9pyw5akm9-cargo-semver-checks-0.24.0/bin/cargo-semver-checks semver-checks --bugreport
```

#### cargo version

```
> cargo -V
cargo 1.76.0-nightly (6790a5127 2023-11-10)
```

#### Compile time information

- Profile: release
- Target triple: x86_64-unknown-linux-gnu
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2
- Host: x86_64-unknown-linux-gnu

### Build Configuration

```
CARGO_BUILD_TARGET_DIR=/run/user/1000/cargo-home/target/shared
CARGO_HOME=/run/user/1000/cargo-home
```

### Additional Context

_No response_

Nemo157:
This allows checking the correct directory for the json files. There are two more correct approaches to this, but neither are usable currently:

 1. The most correct approach would be to read the `compiler-artifact` messages from `cargo --message-format=json`. Unfortunately as of the current nightly these don't work when using `--output-format=json`, they have an empty list of files.
 2. The next option would be to use `--out-dir` to tell cargo/rustdoc to place the json file in a known directory. As of the current nightly cargo doesn't support `--out-dir` with `cargo doc`, and attempting to pass it through to rustdoc directly results in an error because it is given more than once.

So instead, we query cargo to see if the user has configured `build.target` through either a config file or environment variables.

fixes #586, #579 

Also, because it was easy after fixing the above I added a CLI flag to allow setting which target to build for.

fixes #579 more :grin:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -38,6 +38,7 @@ pub struct Check {
     release_type: Option<ReleaseType>,
     current_feature_config: rustdoc_gen::FeatureConfig,
     baseline_feature_config: rustdoc_gen::FeatureConfig,
+    /// Which `--target` to use, if unset pass no flag
     build_target: Option<String>,
 }
 
@@ -302,6 +303,8 @@ impl Check {
         self
     }
 
+    /// Set what `--target` to build the documentation with, by default will not pass any flag
+    /// relying on the users cargo configuration.
     pub fn with_build_target(&mut self, build_target: String) -> &mut Self {
         self.build_target = Some(build_target);
         self
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -38,7 +38,7 @@ pub struct Check {
     release_type: Option<ReleaseType>,
     current_feature_config: rustdoc_gen::FeatureConfig,
     baseline_feature_config: rustdoc_gen::FeatureConfig,
-    target: Option<String>,
+    build_target: Option<String>,
 }
 
 /// The kind of release we're making.
@@ -237,7 +237,7 @@ impl Check {
             release_type: None,
             current_feature_config: rustdoc_gen::FeatureConfig::default_for_current(),
             baseline_feature_config: rustdoc_gen::FeatureConfig::default_for_baseline(),
-            target: None,
+            build_target: None,
         }
     }
 
@@ -302,8 +302,8 @@ impl Check {
         self
     }
 
-    pub fn with_target(&mut self, target: String) -> &mut Self {
-        self.target = Some(target);
+    pub fn with_build_target(&mut self, build_target: String) -> &mut Self {
+        self.build_target = Some(build_target);
         self
     }
 
@@ -426,7 +426,7 @@ impl Check {
                                 crate_type: rustdoc_gen::CrateType::Current,
                                 name: &name,
                                 feature_config: &self.current_feature_config,
-                                target: self.target.as_deref(),
+                                build_target: self.build_target.as_deref(),
                             },
                             CrateDataForRustdoc {
                                 crate_type: rustdoc_gen::CrateType::Baseline {
@@ -434,7 +434,7 @@ impl Check {
                                 },
                                 name: &name,
                                 feature_config: &self.baseline_feature_config,
-                                target: self.target.as_deref(),
+                                build_target: self.build_target.as_deref(),
                             },
                         )?;
 
@@ -494,7 +494,7 @@ impl Check {
                                     crate_type: rustdoc_gen::CrateType::Current,
                                     name: crate_name,
                                     feature_config: &self.current_feature_config,
-                                    target: self.target.as_deref(),
+                                    build_target: self.build_target.as_deref(),
                                 },
                                 CrateDataForRustdoc {
                                     crate_type: rustdoc_gen::CrateType::Baseline {
@@ -502,7 +502,7 @@ impl Check {
                                     },
                                     name: crate_name,
                                     feature_config: &self.baseline_feature_config,
-                                    target: self.target.as_deref(),
+                                    build_target: self.build_target.as_deref(),
                                 },
                             )?;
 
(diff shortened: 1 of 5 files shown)


Nemo157:
For some reason the `with_default` test is failing for me locally even before any changes, because it's building html instead of json. I hope that's just something weird with my local setup.

obi1kenobi:
> Figured it out, I use `-Zhost-config -Ztarget-applies-to-host` which causes issues with `RUSTDOCFLAGS` without a `--target` set, all of these need disabling for the tests to actually work:
> 
> ```
> env -u CARGO_UNSTABLE_BUILD_STD -u CARGO_BUILD_TARGET -u CARGO_UNSTABLE_TARGET_APPLIES_TO_HOST -u CARGO_UNSTABLE_HOST_CONFIG cargo test --test specified_target
> ```

Might you be willing to open a PR that adds a "troubleshooting" section to our CONTRIBUTING.md doc and describes this problem and its solution? I'd imagine it would be useful to other people in the future as well.

src/main.rs
```diff
@@ -257,6 +257,10 @@ struct CheckRelease {
     )]
     all_features: bool,
 
+    /// Which target to build the crate for, to check platform-specific APIs.
+    #[arg(long)]
+    target: Option<String>,
```
obi1kenobi, line 262: Thoughts on calling this field `build_target` instead (+updating all other internal names to `build_target`), then making an alias for `--target` to match cargo?

Cargo seems split: it calls it `--target`, `build.target`, and `CARGO_BUILD_TARGET`. I'm inclined to support `--target` to match it in the CLI, but I feel like internally and via the lib API `build_target` is going to be a less confusing name. For example, we don't want it to be confused with a custom target directory, which one day might also be another `Option<String>` field with `target` in the name.

obi1kenobi: It might also be nice to add an example of a build target in the doc comment, say e.g. `x86_64-unknown-linux-gnu`

src/lib.rs
```diff
@@ -299,6 +301,11 @@ impl Check {
         self
     }
 
+    pub fn with_target(&mut self, target: String) -> &mut Self {
```
obi1kenobi, line 304: nit: this is a public function so let's document what it does, what's the default value if this function isn't used, etc. It's okay if it mirrors some of the CLI's documentation.

src/lib.rs
```diff
@@ -37,6 +37,7 @@ pub struct Check {
     release_type: Option<ReleaseType>,
     current_feature_config: rustdoc_gen::FeatureConfig,
     baseline_feature_config: rustdoc_gen::FeatureConfig,
+    target: Option<String>,
```
obi1kenobi, line 40: Probably a good idea to add a field-level doc comment here, describing what the field means and what the value `None` means.

obi1kenobi (commented):
Looks pretty good, thank you! A couple of nitpicks and minor suggestions, and I'd be happy to merge this.

obi1kenobi:
Nicely done, love how carefully you put all this together! 🚀 Thank you so much!

If you might be interested in making more PRs and want some ideas, give me a ping anytime. I've written up some of my thinking as GitHub issues, but not everything is in there yet.

PR #622, opened 2024-01-06
Add test crate to test async fn and impl Future

### LINKED ISSUE #613: Switching between `async fn` and `fn -> impl Future` is not breaking
It's not a breaking change to switch back and forth between `async fn` and `-> impl Future` phrasing: https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html#can-i-mix-async-fn-and-impl-trait

We need test cases to ensure this does not trigger lints in:
- [x] functions
- [x] functions in inherent `impl` blocks ("associated functions" / "methods")
- [ ] trait functions (issue: #624)

If adding such test cases makes us discover false-positives, we should also fix them :)

jw013:
Closes #613

### DIFF
--- a/test_crates/async_impl_future_equivalence/new/Cargo.toml
+++ b/test_crates/async_impl_future_equivalence/new/Cargo.toml
@@ -1,0 +1,7 @@
+[package]
+publish = false
+name = "async_impl_future_equivalence"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
--- a/test_crates/async_impl_future_equivalence/new/src/lib.rs
+++ b/test_crates/async_impl_future_equivalence/new/src/lib.rs
@@ -1,0 +1,27 @@
+//! async fn and return position impl Future are equivalent.
+//! Switching from one to the other should NOT generate any lints
+
+use std::future::Future;
+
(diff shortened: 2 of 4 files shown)


obi1kenobi:
Thanks for the PR, this looks great! I love that you considered all the edge cases of converting between `impl Future` and `async fn`.

CI seems to be having issues on the older Rust versions that don't support `async fn` in traits. This is annoying, and not your fault.

I'd say in the interest of getting this merged, for now:
- comment out the entire `pub trait Trait` portion of the test code
- put in a TODO saying that we should uncomment it after Rust 1.75 is the oldest Rust version we support
- open an issue for uncommenting that code, and link to it from the comment, from this PR, and from #613.

This is probably going to be easiest way to merge this, and the best effort-to-benefit tradeoff.

jw013:
Thanks!, I've created #624 for reenabling the trait tests in the future and added a commit to disable them in the present.

obi1kenobi (approved):

obi1kenobi:
Merging, thanks so much! If you're open to contributing more and are looking for ideas, lmk what you'd find most interesting and I'd be happy to help! 👀

Issue #298, opened 2023-01-20
Displayed timing is unclear and only accounts for rule-checking time


obi1kenobi:
### Describe your use case

Our timing code predates the ability to generate rustdoc within the tool itself, and as such has some unintuitive behaviors:
- It doesn't time rustdoc builds (either baseline or current).
- Included time totals don't include rustdoc build or parse times, which are nontrivial for large crates.

This can be confusing from a user perspective, since running the tool under `time` will report noticeably different timing data than the tool's own printouts.

Since rustdoc build/parse timing information is not broken out and reported, it also makes it difficult for users to know how time is spent and how to attempt to speed up the tool's runtime.

### Describe the solution you'd like

In verbose mode:
- Report baseline and current rustdoc generation times.
- Report rustdoc file parsing time.

In general:
- Include rustdoc generation and parsing time in reported totals.

### Additional Context

Based on this experience report: https://github.com/obi1kenobi/cargo-semver-checks/discussions/296

AntoniosBarotsis:
Hello, I'd be interested to look into this!

I will be rather busy for the next week/week and a half but I can definitely implement this right after *seems simple enough*.

I was wondering how should I reach out if I need any help or clarifications on the issue? Should I just reply here? Thanks in advance!

> If anyone else wants to work on this before me, feel free to do so! I don't want to hog the issue

obi1kenobi:
Hey! Awesome, thanks for looking into it!

Replying here is fine. I'm also on the rust-lang Zulip if you prefer that: https://rust-lang.zulipchat.com/

jw013:
Current output:

```text
     Parsing <crate> ... (current)
     Parsing <crate> ... (baseline)
    Checking <crate> ... 
#if VERBOSE
    Starting ...
        PASS [  S.SSSs] ...
        FAIL [  S.SSSs] ...
#endif
   Completed [   S.SSSs] ...
#if CHECKS_FAILED
--- failure reports ...
       Final [   S.SSSs] semver requires ...
#endif
```

Proposed changes (The `+` and `*` are just markers, not intended as output.)

```text
     Parsing <crate> ... (current)
     Parsing <crate> ... (baseline)
+     Parsed [   S.SSSs]
    Checking <crate> ... 
#if VERBOSE
*   Starting ... (threads: #)
        PASS [   S.SSSs] ...
        FAIL [   S.SSSs] ...
#endif
   Completed [   S.SSSs] ...
[...]

PR #625, opened 2024-01-08
Parallel lints

### LINKED ISSUE #621: Run lints in parallel using `rayon`
At a high level, `cargo-semver-checks` is just a for-loop over a list of lints. Each lint is implemented as a query — for example, "find all public functions that used to exist in the previous version, but no longer exist in the new version."

Currently, all such lint queries run sequentially. However, they are all read-only and logically independent from each other — so they could also run in parallel! Instead of a for-loop over a list of lints, we'd have a parallel-for-loop — and a ~10x speedup of checking time on most modern hardware. (The final wall-clock speedup will be somewhat smaller due to [Amdahl's Law](https://en.wikipedia.org/wiki/Amdahl%27s_law) as we still have to generate rustdoc JSON, and that process is not affected by this speedup.)

**Expected result**
Use the `rayon` library to run lint queries over a given crate in parallel. Refactor the way `cargo-semver-checks` passes data and reports progress to the user as necessary to accommodate parallel lint runs. In particular, ensure that reported timing information accurately communicates how much time is spent generating rustdoc data (serial, blocking) and how much time is spent actually running each lint (after this change, parallel across lints) — this is covered by #298.

jw013:
Closes #621.

The rayon implementation lacks the per query "Running {query}\r" dynamic messages as I couldn't think of any sensible way to do that with multiple threads.

### DIFF
--- a/src/check_release.rs
+++ b/src/check_release.rs
@@ -1,4 +1,4 @@
-use std::{collections::BTreeMap, env, io::Write, iter::Peekable, sync::Arc, time::Instant};
+use std::{collections::BTreeMap, sync::Arc, time::Instant};
 
 use anyhow::Context;
 use clap::crate_version;
@@ -6,29 +6,13 @@ use itertools::Itertools;
 use rayon::prelude::*;
 use termcolor::Color;
 use termcolor_output::{colored, colored_ln};
-use trustfall::{FieldValue, TransparentValue};
+use trustfall::TransparentValue;
 use trustfall_rustdoc::{VersionedCrate, VersionedIndexedCrate, VersionedRustdocAdapter};
 
 use crate::{
     query::{ActualSemverUpdate, RequiredSemverUpdate, SemverQuery},
     CrateReport, GlobalConfig, ReleaseType,
 };
-
-type QueryResultItem = BTreeMap<Arc<str>, FieldValue>;
-
-struct QueryWithResults<'a> {
-    name: &'a str,
-    results: Peekable<Box<dyn Iterator<Item = QueryResultItem> + 'a>>,
-}
-
-impl<'a> QueryWithResults<'a> {
-    fn new(
-        name: &'a str,
-        results: Peekable<Box<dyn Iterator<Item = QueryResultItem> + 'a>>,
-    ) -> Self {
-        Self { name, results }
-    }
-}
 
 fn classify_semver_version_change(
     current_version: Option<&str>,
--- a/src/check_release.rs
+++ b/src/check_release.rs
@@ -119,14 +119,26 @@ pub(super) fn run_check_release(
     )?;
     config
         .log_verbose(|config| {
-            config.shell_status(
-                "Starting",
-                format_args!(
-                    "{} checks, {} unnecessary",
-                    queries_to_run.len(),
-                    skipped_queries
-                ),
-            )
+            let current_num_threads = rayon::current_num_threads();
+            if current_num_threads == 1 {
+                config.shell_status(
+                    "Starting",
+                    format_args!(
+                        "{} checks, {} unnecessary",
+                        queries_to_run.len(),
+                        skipped_queries
+                    ),
+                )
+            } else {
+                config.shell_status(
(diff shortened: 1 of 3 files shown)


obi1kenobi:
Thanks, this is great!

Any chance we can tackle this portion of the linked issue in the process as well?

> In particular, ensure that reported timing information accurately communicates how much time is spent generating rustdoc data (serial, blocking) and how much time is spent actually running each lint (after this change, parallel across lints) — this is covered by https://github.com/obi1kenobi/cargo-semver-checks/issues/298.

obi1kenobi:
Let's go for a new branch, just to make it easy to review.

If you'd like, feel free to put together a draft of what the output could
look like before taking the time to implement it. There are some edge cases
to consider (e.g. what if there are multiple crates to check in the
workspace, etc. with other CLI flags) that could be good to work out "on
paper" before investing into an implementation.

Similar edge cases come up when reporting parallel timing as well, which is
why I suggested tackling these two issues together.

src/check_release.rs
```diff
@@ -135,56 +119,58 @@ pub(super) fn run_check_release(
...
+        })
+        .collect::<anyhow::Result<Vec<_>>>()?;
 
-    for (query_id, semver_query) in queries_to_run.iter().copied() {
-        let category = match semver_query.required_update {
-            RequiredSemverUpdate::Major => "major",
-            RequiredSemverUpdate::Minor => "minor",
-        };
+    // print PASS/FAIL summaries
+    all_results.retain(|(query_id, required_update, time_to_decide, results)| {
```
obi1kenobi, line 166: nit: I'm not a huge fan of this `.retain()`, for two reasons. It means that we rely on `retain()` to do side-effects, which relies on `retain()` running in a particular way (sequential, synchronous, etc.) and that seems kind of unintuitive and convoluted. We're also mutating `all_results` in a way that explicitly makes its contents no longer match the description `all_results`, which might come back to bite us in a variety of ways in the future.

I think I'd rather have an explicit for-loop here, for the sake of readability and intuitiveness. We can _probably_ avoid a `.clone()` of `results` by capturing a slice in `results_with_errors` instead, but if that's wrong and we can't avoid cloning, that's okay too. The "found errors" path is the rarer case, and users will be happy to wait an extra few ms worth of `memcpy` to know that breaking changes were found.

obi1kenobi (commented):
One nitpick, otherwise this is good to go! 🚀

Thanks for putting it together!

obi1kenobi:
Based on an unscientific "I compared two CI runs of one of one of our beefier tests," GitHub Actions runners have 4 cores and the speedup is in the neighborhood of 2-2.5x. This would make sense if they have 2 physical cores with hyperthreading enabled.

obi1kenobi (approved):

obi1kenobi:
Awesome! 🚀 

I'm planning to post about this PR on social media, and I'd like to give you a shout-out if you're okay with that (no pressure if not!). Lmk if you have a Mastodon / Twitter / bsky I should tag in the post?

jw013:
> Awesome! 🚀
> 
> I'm planning to post about this PR on social media, and I'd like to give you a shout-out if you're okay with that (no pressure if not!). Lmk if you have a Mastodon / Twitter / bsky I should tag in the post?

That's fine with me. Outside of lurking occasionally on Mastodon I don't really use social media.

obi1kenobi:
Happy to tag your Mastodon if you'd like (just let me know what it is). Again, no pressure!

obi1kenobi:
Also, if you're interested in contributing more things, happy to help however I can! There's a mix of issues on the issue tracker, but I haven't done an amazing job of curating them thus far so feel free to ping me with any questions or ask for suggestions.

jw013:
Mastodon is how I found out about this project - I only use it to follow people who post interesting things about Rust and other topics, so probably not worth mentioning. 

I'm happy just picking through open issues for now - what I have seen so far has been well documented and fairly easy to understand so I definitely appreciate that plus the additional feedback. Thanks!

PR #626, opened 2024-01-09
update repr-related reference links

fprasx:
Looks like an `enum_repr_transparent_removed` is missing (it's there for `struct`). I'll add that on in another PR.

Reference: https://doc.rust-lang.org/cargo/reference/semver.html#repr-transparent-remove

### DIFF
--- a/src/lints/enum_repr_c_removed.ron
+++ b/src/lints/enum_repr_c_removed.ron
@@ -5,9 +5,7 @@ SemverQuery(
     reference: Some("An enum that used to be repr(C) is no longer repr(C). This can cause its memory layout to change, breaking FFI use cases."),
     required_update: Major,
 
-    // TODO: Change the reference link to point to the cargo semver reference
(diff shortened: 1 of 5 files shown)


obi1kenobi (approved):
I didn't know `repr(transparent)` was allowed on enums! Apparently it is, but it's a compile error if the enum has more than one variant.

fprasx:
Makes sense - I suppose you could also define it as meaning each variant has `repr(transparent)` 🤔

obi1kenobi:
That's subtly not quite the same thing. Consider the enum discriminant — it ends up being like another field, and is not a zero-sized type if the enum has more than one variant. We don't allow having more than one field that isn't a 1-ZST in structs, so we wouldn't allow it for enums either.

fprasx:
Oh good point - for some reason I as assuming they would all be "transparent to the same type" in my head, but even still the discriminant messes things up (especially because it may or may not be present due to niche optimization!).

PR #627, opened 2024-01-09
implement enum_repr_transparent_removed lint

fprasx:
Still getting used to query syntax, so there might be some subtle bugs in the query. I mostly reused the test suite from `struct_repr_transparent_removed`.

My understanding is that:
```graphql
variant @fold @transform(op: "count") @filter(op: "=", value: ["$one"]) {
    transparent_variant_name: name @output
    field @fold @transform(op: "count") @filter(op: "=", value: ["$one"]) {
        raw_type @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
            ... on ResolvedPathType {
                name @filter(op: "one_of", value: ["$phantom_data"])
            }
        }
    }
}
```
translates to:
```
[take enums]: where
    there is a exactly one variant: where
        it contains exactly on field: where
            it contains exactly one field: where
                the number of rawtypes in this field that are not PhantomData is 1 (could be 0 or 1)
```

### DIFF
--- a/src/lints/enum_repr_transparent_removed.ron
+++ b/src/lints/enum_repr_transparent_removed.ron
@@ -1,0 +1,108 @@
+SemverQuery(
+    id: "enum_repr_transparent_removed",
+    human_readable_name: "enum repr(transparent) removed",
+    description: "A enum is no longer repr(transparent).",
+    reference: Some(r#"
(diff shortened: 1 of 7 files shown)


fprasx:
Thanks! I mostly just copied over the lint for structs and then tweaked it haha, glad it turned out ok 😅

PR #647, opened 2024-01-29
Only run test with_env_var and with_flag on x86_64

Xeonacid:


obi1kenobi (approved):
Good catch, thanks!

obi1kenobi:
@Xeonacid assuming you aren't running x86_64 yourself, would you care to add a couple of equivalent tests for your arch in a new PR?

Otherwise, folks on other architectures might not realize if they've accidentally broken this functionality, since they'd never run the tests for it locally.

Xeonacid:
> @Xeonacid assuming you aren't running x86_64 yourself, would you care to add a couple of equivalent tests for your arch in a new PR?

Sure! I will do that these days.

obi1kenobi:
Awesome, thank you so much!

PR #655, opened 2024-02-11
feat: Trailing comma error message

devanbenz:
- add error message for missing trailing commas in add_lints macro


src/query.rs
```diff
@@ -463,6 +463,9 @@ macro_rules! add_lints {
             ]
         }
     }
+    ($($name:ident),*) => {
+        compile_error!("A trailing comma is required after each lint identifier in `add_lints!` macro.");
```
obi1kenobi, line 467:
```suggestion
        compile_error!("Please add a trailing comma after each lint identifier. This ensures our scripts like 'make_new_lint.sh' can safely edit invocations of this macro as needed.");
```

obi1kenobi (approved):

PR #659, opened 2024-02-12
feat: New lint trait_no_longer_object_safe 

devanbenz:
- Adds a new lint for traits to see if they are still object safe


devanbenz:
@obi1kenobi Having a few issues with this PR. I'm noticing that the test is failing with the following error: 
```
thread 'query::tests_lints::trait_no_longer_object_safe' panicked at src/query.rs:342:13:
The query produced a non-empty output when it compared two crates with the same rustdoc.
Query trait_no_longer_object_safe produced incorrect output (./src/lints/trait_no_longer_object_safe.ron).

Differences (-expected (empty)|+actual (trait_now_doc_hidden/new)):
-{}
+{
+    "./test_crates/trait_now_doc_hidden/": [
+        {
+            "name": String("AliasedAsDocHidden"),
+            "path": List([
+                String("trait_now_doc_hidden"),
+                String("AliasedAsDocHidden"),
+            ]),
+            "span_begin_line": Uint64(44),
+        },
+        {
+            "name": String("PublicTraitHiddenVariant"),
+            "path": List([
+                String("trait_now_doc_hidden"),
+                String("PublicTraitHiddenVariant"),
+            ]),
+            "span_begin_line": Uint64(49),
+        },
+        {
+            "name": String("PublicTraitDocumentedWithStringHidden"),
+            "path": List([
+                String("trait_now_doc_hidden"),
+                String("PublicTraitDocumentedWithStringHidden"),
+            ]),
+            "span_begin_line": Uint64(62),
+        },
+    ],
+}
```
It looks like its picking up a query from a different lint. When I go through and try to debug what exactly is happening I see that https://github.com/devanbenz/cargo-semver-checks/blob/ea9a307765ea5d80b12bf8ada784019ad3f23cab/src/query.rs#L140 is outputting trait_no_longer_object_safe in the test crates returned array which looks okay. Looking here https://github.com/devanbenz/cargo-semver-checks/blob/ea9a307765ea5d80b12bf8ada784019ad3f23cab/src/query.rs#L325 getting the output of crate_pair_path I'm seeing 

```
"./test_crates/function_changed_abi/"
"./test_crates/function_changed_abi/"
"./test_crates/trait_now_doc_hidden/"
```

Which is the incorrect crate path 🤔

src/lints/trait_no_longer_object_safe.ron
```diff
@@ -0,0 +1,51 @@
+SemverQuery(
+    id: "trait_no_longer_object_safe",
+    human_readable_name: "trait no longer object safe",
+    description: "A trait is no longer object safe",
```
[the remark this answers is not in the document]
devanbenz: yep! I plan on going through and cleaning up the messages :)

obi1kenobi:
This is the key line in that test output:
> The query produced a non-empty output when it compared two crates with the same rustdoc.

This means that comparing a crate to itself (i.e. no changes whatsoever) triggered a lint. This is always a false-positive, since there was no source code change at all.

Two ideas for follow-up PRs, if you're open to them:
- Add a note to the CONTRIBUTING doc that the test cases run all lints on all test crates, so it's perfectly normal to see lint output from other crates if those crates have code that triggers your new lint. Also add a note describing the linting of crates with an identical copy of themselves, to catch false-positives.
- Update the test output in that test, to do a better job explaining why a lint reporting issues when comparing some rustdoc to itself is a problem. In retrospect, there's no reasonable way you could have understood what that test was trying to do based on that message, so it's a clear and unnecessary onboarding papercut.

devanbenz:
> Two ideas for follow-up PRs, if you're open to them:
> 
> * Add a note to the CONTRIBUTING doc that the test cases run all lints on all test crates, so it's perfectly normal to see lint output from other crates if those crates have code that triggers your new lint. Also add a note describing the linting of crates with an identical copy of themselves, to catch false-positives.
> * Update the test output in that test, to do a better job explaining why a lint reporting issues when comparing some rustdoc to itself is a problem. In retrospect, there's no reasonable way you could have understood what that test was trying to do based on that message, so it's a clear and unnecessary onboarding papercut.

Amazing - was able to resolve. Going to take note of those pr ideas you indicated :D

build.rs
```diff
@@ -0,0 +1,14 @@
...
+        .arg("--version")
+        .output()
+        .expect("Failed to execute rustc");
+
+    let rustc_version = String::from_utf8_lossy(&rustc_output.stdout);
+
+    if rustc_version.contains("nightly") {
+        println!("cargo:rustc-cfg=nightly");
+    }
+}
```
obi1kenobi, line 14: Let's not do this, the complexity isn't worth it.

Nightly detection is often finicky and causes issues in downstream crates (remember, `cargo-semver-checks` is a library too — that's how `release-plz` integrates it) and is generally not recommended.

This is taking our test suite's inability to specify different expected results for different Rust versions, and pushing the complexity onto our users. It's not a good idea.

obi1kenobi:
Let's do this:
- Split the test crate and expected nightly test output out of this PR and into a separate PR. Having seen that it would pass right now, we'll leave that one as a draft and merge it in the future.
- That will leave the new lint unconditionally included and available on all Rust versions here in this PR. That we can merge today and ship to our users who run nightly or beta Rust right now.

build.rs
devanbenz: I see! Okay sounds good 👍

src/lints/trait_no_longer_object_safe.ron
```diff
@@ -0,0 +1,51 @@
...
+    query: r#"
+    {
+        CrateDiff {
+            baseline {
+                item {
+                    ... on Trait {
+                        visibility_limit @filter(op: "=", value: ["$public"]) @output
+                        object_safe @filter(op: "=", value: ["$true"])
+
+
```
obi1kenobi, line 16:
```suggestion

```

src/lints/trait_no_longer_object_safe.ron
```diff
@@ -0,0 +1,51 @@
...
+                            begin_line @output
+                        }
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true
```
obi1kenobi, line 47:
```suggestion
        "true": true,
```

obi1kenobi (commented):
Will merge after applying a couple of minor nit fixes + after you've had a chance to open the draft PR with the test crate and test outputs.

devanbenz:
@obi1kenobi https://github.com/obi1kenobi/cargo-semver-checks/pull/661 opened a draft PR with the tests

PR #666, opened 2024-02-18
Add lints for repr(packed) being added or removed.

### LINKED ISSUE #632: New lints: `repr(packed)` added or removed on a struct or union
Both adding and removing `repr(packed)` is a breaking change when it happens on either structs or unions:
https://doc.rust-lang.org/cargo/reference/semver.html#repr-packed-add
https://doc.rust-lang.org/cargo/reference/semver.html#repr-packed-remove

Adding and removing are definitely going to need separate lints. Whether structs and unions each have their own lints (so, total of 4 lints across all combinations) or we combine them together (so, 2 lints total: `repr(packed)` added, `repr(packed)` removed) is up to the implementer.

jw013:
Resolves #632

### DIFF
--- a/src/lints/repr_packed_added.ron
+++ b/src/lints/repr_packed_added.ron
@@ -7,32 +7,10 @@ SemverQuery(
     query: r#"
     {
         CrateDiff {
-            baseline {
-                item {
-                    ... on ImplOwner {
-                        type: __typename @filter(op: "one_of", value: ["$types"]) @output @tag
-                        visibility_limit @filter(op: "=", value: ["$public"])
-
-                        attribute @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
-                            content {
-                                base @filter(op: "=", value: ["$repr"])
-                                argument {
-                                    base @filter(op: "=", value: ["$packed"])
-                                }
-                            }
-                        }
-
-                        importable_path {
-                            path @tag @output
-                            public_api @filter(op: "=", value: ["$true"])
-                        }
-                    }
-                }
-            }
             current {
                 item {
                     ... on ImplOwner {
-                        __typename @filter(op: "=", value: ["%type"])
+                        type: __typename @filter(op: "one_of", value: ["$types"]) @output @tag
                         visibility_limit @filter(op: "=", value: ["$public"])
                         name @output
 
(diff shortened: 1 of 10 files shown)


src/lints/repr_packed_added.ron
```diff
@@ -0,0 +1,72 @@
...
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        span_: span @optional {
+                            filename @output
+                            begin_line @output
+                        }
+                    }
+                }
+            }
```
obi1kenobi, line 59: Minor perf nit here that you couldn't possibly have known about, because I need to write more docs 😅

Trustfall has no prior knowledge of how often `repr(packed)` structs are, so it will run the `baseline` and `current` legs of this query in lexical order ("in the order they appear in the query"). But the query will run faster if we do the more-rare case first (since it will prune more aggressively and do less work). We as Rustaceans know that most structs aren't `repr(packed)`. The `current` block looks for `repr(packed)` structs, so doing it first will make the query run faster.

src/lints/repr_packed_added.ron
```diff
@@ -0,0 +1,72 @@
...
+    }"#,
+    arguments: {
+        "public": "public",
+        "packed": "packed",
+        "repr": "repr",
+        "true": true,
+        "types": ["Struct", "Union"],
+        "zero": 0,
+    },
+    error_message: "repr(packed) was added to a struct or union. This can break code that references individual fields.",
```
obi1kenobi, line 70:
```suggestion
    error_message: "repr(packed) was added to a type. This can break code that references individual fields.",
```
just to make sure that users don't get confused about "struct or union" — the lovely templated error message you added will tell them exactly which flavor of type has the problem

src/lints/repr_packed_removed.ron
```diff
@@ -0,0 +1,72 @@
+SemverQuery(
+    id: "repr_packed_removed",
+    human_readable_name: "repr(packed) removed",
+    description: "A struct or union that used to be repr(packed) is no longer repr(packed).",
```
obi1kenobi, line 4:
```suggestion
    description: "A struct or union that used to be #[repr(packed)] is no longer #[repr(packed)].",
```
for consistency with the other lint's description

src/lints/repr_packed_removed.ron
```diff
@@ -0,0 +1,72 @@
...
+    }"#,
+    arguments: {
+        "public": "public",
+        "packed": "packed",
+        "repr": "repr",
+        "true": true,
+        "types": ["Struct", "Union"],
+        "zero": 0,
+    },
+    error_message: "repr(packed) was removed from a struct or union. This can break code that depends on the type's alignment or layout.",
```
obi1kenobi, line 70:
```suggestion
    error_message: "repr(packed) was removed from a type. This can break code that depends on the type's alignment or layout.",
```
same as in the other lint

obi1kenobi (commented):
Wonderfully done! Just a couple of tiny nitpicks fixable in 2min and I'll be thrilled to merge and release this promptly! 🚀

src/lints/repr_packed_added.ron
obi1kenobi: In principle we could have added this knowledge ("`repr(packed)` structs are rare, prioritize that") into the adapter, but we haven't yet and it's likely not worth the effort since this isn't something that comes up often.

obi1kenobi (approved):

obi1kenobi:
Nicely done, thank you!

obi1kenobi:
Cutting a new release with this now: #669

PR #667, opened 2024-02-18
Minor QOL update to speed up reruns of regenerate_test_rustdocs script

jw013:
- By default, check timestamps and skip generation if not needed
- Allow explicit test_crates as arguments to force regeneration

On my ancient Linux laptop, a full regeneration takes

```
$ time ./scripts/regenerate_test_rustdocs.sh test_crates/*
...
real	1m33.753s
user	1m18.191s
sys	0m13.961s
```

while a run where regenerate does nothing takes

```
$ time ./scripts/regenerate_test_rustdocs.sh
...
real	0m0.183s
user	0m0.125s
sys	0m0.055s
```

Not earth shattering, but I make a lot of typos when making the test crates and this makes my life a little easier, figured I'd share.


obi1kenobi:
Wow, very cool! Thanks for making this!

Issue #634, opened 2024-01-13
Consider switching to `insta` for snapshot testing


obi1kenobi:
This project relies heavily on snapshot testing: we write test crates and lint them, then ensure that the outputs are both expected and haven't changed since the last run.

When `cargo-semver-checks` came into existence, `insta` wasn't nearly as polished as it's become since. These days, it's feature-rich and fairly commonly used in the ecosystem.

Our own hand-rolled snapshot testing can be a bit annoying to manage: one has to manually edit snapshot files, and observe proper item order since they are order-sensitive. However, it also allows us some conveniences: we can leave comments inside snapshot files (e.g. look at the removed lines in #631 for a useful snapshot comment), and we have tight control over how information is stored and where.

Can an `insta`-based workflow replicate the conveniences we care about, or offer other conveniences that make switching to it compelling?

To close this issue, we need to:
- [ ] figure out what a good `insta`-based workflow would look like
- [ ] consider whether it's better than what we have right now -- and if so
  - [ ] replace our snapshot testing with `insta`
  - [ ] update our [CONTRIBUTING doc](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md) with the new workflow, making sure to add guidance on how to install `cargo-insta` and how to use `insta` and `cargo insta` productively

PR #677, opened 2024-03-01
Fix description and tests for inherent_method_must_use_added

### LINKED ISSUE #674: Description and tests for `inherent_method_must_use_added` is not complete

Master branch

### Actual Behaviour

Description is `An inherent method has been marked with #[must_use].`

and the tests only test methods (`fn` with `self`)

### Expected Behaviour

Follow the same pattern as the other inherent method lints:

Description to change to `A method or associated fn has been marked with #[must_use].`

And add associated fn tests for this lint.

master

### Build Configuration

_No response_

### Additional Context

_No response_

pksunkara:
Fixes #674


src/lints/inherent_method_must_use_added.ron
```diff
@@ -1,7 +1,7 @@
 SemverQuery(
     id: "inherent_method_must_use_added",
     human_readable_name: "inherent method #[must_use] added",
-    description: "An inherent method has been marked with #[must_use].",
+    description: "A method or associated fn has been marked with #[must_use].",
```
obi1kenobi, line 4:
```suggestion
    description: "An inherent method or associated fn has been marked #[must_use].",
```
"Inherent" is a load bearing word here, it means the `impl` was on the type itself and not on a trait implemented for that type. This lint only looks at associated functions (including methods) of inherent impls.

obi1kenobi (approved):
Very nice, thank you! 🚀

I really appreciate your work on improving the lints! I'd love to merge any more improvements you find. If you might be interested in some opportunities for improvement deeper than the lints themselves, I'd be happy to suggest some ideas as well.

pksunkara:
> All functions defined within an impl block are called associated functions because they’re associated with the type named after the impl. 

From my understanding, by default, associated fn of a type means the ones within inherent impl.

pksunkara:
> If you might be interested in some opportunities for improvement deeper than the lints themselves

Honestly, I am really itching to revamp the whole test system so that it can be used to auto-generate documentation. But I have been holding back because I know it needs to be discussed first.

obi1kenobi:
> > All functions defined within an impl block are called associated functions because they’re associated with the type named after the impl.
> 
> From my understanding, by default, associated fn of a type means the ones within inherent impl.

I'm not sure — and even if that is the case, it's best to err on the side of clarity. It's better to avoid causing misunderstandings than to be technically correct while confusing users.

> Honestly, I am really itching to revamp the whole test system so that it can be used to auto-generate documentation. But I have been holding back because I know it needs to be discussed first.

Neat! I agree it could be better, in a variety of ways. Would you care to open an issue and write down your thoughts on it? I'd love to hear what you'd change, how you'd change it, and what auto-generated documentation would be created as a result!

I'm particularly curious about:
- Whether we can make the revamp be a sequence of small, smooth steps, each of which is easy to review and merge, and leaves the overall system better off than before. This would massively de-risk the transition.
- Whether we can easily add testing of the error message templating, since it's had issues in the past — some of which you've flagged as well.
- Whether we should try to adopt a tool like `insta` (#634) for managing our snapshot tests, or if it would be too slow or otherwise add too much friction.

Another key thing I strongly care about is making sure to keep the `CONTRIBUTING.md` doc updated to match how development happens, since otherwise our project wouldn't be very contributor-friendly.

PR #678, opened 2024-03-01
Replace per-type `repr(C)` lints with a single lint for all of struct/enum/union

pksunkara:


src/lints/repr_c_removed.ron
```diff
@@ -67,6 +69,6 @@ SemverQuery(
         "true": true,
         "zero": 0,
     },
-    error_message: "repr(C) was removed from an enum. This can cause its memory layout to change, breaking FFI use cases.",
-    per_result_error_template: Some("enum {{name}} in {{span_filename}}:{{span_begin_line}}"),
+    error_message: "repr(C) was removed from a type. This can cause its memory layout to change, breaking FFI use cases.",
+    per_result_error_template: Some("{{owner_type}} {{name}} in {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 73:
```suggestion
    per_result_error_template: Some("{{lowercase owner_type}} {{name}} in {{span_filename}}:{{span_begin_line}}"),
```
so that it prints `struct` instead of `Struct` etc.

obi1kenobi (approved):

obi1kenobi:
Thanks!

src/lints/repr_c_removed.ron
pksunkara: We don't do this in `src/lints/type_marked_deprecated.ron` though.

obi1kenobi: Nice catch, we should! Would you care to open a PR?

This is a good data point for a possible improvement to our current test system btw.

Issue #680, opened 2024-03-02
Running scripts on systems with bash version older than v4 causes errors


obi1kenobi:
Running `./scripts/regenerate_test_rustdocs.sh` also gives the error,
`cargo-semver-checks/scripts/regenerate_test_rustdocs.sh: line 23: shopt: globstar: invalid shell option name`

_Originally posted by @vasucp1207 in https://github.com/obi1kenobi/cargo-semver-checks/pull/679#discussion_r1509730892_
            
The `globstar` option was introduced by Bash 4, released 15 years ago. Unfortunately, some systems still ship with only Bash 3 and will have this error.

Our scripts should check the installed bash version and exit with an error if it's older than v4. If it's possible to accommodate other shells (e.g. zsh, fish, nushell, etc.), ideally we should do that as well.

jw013:
@vasucp1207 can confirm but based on what I saw in the comments I am almost certain they are on Mac OS, which ships an ancient version of Bash by default (3.something, apparently for GPL-avoidance reasons) as well as BSD flavors of command line tools which differ from GNU versions. The missing Cargo.toml is likely due to a difference in `sed` behavior between Mac and GNU, https://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux

Writing reliable cross-platform scripts can be a rather tedious rabbit hole to dive into but there are some easy things we can do without going too deep.

- The `#!/usr/bin/env bash` at the top of the scripts should ensure that some version of bash is running the scripts so I don't think we need to worry about non-bash shells (zsh/fish/nushell etc) unless users go out of their way, e.g. `zsh some_script.sh` instead of `./some_script.sh`.

- Adding a check for Bash version 4+ is simple, basically checking that `$BASH_VERSION` exists and that `${BASH_VERSINFO[0]} -ge 4`. This should also catch any non-bash shells too since they would not set those variables. 

- The real tedium comes from trying to ensure that every external tool is installed and supports the required features. I took a brief look into the scripts directory and see `grep`, `sort`, `curl`, `jq`, and of course `sed`.  There may be more I missed. Not sure this level of effort is worth it for development-only scripts. Perhaps an easier approach would be to document in the Contributing guide that the scripts expect updated bash + a list of the required external programs.

obi1kenobi:
Great points! If you have a few minutes, would you care to open some PRs for the bash version check and the contributing guide suggestions? I'd be thrilled to merge them.

[...]

PR #682, opened 2024-03-04
Replace GNU-specific commands with more portable alternatives

jw013:
#680 enhancements

The replacement commands are functionally equivalent with one exception: new lints are now added to the beginning of the `add_lint!()` macro instead of the end. (There is a way to append but would require a slightly more complex `awk` command). More details in the commit messages.

### DIFF
--- a/scripts/make_new_lint.sh
+++ b/scripts/make_new_lint.sh
@@ -85,7 +85,7 @@ if awk -v lint_name="$NEW_LINT_NAME" '
 ' "$SRC_QUERY_FILE"; then
     printf ' already exists.\n'
 else
-    tmp=${SRC_QUERY_FILE}.tmp
+    tmp="${SRC_QUERY_FILE}.tmp"
     sed -e '/^add_lints!($/ a\'"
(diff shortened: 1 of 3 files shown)


scripts/make_new_lint.sh
```diff
@@ -78,19 +78,21 @@ fi
...
+    searching && $0 ~ "[[:space:]]" lint_name "," { found = 1; exit }
+    END { if (found) { exit 0 } else { exit 1 } }
+' "$SRC_QUERY_FILE"; then
+    printf ' already exists.\n'
 else
-    # -E = extended regex mode, which behaves more similarly to regex in most programming languages.
-    # -z = use \0 as separators instead of \n, so that we can do multi-line matches.
-    sed -i'' -Ez "s/add_lints\!\\(([^)]+)\\)/add_lints\!(\\1    $NEW_LINT_NAME,\n)/" "$SRC_QUERY_FILE"
-    echo ' done!'
+    tmp=${SRC_QUERY_FILE}.tmp
```
obi1kenobi, line 88:
```suggestion
    tmp="${SRC_QUERY_FILE}.tmp"
```
in case there are spaces in the path

obi1kenobi (approved):
Awesome, thanks so much for putting this together! 🚀

PR #683, opened 2024-03-06
Remove script usage of globstar, improving compatibility

jw013:
More enhancements for #680. This removes the requirement for v4+ of bash.


obi1kenobi (approved):
Wow, TIL about `$PPID`. Amazing job! 🚀

PR #684, opened 2024-03-06
add trait_method_now_doc_hidden lint

suaviloquence:
Matches when a previously public trait method/function has been marked as #[doc(hidden)], which is a semver-breaking change. 

It matches trait methods and trait associated functions, but the other lints were listed under trait_method instead of trait_function or something else, so I followed that pattern.

Fixes part of #578  https://github.com/obi1kenobi/cargo-semver-checks/issues/578#issue-1982324505


obi1kenobi (approved):
Nicely done! 🚀

PR #688, opened 2024-03-08
feat: insta for snapshot testing

vasucp1207:
Related to https://github.com/obi1kenobi/cargo-semver-checks/issues/634

> [!WARNING]  
> This is just for checking how to update snapshots with `insta`. Not a migration from the current test system to `insta` snapshots.

Trying out the insta for snapshot testing and find out that moving to `insta` is not a big deal we just have to adjust the current testing logic.
Run `cargo-test` and `cargo-insta-accept`, snapshots will created in `src/snapshots` with ron format. No need for manually paste the diff.

cc @obi1kenobi


Cargo.toml
```diff
@@ -48,6 +48,7 @@ rayon = "1.8.0"
 
 [dev-dependencies]
 assert_cmd = "2.0"
+insta = { version = "1.36.1", features = ["yaml"] }
```
CommanderStorm, line 51: Adding the following at the bottom of this file might be benefitial according to their docs

```toml
# See https://insta.rs/docs/quickstart/#optional-faster-runs
[profile.dev.package]
insta.opt-level = 3
similar.opt-level = 3
```

src/query.rs
```diff
@@ -407,6 +407,10 @@ mod tests {
         sort_individual_outputs(&mut expected_results);
         sort_individual_outputs(&mut actual_results);
 
+        let output_snap =
+            ron::ser::to_string_pretty(&expected_results, ron::ser::PrettyConfig::default()).unwrap();
+        insta::assert_snapshot!(query_name, output_snap);
```
CommanderStorm, line 412: Out of curiosity: Why did you use this vs `insta::assert_ron_snapshot!(..)`?
This seems a bit simpler, or am I missing something?

Cargo.toml
obi1kenobi: This PR is a bit outdated at this point, and the `Cargo.toml` already has a broader version of those directives already: `profile.dev.package."*"` is `opt-level = 3`

obi1kenobi:
Hey @CommanderStorm, thanks for taking a look at this!

This PR has been dormant for a while, and I haven't heard from @vasucp1207 about any specific plans to resume working on it. (@vasucp1207 lmk if you're still interested, and no worries if not!)

Assuming @vasucp1207 isn't able to continue working on it (and we should give them a few days to reply), I'd be happy to have you take over the work and drive it home!

vasucp1207:
Since this is more than 4 months old I don't how the project has changed since then and what are the implications of this PR  overall. So yes @CommanderStorm you can take onto this PR I have no problem.

vasucp1207:
Close this in favour of PR  https://github.com/obi1kenobi/cargo-semver-checks/pull/951.

PR #689, opened 2024-03-09
New lint: add function_abi_no_longer_unwind lint

qstommyshu:
This PR is to implement the second task in #503


src/lints/unwind_function_missing.ron
```diff
@@ -0,0 +1,51 @@
+SemverQuery(
+    id: "unwind_function_missing",
+    human_readable_name: "unwind function missing",
```
obi1kenobi, line 3: I'm confused about the name of this lint. It seems like the function isn't missing, and the issue instead is that its ABI changed, right?

src/lints/unwind_function_missing.ron
```diff
@@ -0,0 +1,51 @@
...
+    description: "pub fn changed from an unwind-capable ABI to the same-named ABI without unwind: \"C-unwind\" to \"C\"",
```
obi1kenobi, line 4: We don't necessarily know that the change was *specifically* `C-unwind` to `C`. It's possible that the change is from a different unwind-capable ABI to its non-unwind-capable equivalent.

src/lints/unwind_function_missing.ron
```diff
@@ -0,0 +1,51 @@
...
+    required_update: Major,
+    reference_link: None,
```
obi1kenobi, line 6: Please include an appropriate reference link here. Anything is better than nothing. For example, we could link to the reference page that other ABI-related lints use, if there's nothing more specific.

src/lints/unwind_function_missing.ron
```diff
@@ -0,0 +1,51 @@
...
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+    },
+    error_message: "An unwind-capable ABI changed to same-named ABI without unwind",
+    per_result_error_template: Some("pub fn {{name}} changed ABI from {{abi_raw_name}} to {{new_abi_raw_name}} in {{span_filename}}:{{span_begin_line}}"),
+)
```
obi1kenobi, line 51: Please ensure all files end in a trailing newline — you should be able to configure your editor to do this automatically for you.
```suggestion
)

```

test_crates/unwind_function_missing/new/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub extern "C" fn unwind_function_becomes_without_unwind() -> () {}
```
obi1kenobi, line 1: In Rust, `-> ()` is implied when functions are not annotated as returning something else, so it's convention to omit it.
```suggestion
pub extern "C" fn unwind_function_becomes_without_unwind() {}
```

obi1kenobi (commented):
The lint query looks good! There are a few issues in the metadata that need fixing, and I recommend taking a close look at other ABI-related lints and using them as inspiration here.

src/lints/unwind_function_missing.ron
qstommyshu: Got it, thanks! Just changed my editor configuration.

src/lints/unwind_function_missing.ron
qstommyshu: Sorry for the confusion. My intention for this lint name is: the new version of code is "missing the original unwind function" as the ABI has been changed from an unwind ABI to an non-unwind ABI. 

If the lint name is misleading, would changing the lint name to pub_function_no_longer_unwind be better?

obi1kenobi: We want to use similar names for similar lints, since that consistency will help our users find and remember related lints. This lint is most similar to the existing `function_changed_abi` lint, so perhaps something like `function_abi_no_longer_unwind`?

We historically haven't been good about whether we include `pub` or not =/ Around when #58 gets resolved I'll probably go through all the lints and remove all `pub` prefixes from lints, since most lint names don't include them and `pub` is implied for semver checks anyway.

src/lints/unwind_function_missing.ron
qstommyshu: I see, I will change the description to "pub fn changed from an unwind-capable ABI to the same-named ABI without unwind" then. So that the description doesn't specify if the change was *specifically* `C-unwind` to `C` to anticipate new types of unwind capable ABI being added to rust in the future. Is this an appropriate approach to accommodate this issue?

obi1kenobi: Sounds good to me!

Here are a few more unwind-capable ABIs in case you're curious: https://github.com/rust-lang/rfcs/blob/master/text/2945-c-unwind-abi.md#other-unwind-abi-strings

If we expand the set to not-yet-stable ABIs, the set is even bigger:
https://doc.rust-lang.org/beta/unstable-book/language-features/c-unwind.html

We might want to use the RFC that introduced `unwind` ABIs as the reference link on line 6 btw: https://github.com/rust-lang/rfcs/blob/master/text/2945-c-unwind-abi.md

src/lints/unwind_function_missing.ron
obi1kenobi: Also, let's rename the PR to match the new name of the lint — if you don't get a chance to do it, I can do it before merging as well.

qstommyshu: Sure, I just updated the lint name and the PR name

src/lints/unwind_function_missing.ron
qstommyshu: Attached the [proposal](https://rust-lang.github.io/rfcs/2945-c-unwind-abi.html) to add `C-unwind` ABI. It contains lots of information about the design and behaviours. It also links to the RFC PR and issue to `C-unwind`.

src/lints/function_abi_no_longer_unwind.ron
```diff
@@ -0,0 +1,51 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+    },
+    error_message: "A pub fn changed from an unwind-capable ABI to the same-named ABI without unwind",
```
obi1kenobi, line 49:
```suggestion
    error_message: "A pub fn changed from an unwind-capable ABI to the same-named ABI without unwind. If that function causes an unwind (e.g. by panicking), its behavior is now undefined.",
```

obi1kenobi (approved):
Very nice! I added a brief bit of extra info to help users understand why this is a major breaking change, and will merge this promptly.

I'm planning to make a new release later today, and this lint will be part of it!

PR #714, opened 2024-03-22
Add lint inherent_associated_pub_const_missing

arpity22:
Lint for removed associated pub constant inside `ImplOwner`'s inherent impl

Issue mentioned in https://github.com/obi1kenobi/cargo-semver-checks/issues/366#issuecomment-1662820577

P.S : I named it missing inspired from lints like `enum_missing`, `trait_missing`, `pub_module_level_const_missing` etc but should it be removed like lints `trait_removed_associated_type`, `trait_removed_supertrait`?


test_outputs/inherent_associated_pub_const_missing.output.ron
```diff
@@ -0,0 +1,26 @@
...
+            "path": List([
+                String("inherent_associated_pub_const_missing"),
+                String("StructA"),
+            ]),
+            "span_begin_line": Uint64(7),
+            "span_filename": String("src/lib.rs"),
+            "visibility_limit": String("public"),
+        },
+    ],
+}
```
obi1kenobi, line 26:
```suggestion
}

```
Consider making your editor add these trailing newlines automatically.

src/lints/inherent_associated_pub_const_missing.ron
```diff
@@ -0,0 +1,64 @@
...
+                            path @output @tag
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        inherent_impl {
+                            public_api_eligible @filter(op: "=", value: ["$true"])
+                            
+                            associated_constant {
+                                associated_constant: name @output @tag
+                                public_api_eligible @filter(op: "=", value: ["$true"])
```
obi1kenobi, line 25: If you get a chance, it would be great to expand the test coverage to include checking that these `public_api_eligible` filters are applied properly.

Right now, I'm pretty sure we could remove those filters (lines 21 and 25) and the tests would still pass.

Test cases for this would include situations where `#[doc(hidden)]` exists on the `impl` block, on the `pub const` itself, or both. It could of course exist in both the baseline and the current versions, or exist only in one of them.

This lint should fire only when the constant is truly removed (i.e. using it produces a compilation error) and not when the constant stops being public API — as I'm sure you remember from writing a bunch of lints for that, that's going to be a different lint.

obi1kenobi (commented):
Good point on the name consistency question. I think you chose a reasonable name for this lint. We may change up all the lint names in the future to improve our consistency. Would you like to open an issue for that and include your consistency suggestion there?

The lint query looks correct and I'd be happy to merge it. If you get a chance, I'd love to get a bit more test coverage first, since there are some tricky edge cases here that we currently aren't testing. I made a concrete suggestion for this in the PR itself.

test_crates/inherent_associated_pub_const_missing/new/src/lib.rs
```diff
@@ -1,10 +1,26 @@
...
+#[doc(hidden)]
+pub struct StructD {}
+impl StructD {
+    pub const PublicConstantG: i32 = 0;
+}
+pub struct StructE {}
+#[doc(hidden)]
+impl StructE {
+    pub const PublicConstantI: i32 = 0;
+}
```
obi1kenobi, line 26: Could you add these kinds of variants to the baseline as well, so that we also test whether removals of `doc(hidden)` in all these places results in the lint being triggered?

Right now, I think these tests _still_ don't target the `public_api_eligible` lines in the query I pointed out, because those lines were in the `baseline` arm of the query, whereas this code is only evaluated in the `current` arm since it's in the `new` directory in the test crate.

Also, as a bit of a nitpick: please consider adding blank lines between different items, such as between structs and impl blocks etc. Code this tight is quite difficult to read, even if `rustfmt` lets it go through.

test_crates/inherent_associated_pub_const_missing/old/src/lib.rs
```diff
@@ -16,3 +18,20 @@ impl StructB {
...
     const ConstantB: i32 = 0;
 }
+#[doc(hidden)]
+pub struct StructC {}
+impl StructC {
+    // should not be caught on removing since the struct #[doc(hidden)]
+    pub const PublicConstantF: i32 = 0;
+}
+//should not be caught by this lint on marking #[doc(hidden)]
+pub struct StructD {}
```
obi1kenobi, line 28: Another formatting nitpick that isn't super important in this test code, but will be critical if working on the cargo-semver-checks source itself: it's good to stay consistent with style and formatting. `rustfmt` does a lot, but it can't do everything yet.

In this block I've highlighted, there are four mutually-inconsistent styles of comments: capitalized vs uncapitalized, each both seen with a space after the `//` and without a space. Ideally, everything should be `// Capitalized concise sentence.`.

Consistent style improves readability and gives readers a sense of ease and quality when looking at the code. If code looks messy, people will have negative impressions of it (and often, of its authors), even if the code itself works well.

test_crates/inherent_associated_pub_const_missing/new/src/lib.rs
arpity22: Sorry for the misunderstanding, so we want test cases for when we remove the const from impl block/struct that were #[doc(hidden)] or we remove the const that was #[doc(hidden)] and the lint should be triggered in such cases

obi1kenobi: We want both kinds of test cases.

At a high level, our tests serve three purposes:
- **True-positive vs false-negative:** we expected a lint, and got a lint.
- **False-positive vs true-negative:** we didn't expect a lint, and we didn't get one.
- **One lint per problem:** we don't want to flag the same issue more than once, e.g. if a struct gets deleted, we don't _also_ want to trigger lints for each of its methods and associated constants that got deleted too.

So we want to check `#[doc(hidden)]` in all the places where it can appear, both where it's added in `new` and where it's removed in `new`, as well as where it remains unchanged between `old` and `new`.

Currently, you have the first kind of test covered well.

Removing `#[doc(hidden)]` from any of these places while also removing the constant is not a major change, and shouldn't trigger the lint. This is the second kind of test, and is not currently covered.

Making the const, its impl block, or the owning type become `#[doc(hidden)]` is a major change, but one for a different lint. (Same if the owning type got deleted altogether.) Even if the const itself is removed altogether, this lint shouldn't trigger here. This is the third kind of test, and is partially covered.

arpity22: Summary of added test cases:
1. Removing const for doc hidden struct - not triggered
2. Removing const for doc hidden impl - not triggered
3. Adding doc hidden for impl block but not removing const - not triggered
4. Adding doc hidden for impl block but removing const - triggered
5. Adding doc hidden for struct -  only `struct_now_doc_hidden` triggered
6. Removing doc hidden for struct, impl block and/or removing the const - not triggered

obi1kenobi: Excellent, nicely done!

src/lints/inherent_associated_pub_const_missing.ron
```diff
@@ -0,0 +1,64 @@
...
+                    ... on ImplOwner {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        name @output
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        inherent_impl @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]){
```
obi1kenobi, line 47: Ah one last test case is needed here.

The following might not always be a breaking change, and should be covered by a separate lint:
```rust
pub trait Trait {}

pub struct Example;

impl Example {
    pub const N: i64 = 0;
}

impl Trait for Example {}
```
to
```rust
pub trait Trait {
    const N: i64 = 0;
}

pub struct Example;

impl Trait for Example {}
```

The reason this might not be breaking is that if both `Example` and `Trait` are in a prelude, or if the `Trait` trait is already otherwise in scope, `Example::N` will continue to work. In the original code that refers to an inherent const, but in the new code it's a trait's const, equivalent to `<Example as Trait>::N` in explicit form.

To fix this, we allow the query to look at _all_ `impl` blocks (including `impl Trait for Example` ones) and not just inherent `impl` (only `impl Example`) blocks:
```suggestion
                        impl @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]){
```

It would be great if you could add the above code as a test case here.

obi1kenobi (commented):
Sorry, one last thing I didn't notice in the last pass, and this is good to go!

test_outputs/inherent_associated_pub_const_missing.output.ron
```diff
@@ -33,5 +33,16 @@
...
+            "associated_constant": String("N"),
+            "name": String("Example"),
+            "path": List([
+                String("inherent_associated_pub_const_missing"),
+                String("Example"),
+            ]),
+            "span_begin_line": Uint64(87),
+            "span_filename": String("src/lib.rs"),
+            "visibility_limit": String("public"),
+        },
```
obi1kenobi, line 46: Oooh interesting, this is a false-positive — we wanted this to _not_ get flagged.

Perhaps because the `N` constant isn't defined in the `impl` block and is instead a default item on the trait? 🤔

Sorry this is turning into a bit of a rabbit hole. I see two options to move forward here.

If this PR isn't something you find that interesting and you don't want to go down the rabbit hole (which is totally fine!), we can merge it as-is and I can open a separate issue for resolving this false-positive.

If you find this PR interesting and you'd like to continue polishing it, then it would be great to do the following:
- Add a new test case, similar to the trait+struct one we have already, but where the `const` doesn't have a default value and is instead given a value in the `impl` block for the trait.
- Add comments on both the old and new test cases, to explain what they test, how they are different, and why it's important to have both of them and not just one or the other.
- Depending on how this new test case interacts with the query (whether it gets flagged or not), we'd continue tweaking the query -- for example, by checking two separate cases explicitly: "inherent impl must not have the constant" and "impl of a trait, where that trait must not have a const by that name either."

Your call! Let me know how you'd prefer to move forward.

arpity22: I think I would keep working on it but if you think this is a rare case and it would be more beneficial to have the lint as it is currently than to not have it at all you can merge it. I don't have a problem either way.

obi1kenobi: Nice! There's no rush to merge this if you are happy to keep working on it. I was just offering it in case you felt frustrated by the back-and-forth on it, or if you just wanted a change of pace to something else.

It's great that the new test case didn't trigger the lint. Let's try tweaking the lint then.

We probably want to go back to `inherent_impl` like we had before. Then we probably want to add another block that also says "no implemented trait has a const by that name either." Here's a short snippet in the playground to help you get started: [link](https://play.predr.ag/rustdoc#?f=2&q=query---0Crate---2item---4impl---6implemented_trait---8*3-continue-from-here*l--_6--*4--*2--*0*J*l*J&v=*C*l*l*J)

Lmk if you get stuck or would like another pair of eyes on something!

src/lints/inherent_associated_pub_const_missing.ron
```diff
@@ -0,0 +1,64 @@
...
+                        name @output
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        impl @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]){
+                            associated_constant {
+                            name @filter(op: "=", value: ["%associated_constant"])
```
obi1kenobi, line 49:
```suggestion
                                name @filter(op: "=", value: ["%associated_constant"])
```
small indentation error I noticed

src/lints/inherent_associated_pub_const_missing.ron
```diff
@@ -0,0 +1,86 @@
...
+                            path @filter(op: "=", value: ["%path"])
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        impl @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]){
+                            implemented_trait {
+                                trait {
+                                    associated_constant {
+                                        name @filter(op: "=", value: ["%associated_constant"])
+                                    }
```
arpity22, line 70: Block to check if there is an implemented trait with the const

arpity22, line 70: Block to check if there is an implemented trait with the const

obi1kenobi: You could actually take this `impl` block and move it into the other `item` block, side by side with its existing `inherent_impl` block.

The query engine will still optimize this query to essentially the same thing as written, but it's still unnecessarily long and confusing for human readers so probably a good idea to apply this change.

Also, a nitpick: try to polish up the small details, like indentation and formatting. Right now, the query has some unexpected indentation, inconsistency in spacing (e.g. no space between `){`), and other elements that seem possibly accidental. While no individual thing like this is a big deal on its own, as a whole they do get in the way and make maintainability harder over time. Sometimes they even cause bugs, and even if they don't, they still make code harder to read and understand for maintainers and new contributors.

src/lints/inherent_associated_pub_const_missing.ron
```diff
@@ -0,0 +1,75 @@
...
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        inherent_impl @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
+                            associated_constant {
+                                name @filter(op: "=", value: ["%associated_constant"])
+                            }
+                        }
+                        
+                        or:
```
obi1kenobi, line 53:
```suggestion
```
this seems possibly unintentional -- it defines an alias that isn't used anywhere

test_crates/inherent_associated_pub_const_missing/old/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+    pub const PublicConstantRenameC: i32 = 0;
+}
+
+pub struct StructD {}
+
+#[doc(hidden)]
+impl StructD {
+    pub const PublicConstantE: i32 = 0;
+    pub const PublicConstantF: i32 = 0;
+}
```
obi1kenobi, line 41:
```suggestion
// The constants in this block shouldn't be flagged as removed,
// since they are `#[doc(hidden)]` as a result of the block itself.
#[doc(hidden)]
impl StructD {
    pub const PublicConstantE: i32 = 0;
    pub const PublicConstantF: i32 = 0;
}
```

test_outputs/inherent_associated_pub_const_missing.output.ron
```diff
@@ -0,0 +1,37 @@
...
+        {
+            "associated_constant": String("PublicConstantI"),
+            "name": String("StructE"),
+            "path": List([
+                String("inherent_associated_pub_const_missing"),
+                String("StructE"),
+            ]),
+            "span_begin_line": Uint64(50),
```
obi1kenobi, line 32:
```suggestion
            "span_begin_line": Uint64(52),
```

obi1kenobi (approved):
Looks good! I'm making a couple of tiny tweaks: adding a small explanatory comment as a "note to future self" to improve maintainability, and removing what seems to be a spurious addition to the query.

obi1kenobi:
Marked for merge! Thanks again for putting this together! 🚀

PR #721, opened 2024-03-29
Add `--color` setting to CLI.

### LINKED ISSUE #719: Support flags like `--color=always` and `--color=never`
These flags would override when `cargo-semver-checks` uses ANSI codes to print colorized output. Both `cargo` and `rustc` use these flags, and we should support them too.

Use case came up in https://github.com/obi1kenobi/cargo-semver-checks-action/pull/65

suaviloquence:
Fixes #719 

There is already code to recognize the `CARGO_TERM_COLOR` environment variable for setting color policy in `GlobalConfig::new()`.  This PR makes it so that the `--color=...` flag takes precedence over the env var, but if the CLI flag is not set, then the env var still applies if it is set.

I'm not super happy with needing the `if let` block that comes with `set_color_choice` in order to not overwrite the env var - I'd definitely be willing to refactor the env var detection into a different function or make some other change.

I wasn't sure if tests were necessary/how to test this.  If you want tests, let me know and I can add them.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -284,7 +284,7 @@ impl Check {
     pub fn log_level(&self) -> Option<&log::Level> {
         self.log_level.as_ref()
     }
-    
+
     /// Set the termcolor [color choice](termcolor::ColorChoice)
     /// If `None`, the use of colors will be determined automatically by
     /// CARGO_TERM_COLOR env var and tty type of output`
@@ -292,7 +292,7 @@ impl Check {
         self.color_choice = choice;
         self
     }
-    
+
     /// Get the current color choice.  If `None`, the use of colors is determined
     /// by the `CARGO_TERM_COLOR` env var and whether the output is a tty
     #[inline]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,6 +35,7 @@ pub struct Check {
     current: Rustdoc,
     baseline: Rustdoc,
     log_level: Option<log::Level>,
+    color_choice: Option<termcolor::ColorChoice>,
     release_type: Option<ReleaseType>,
     current_feature_config: rustdoc_gen::FeatureConfig,
     baseline_feature_config: rustdoc_gen::FeatureConfig,
@@ -247,6 +248,7 @@ impl Check {
             current,
             baseline: Rustdoc::from_registry_latest_crate_version(),
             log_level: Default::default(),
+            color_choice: None,
             release_type: None,
             current_feature_config: rustdoc_gen::FeatureConfig::default_for_current(),
             baseline_feature_config: rustdoc_gen::FeatureConfig::default_for_baseline(),
@@ -281,6 +283,21 @@ impl Check {
     #[inline]
     pub fn log_level(&self) -> Option<&log::Level> {
         self.log_level.as_ref()
+    }
+    
+    /// Set the termcolor [color choice](termcolor::ColorChoice)
+    /// If `None`, the use of colors will be determined automatically by
+    /// CARGO_TERM_COLOR env var and tty type of output`
+    pub fn with_color_choice(&mut self, choice: Option<termcolor::ColorChoice>) -> &mut Self {
+        self.color_choice = choice;
+        self
+    }
+    
+    /// Get the current color choice.  If `None`, the use of colors is determined
+    /// by the `CARGO_TERM_COLOR` env var and whether the output is a tty
+    #[inline]
+    pub fn color_choice(&self) -> Option<&termcolor::ColorChoice> {
+        self.color_choice.as_ref()
     }
 
     pub fn with_release_type(&mut self, release_type: ReleaseType) -> &mut Self {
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -283,6 +285,21 @@ impl Check {
         self.log_level.as_ref()
     }
 
+    /// Set the termcolor [color choice](termcolor::ColorChoice)
+    /// If `None`, the use of colors will be determined automatically by
+    /// CARGO_TERM_COLOR env var and tty type of output`
```
obi1kenobi, line 290: the trailing backtick seems like a possible typo
```suggestion
    /// the `CARGO_TERM_COLOR` env var and tty type of output
```

obi1kenobi (commented):
Looks great so far!

Re: testing, could you add some tests of the various `--color` options and env var settings into a new file in the `tests` directory? You can use existing tests like the ones in `rustdoc_edge_cases.rs` as inspiration for how to pass arguments, manipulate env vars, and check stdout / stderr to ensure the output includes / avoids color codes, as expected.

Ideally, we'd have both a "all lints passed" and a "some lints failed" run of `cargo-semver-checks`, and the test would ensure each of them produces the appropriate output (non-)colorization based on the env vars and CLI flags, with the appropriate precedence between them.

pksunkara:
Tangential comment. Please disregard since it's only for future note.

The recommended SOTA color-related logic is to use the [rust-cli/anstyle](https://github.com/rust-cli/anstyle) crates.

suaviloquence:
Added some tests in the `tests/color_config.rs`.  Right now it just tests for the presence/absense of the ANSI color sequence `\u{1b}[` in stderr based on the flag/env var combination.  I didn't do anything with `auto` because I'm not sure how to emulate a tty with assert-cmd.  I also introduced the `predicates` crate as a direct dev dependency (it was already required by assert_cmd; I just put it in Cargo.lock) in order to test if the stderr contained the sequence vs. matching the whole input which wouldn't be static (because of timing, for one).

src/main.rs
```diff
@@ -282,6 +289,14 @@ struct CheckRelease {
 
     #[command(flatten)]
     verbosity: clap_verbosity_flag::Verbosity<clap_verbosity_flag::InfoLevel>,
+
+    #[arg(value_enum, long = "color")]
+    /// Whether to print colors to the terminal:
+    /// always, alwaysansi (always use only ANSI color codes),
+    /// auto (based on whether output is a tty), and never
+    ///
+    /// Default is auto (use colors if output is a TTY, otherwise don't use colors)
```
obi1kenobi, line 298: Just matching the order between the macro attribute and doc comment as used elsewhere in the file.
```suggestion
    /// Whether to print colors to the terminal:
    /// always, alwaysansi (always use only ANSI color codes),
    /// auto (based on whether output is a tty), and never
    ///
    /// Default is auto (use colors if output is a TTY, otherwise don't use colors)
    #[arg(value_enum, long = "color")]
```

src/lib.rs
```diff
@@ -283,6 +285,21 @@ impl Check {
         self.log_level.as_ref()
     }
 
+    /// Set the termcolor [color choice](termcolor::ColorChoice)
+    /// If `None`, the use of colors will be determined automatically by
+    /// the `CARGO_TERM_COLOR` env var and tty type of output
```
obi1kenobi, line 290: Just ensuring the sentences don't run into each other.
```suggestion
    /// Set the termcolor [color choice](termcolor::ColorChoice).
    /// If `None`, the use of colors will be determined automatically by
    /// the `CARGO_TERM_COLOR` env var and tty type of output.
```

tests/color_config.rs
```diff
@@ -0,0 +1,97 @@
...
+            "--baseline-root=old/",
+        ]);
+
+    if let Some(color_arg) = color_arg {
+        cmd.arg("--color").arg(color_arg);
+    }
+
+    if let Some(color_env) = color_env_var {
+        cmd.env("CARGO_TERM_COLOR", color_env);
+    }
```
obi1kenobi, line 34: GitHub Actions might set `CARGO_TERM_COLOR` on the process that runs `cargo test`, and in this case this command will probably inherit it. We had a flaky test for this reason in the past.

Let's explicitly un-set the env var if `color_env_var` is None.
```suggestion
    if let Some(color_env) = color_env_var {
        cmd.env("CARGO_TERM_COLOR", color_env);
    } else {
        cmd.env_remove("CARGO_TERM_COLOR");
    }
```

obi1kenobi (approved):
Nicely done, thank you! I'm going to slip in a couple of tiny tweaks for a couple of nitpicks I had, and mark for merge as soon as CI passes.

obi1kenobi:
> Added some tests in the `tests/color_config.rs`. Right now it just tests for the presence/absense of the ANSI color sequence `\u{1b}[` in stderr based on the flag/env var combination. I didn't do anything with `auto` because I'm not sure how to emulate a tty with assert-cmd. I also introduced the `predicates` crate as a direct dev dependency (it was already required by assert_cmd; I just put it in Cargo.lock) in order to test if the stderr contained the sequence vs. matching the whole input which wouldn't be static (because of timing, for one).

Also, this explanation was great! It was very useful context that helped me understand the pull request much more easily. Thanks for putting it together!

PR #725, opened 2024-03-30
Print crate features in verbose mode (#711)

### LINKED ISSUE #711: Verbose runs should show which features were enabled in the crate versions being tested
Recently, a user had an issue where there was some confusion over which features were being enabled. The user submitted the verbose log (`--verbose`) of their run, but that still did not indicate which features were being enabled.

As a result, the user couldn't easily tell if the feature in question was enabled or not in their run. This is something we should fix.

Often, crates have many features, and the feature names may be long. So it might not be appropriate to output all the enabled features by default. But a `--verbose` run should *definitely* output which features were selected, both in the baseline and in the current version.

jw013:
Resolves #711. 

Implements the format as described in https://github.com/obi1kenobi/cargo-semver-checks/issues/711#issuecomment-2025950808. 

Note: I used a join without spaces to make the output usable as a command-line argument if copy-pasted.

### DIFF
--- a/src/rustdoc_cmd.rs
+++ b/src/rustdoc_cmd.rs
@@ -405,6 +405,17 @@ fn create_placeholder_rustdoc_manifest(
                     ..DependencyDetail::default()
                 },
             };
+            config.log_verbose(|config| {
+                if project_with_features.features.is_empty() {
+                    return Ok(());
+                }
+                writeln!(
+                    config.stderr(),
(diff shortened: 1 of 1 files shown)


pksunkara (commented):
LGTM unless I am missing something in the context.

obi1kenobi:
Excellent, thank you both 🚀

PR #726, opened 2024-03-30
Do not print --features with empty list (#718)

### LINKED ISSUE #718: Build failure repro command may include `--features` flag without any features
Note that the repro command below has an empty list of features after the `--features` flag:
```
note: the following command can be used to reproduce the compilation error:
      cargo new --lib example &&
          cd example &&
          echo '[workspace]' >> Cargo.toml &&
          cargo add --path /tmp/sequoia-cert-store --no-default-features --features  &&
          cargo check
```

This should be fixed.

Surfaced in #717.

jw013:
Fixes #718

### DIFF
--- a/src/rustdoc_cmd.rs
+++ b/src/rustdoc_cmd.rs
@@ -192,14 +192,18 @@ impl RustdocCommand {
                             .expect("failed to create path string")
                     ),
                 };
-                let feature_list = features.into_iter().join(",");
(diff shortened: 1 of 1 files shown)


pksunkara (commented):
LGTM.

obi1kenobi:
Awesome, thank you both! 🚀

Issue #502, opened 2023-07-21
New lints: functions with explicit ABI export names


obi1kenobi:
Consider a function like this:
```rust
#[no_mangle]
extern "C" fn example() {
    // does stuff
}
```

The following change is non-breaking, since it's equivalent:
```rust
#[export_name = "example"]  // matches original function name
extern "C" fn renamed() {  // this function name doesn't matter anymore
    // does stuff
}
```

However, changing to any other `export_name` value (one that does not match the _original_ name of the function) is breaking.

Example code where you can play with `#[no_mangle]` and `#[export_name]`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=64663d9ca343dee3ba9d2c7cabe61ab4

We want the following lints:
- [ ] `#[no_mangle]` function starts being mangled (doesn't have `#[no_mangle]` nor `#[export_name]` anymore)
- [ ] `#[export_name]` function starts being mangled (doesn't have `#[no_mangle]` nor `#[export_name]` anymore)
- [ ] `#[no_mangle]` function has its export name changed to an `#[export_name]` where the new name doesn't match the old
- [ ] `#[export_name]` function has its export name changed to a new `#[export_name]` name
- [ ] `#[export_name]` function has its export name changed to `#[no_mangle]` with a function name that doesn't match the old `#[export_name]`

suaviloquence:
Hi! I started to work on this.  A couple of questions/uncertainties:
1. I'm assuming it's a major-level change based on other violations like changing an established `repr`, although I couldn't find anything on [the cargo semver page](https://doc.rust-lang.org/cargo/reference/semver.html). for no_mangle/export_name specifically
2. Is it a violation when a non-public function changes export name? For example, in the example you linked, you can still access `im_secretly_public` even when it's not public in the Rust API.  
3. Should this (these) lints proc when the export name of one function moves to a different function? For example,
```rust
#[no_mangle]
pub fn func1() {}

pub fn func2() {}
```

to

```rust
pub fn func1() {}

#[export_name = "func1"]
pub fn func2() {}
```

Thanks!

obi1kenobi:
Great questions!

> I'm assuming it's a major-level change based on other violations like changing an established `repr`, although I couldn't find anything on [the cargo semver page](https://doc.rust-lang.org/cargo/reference/semver.html). for no_mangle/export_name specifically

[...]

PR #728, opened 2024-03-30
add function_export_name_changed lint

suaviloquence:
Second bullet point of #502.  Should I put all the lints in one (draft, I haven't finished them yet) PR or do separate PRs?


suaviloquence:
FWIW [here](https://github.com/obi1kenobi/cargo-semver-checks/actions/runs/8492696855/job/23265966971?pr=728#step:8:497)'s the line that breaks nightly tests - it doesn't seem like an issue with this lint.

obi1kenobi:
Separate PRs for each lint is fine. That's usually faster and easier to review, and shows up more nicely in the changelog as well.

That test has been flaky for a while and I'm not sure why. I've never managed to reproduce the flakiness locally, so my gut feeling is that it might possibly be a GitHub Actions bug of some sort? Not sure, haven't had time to dig into it... Don't worry about it, I don't think you broke any of it — I just restarted it and I expect it'll probably pass.

src/lints/function_export_name_changed.ron
```diff
@@ -0,0 +1,55 @@
+SemverQuery(
+    id: "function_export_name_changed",
+    human_readable_name: "function's export name has changed",
+    description: "A function's ABI name with #[no_mangle] or #[export_name = \"name\"] has changed",
+    required_update: Major,  // Not specified in rust semver doc? most similar is for #[repr] changes: https://doc.rust-lang.org/cargo/reference/semver.html#type-layout
```
obi1kenobi, line 5: Yeah this is definitely major, and using the `#[no_mangle]` doc link as a reference is a reasonable place for now until the cargo semver docs get a section for this.

src/lints/function_export_name_changed.ron
```diff
@@ -0,0 +1,55 @@
...
+                    }
+                }
+            }
+            
+            current {
+                item {
+                    ... on Function {
+                        name @output
+
+                        export_name @filter(op: "!=", value: ["%export_name"]) @filter(op: "!=", value: ["$null"])
```
obi1kenobi, line 27: This is a failure of our docs, so don't worry about it -- there's a built-in filter operator for this:
```suggestion
                        export_name @filter(op: "!=", value: ["%export_name"]) @filter(op: "is_not_null")
```

src/lints/function_export_name_changed.ron
```diff
@@ -0,0 +1,55 @@
...
+                    ... on Function {
+                        export_name @filter(op: "=", value: ["%export_name"])
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "zero": 0,
+        "null": null,
```
obi1kenobi, line 50:
```suggestion
```
since we used the built-in filter operator for this

src/lints/function_export_name_changed.ron
```diff
@@ -0,0 +1,55 @@
...
+                            begin_line @output
+                        }
+                    }
+                }
+                
+                item @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
+                    ... on Function {
+                        export_name @filter(op: "=", value: ["%export_name"])
+                    }
+                }
```
obi1kenobi, line 44: It might be a good idea to write the lint for "export name no longer exists" first, just to get us a bit more test coverage and help us decide how to split the responsibilities between this lint and that one.

Here's what I'm worried about.

If both this lint and that other one enforce that the export name doesn't exist anymore, then it sounds like this lint is just a stricter version of that other one -- so whenever this one triggers, the other one will too. We want to avoid that.

So maybe we reshape this lint a bit to say "a function in the public API has changed its export name in the ABI" instead. So the invariant that is broken (and that we're reporting) is "call `func()` via the API, and `export_func()` via the ABI, those do the same thing because they are the same function." If the ABI export name changes to something else, then even if the ABI name `export_func()` continues to exist on some other function, this is likely worth flagging because it's a possible bug.

If we do end up taking that route, then I think:
- This block I've highlighted becomes unnecessary.
- The other lint doesn't need to look at `importable_path` at all, and can just focus on `export_name` alone.

What do you think?

test_crates/function_export_name_changed/new/src/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+
+/// negative test - export name on one function gets moved to the other
+/// this is not necessarily a breaking change, as long as the ABIs are the same
+/// but that is a different lint
+pub mod export_name_moved {
+    pub fn export_name_moved_1() {}
+    
+    #[export_name = "export_name_moved"]
+    pub fn export_name_moved_2() {}
+}
```
obi1kenobi, line 37: I think we may want to flag this, because a public API call to `export_name_moved::export_name_moved_1()` is no longer equivalent to a call to `export_name_moved()` via the ABI.

See comment on the lint above for details.

obi1kenobi (commented):
I might have unintentionally pointed you a bit in the wrong direction on this. Sorry about that!

It should be easy to straighten things out. This is part of why semver linting is so hard! Too many edge cases! But getting them right is important, and I feel like our users will appreciate it.

src/lints/function_export_name_changed.ron
suaviloquence: Yeah, I was thinking about that when I was writing it - it was pretty much the same query except checking if the current `export_name` *is* null. If we don't need granular feedback between missing vs changed, they can be easily combined. Do you think it's necessary?

suaviloquence: I'm not at my computer right now but I can push the other lint to a branch when I get a chance

obi1kenobi:
Yeah, we can just the missing one for both cases, I think that's fine.

suaviloquence:
All good. For these categories from the original issue:


> -  function with the same export name takes different number of arguments
> - function with the same export name has a parameter whose type changes to one with a different ABI
it doesn't matter what the type is called, just whether it has the same ABI or not
> - function with the same export name changes what it returns (starts / stops returning data, or changes to return type with different ABI)
this can be multiple lints; e.g. separate lints for starts / stops / changes return type ABI

Would these (just) apply to exported non-public functions? Because changing the API of a public function would already be a breaking change, so do we want to flag this twice?

obi1kenobi:
That's a great question.

I think it's reasonable to start by flagging twice here, because we're flagging two separate problems: the API is broken, and the ABI is also broken.

Down the line, if this proves to be problematic for any reason, we can build something more sophisticated. For example, we could have three lints instead of two: broken API but not ABI, broken ABI but not API, and both broken at once. But going straight for this seems like major overkill right now, so let's go with the simpler option of flagging twice for now.

suaviloquence:
OK, I consolidated the two cases

src/lints/function_export_name_changed.ron
```diff
@@ -0,0 +1,47 @@
+SemverQuery(
+    id: "function_export_name_changed",
+    human_readable_name: "function's export name has changed or been removed",
+    description: "A function's ABI name with #[no_mangle] or #[export_name = \"name\"] has changed or been removed",
+    required_update: Major,  // Not specified in rust semver doc? most similar is for #[repr] changes: https://doc.rust-lang.org/cargo/reference/semver.html#type-layout
+    reference_link: Some("https://doc.rust-lang.org/reference/abi.html#the-no_mangle-attribute"),  // TODO
```
obi1kenobi, line 6: Did you mean to leave both of these comments in the file that will get merged?

Asking because the TODO seems completed, and the other comment seemed like a question for reviewers that I believe I replied to in my previous comment near here.

src/lints/function_export_name_changed.ron
```diff
@@ -0,0 +1,47 @@
...
+                            begin_line @output
+                        }
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+    },
+    error_message: "the ABI export name of a function has been changed or been removed through the #[no_mangle] or #[export_name] attributes",
```
obi1kenobi, line 44: Let's capitalize this and make it a proper sentence to keep it in line with other lints.
```suggestion
    error_message: "The ABI export name of a function has been changed or removed through the #[no_mangle] or #[export_name] attributes.",
```

src/lints/function_export_name_changed.ron
```diff
@@ -0,0 +1,47 @@
...
+    per_result_error_template: Some("function {{join \"::\" path}} in file {{span_filename}}:{{span_begin_line}} with old export name {{export_name}}"),
```
obi1kenobi, line 45: The `span_filename` is a full path, which may be quite long, so I would recommend keeping the span info as the last thing in the template -- anything after it is likely to be missed.

Perhaps something like this:
```suggestion
    per_result_error_template: Some("function {{join \"::\" path}} no longer has ABI name {{export_name}} in {{span_filename}}:{{span_begin_line}}"),
```

test_crates/function_export_name_changed/new/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+
+/// negative test - a function's export name is added
+#[export_name = "export_name_added"]
+pub fn export_name_added() {}
```
obi1kenobi, line 43:
```suggestion
pub fn export_name_added() {}

```
Please consider tweaking your editor config to add terminating newlines on file save :)

test_crates/function_export_name_changed/old/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+/// is no longer equivalent to a call to `export_name_moved` in the ABI
+pub mod export_name_moved {
+    #[export_name = "export_name_moved"]
+    pub fn export_name_moved_1() {}
+    
+    pub fn export_name_moved_2() {}
+}
+
+/// negative test - a function's export name is added
+pub fn export_name_added() {}
```
obi1kenobi, line 44:
```suggestion
pub fn export_name_added() {}

```

test_crates/function_export_name_changed/old/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+#[export_name = "export_name_removed"]
+pub fn export_name_removed() {}
+
+/// positive test - a non-public function changes export name
+#[export_name = "private_export_name_changed_old"]
+pub fn private_export_name_changed() {}
+
+/// positive test - export name on one function gets moved to the other
+/// this is still a breaking change because a call to `export_name_moved_1` in the API
+/// is no longer equivalent to a call to `export_name_moved` in the ABI
```
obi1kenobi, line 35: Great comments throughout, and this one is particularly good! Well done!

src/lints/function_export_name_changed.ron
```diff
@@ -0,0 +1,47 @@
...
+                    }
+                }
+            }
+            
+            current {
+                item {
+                    ... on Function {
+                        name @output
+
+                        export_name @filter(op: "!=", value: ["%export_name"])
```
obi1kenobi, line 27: For the sake of easier debugging, let's capture for output the new export name here. That way the new name will get recorded in the snapshot files and will be available in `--verbose` runs, which simplifies debugging if users report issues with the lint:
```suggestion
                        new_export_name: export_name @filter(op: "!=", value: ["%export_name"]) @output
```

You will probably need to recreate the snapshot test data, or just add the extra column manually.

obi1kenobi (commented):
Just a set of tiny nitpicks. The lint is correct and the tests are in great shape as well. Less than 5min of polishing left and this is good to merge.

suaviloquence:
Thanks for the feedback!

obi1kenobi (approved):
Awesome job!

PR #733, opened 2024-04-01
make_new_lints adds lints in sorted order

### LINKED ISSUE #730: Make the `make_new_lint.sh` script avoid merge conflicts when registering the new lints
Currently, the `make_new_lint.sh` script adds new lints at the beginning of the list of lints, right at the start of the `add_lints!(` macro:
https://github.com/obi1kenobi/cargo-semver-checks/blob/1a48764775be6a536d9e816873e152de3b9c45c7/scripts/make_new_lint.sh#L79-L96

If multiple contributors are working on adding lints concurrently, they both going to attempt to add their new lints on that first line. This will inevitably cause a merge conflict, even though the resolution is simple: add both lints to the list. The conflict adds regrettable friction to our contributing process: contributors have to keep resolving the conflict or risk CI no longer running for them.

One possible way to solve this is to:
- lexicographically sort all lints in the `add_lints!()` macro's list
- have the `make_new_lint.sh` script add each lint into its lexicographically appropriate location, which is likely to avoid most conflicts.

We'd like the solution to work cross-platform with as few extra dependencies as possible, so if taking the above approach, tweaking the existing `awk` approach would be preferable to something completely new and more heavyweight.

Open to other approaches as well, if they happen to be easier to pull off.

Related to #682 so cc @jw013 in case you have an idea on how to best do this 👆

jw013:
Resolves #730

Script now imposes some additional requirements on query.rs which are all documented in the comments.

Note: `"" var` is an awk idiom that concatenates with empty string to force a variable to be string typed and not numeric. Not that I expect any lint names to start with a number but done out of caution anyways.


obi1kenobi (approved):
Amazing, I love it! Excellent work 🚀

You're clearly an awesome engineer, and I'm thrilled and honored you're directing your energy toward improving `cargo-semver-checks`. Both this project and the Rust ecosystem as a whole are better off for it.

jw013:
Thanks for your kind words. Working on this project has been a very pleasant experience and a lot of that has to do with your efforts as a maintainer providing prompt feedback on PR's and writing clear and detailed issues and docs.

jw013:
Sure, let me know what you had in mind and I'd be happy to help out where I can.

obi1kenobi:
Do you have a preference toward a particular facet of the code?

Broadly speaking, there are four major facets in `cargo-semver-checks` in use today:
- the user-facing CLI and surrounding code
- `cargo-semver-checks` as a library
- lints and the diagnostics they output
- test infrastructure and associated tooling

One possible way to give you more leverage and autonomy is to have you work toward owning one of these areas. So instead of me opening issues and you implementing solutions, you could independently get a sense for what's needed and work toward building it or helping others build it.

There are other ways to do this as well, of course. Another example is to delegate a broad, cross-cutting area for improvement, like "help `cargo-semver-checks` be better in GitHub Actions" or "help `cargo-semver-checks` be better when used in a workspace with many crates" and then have you work across all the functional facets to improve the experience there.

Does either of these approaches strike you as appealing?

I'm not set on a particular way of doing things, and I'm happy to experiment until we find something that feels right.

jw013:
Based on the issues I've worked on so far I have some familiarity with the CLI and lints but very little with the other two (the library aspect or testing), so can't say I have a strong preference yet. I've mostly been picking issues to work on by estimated size (anything that fits into a few hours on a weekend works well for me) and also the issues are a good way to learn by doing so I try to throw in some variety every once in a while too. I can keep doing that or I'd also be happy to work on whatever areas are higher priority, and of course you would have a better idea of what those are.

obi1kenobi:
Sounds good. I can keep an eye out for issues that match those parameters (fit into a few hours + some variety across the tool) and we can go from there.

Thanks again, and know that I really appreciate everything you do!

PR #650, opened 2024-02-04
Improves linting for issue 503 by adding lint to check ABI has not changed between functions with same `export_name`

ehiggs:
In this PR we check for ABI changes when the export name stayed the same.

Paired with Predrag at FOSDEM. :)

PR #734, opened 2024-04-02
Fix semver id

pksunkara:
Follow-up to #650. Corrects the mistake of wrong semver id.

For the future, I think the `id` should not exist in the `ron` file and instead should be taken from the file name, letting us avoid these kinds of mistakes.

### DIFF
--- a/src/lints/exported_function_changed_abi.ron
+++ b/src/lints/exported_function_changed_abi.ron
@@ -1,5 +1,5 @@
 SemverQuery(
-    id: "function_changed_export_name",
+    id: "exported_function_changed_abi",
     human_readable_name: "exported function changed ABI",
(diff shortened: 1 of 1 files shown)


obi1kenobi (approved):

obi1kenobi:
> For the future, I think the `id` should not exist in the `ron` file and instead should be taken from the file name, letting us avoid these kinds of mistakes.

Seems reasonable. I'd merge a PR for that, if you or someone else would be open to making one.

pksunkara:
> I'd merge a PR for that, if you or someone else would be open to making one.

I would probably get to it, but please create an issue so that we don't lose track of it.

obi1kenobi:
Opened #735.

In general, if you feel something like this should have an issue, feel free to just open one. I'm happy to open them as well, and I'm also happy to rely on your judgment when something deserves an issue and your description of the situation.

PR #739, opened 2024-04-10
Remove query id in favor of query filename

### LINKED ISSUE #735: Consider either testing that lint file names match their `id` fields, or removing the field entirely.
Sometimes new lint contributions accidentally fail to set the `id` field of new lints correctly. #734 fixed one such issue.

We want a field like `id` available on the deserialized value for purposes of naming the lint and reporting failures.

We should either have a test that makes sure each lint file name matches its `id` field, or remove the field from the file and somehow otherwise populate it during deserialization.

pksunkara:
fixes #735,

### DIFF
--- a/src/lints/union_field_missing.ron
+++ b/src/lints/union_field_missing.ron
@@ -43,7 +43,6 @@ SemverQuery(
                         field @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
                             visibility_limit @filter(op: "=", value: ["$public"])
                             name @filter(op: "=", value: ["%field_name"])
-
                         }
                     }
                 }
--- a/src/lints/union_field_missing.ron
+++ b/src/lints/union_field_missing.ron
@@ -1,8 +1,7 @@
 SemverQuery(
-    id: "union_field_missing",
     human_readable_name: "pub union pub field is removed or renamed",
     description: "pub union pub field is removed or renamed. No longer present under it's previous name, by whatever cause.",
-    required_update: Major, 
+    required_update: Major,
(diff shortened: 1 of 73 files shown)


src/query.rs
```diff
@@ -462,10 +462,10 @@ macro_rules! add_lints {
             )*
         }
 
-        fn get_query_text_contents() -> Vec<&'static str> {
+        fn get_queries() -> Vec<&'static str, &'static str)> {
```
obi1kenobi, line 465:
```suggestion
        fn get_queries() -> Vec<(&'static str, &'static str)> {
```

src/lints/union_field_missing.ron
```diff
@@ -44,7 +43,7 @@ SemverQuery(
                         field @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
                             visibility_limit @filter(op: "=", value: ["$public"])
                             name @filter(op: "=", value: ["%field_name"])
-                            
+
```
obi1kenobi, line 46:
```suggestion
```
let's just delete that line altogether, it isn't necessary

obi1kenobi (approved):

obi1kenobi:
Thanks! Fixed a minor typo to help tests pass, and now queued for merge.

pksunkara:
Trying to fix the test failures made me realize that this is the wrong approach because we are utilising `SemverQuery.id` in many places and passing the id around in all those places makes the code worse.

I would close this PR in favor of adding a test as described in #735 if you agree.

obi1kenobi:
Sounds good!

PR #765, opened 2024-04-19
improve config function consistency

### LINKED ISSUE #764: Configuration methods do not have consistent naming
>>Question on method naming: it seems like we now have a `set_level()` method where before we had a `with_log_level()` method. Is this divergence intentional, or should we instead have either `set_log_level()` or `with_log_level()` as before? Similar question about with the other `set_...()` methods, as opposed to the existing `with_...()` methods' naming convention.

> The `GlobalConfig::set_level` method already existed before this PR, so I left it unchanged.  Additionally, since we used to have that and the  `set_color_choice` method, I used the `set_` prefix for `GlobalConfig` setters.  I agree that it's not consistent between `GlobalConfig` and `Check`, but they are both internally consistent.  Should we use the same prefix for everything? If so, in here, or in a different PR?

_Originally posted by @suaviloquence in https://github.com/obi1kenobi/cargo-semver-checks/issues/737#issuecomment-2065467309_

If we want all the methods to be consistent across `Check` and `GlobalConfig`, we need to choose:
- do we want the `with_` (found in `Check`) prefix or the `set_` (from `GlobalConfig`) prefix?
- do we want `set_log_level` or `set_level`?

I definitely prefer `log_level` over `level`, and I have a slight preference for `set` over `with`. What do others think?

suaviloquence:
Fixes #764 

Config structs (`Check`, `GlobalConfig`, `PackageSelection`) now adhere to the following conventions for setter methods:

- method is called `set_[property]`
- method signature is `fn(&mut self, T) -> &mut Self`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -157,7 +157,7 @@ impl PackageSelection {
         }
     }
 
-    pub fn with_excluded_packages(&mut self, packages: Vec<String>) -> &mut Self {
+    pub fn set_excluded_packages(&mut self, packages: Vec<String>) -> &mut Self {
         self.excluded_packages = packages;
         self
     }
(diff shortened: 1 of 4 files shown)


suaviloquence:
Edited migration guide:
---
# Migration Guide

## Major changes

- `Check::check_release` takes a new `&mut GlobalConfig` parameter, and `GlobalConfig` now holds log level and color settings.
  - Change: 
    ```rust
    let mut check = Check::new(rustdoc);
    check.with_packages(packages)
         .with_log_level(Some(log_level))
         .with_color_choice(Some(termcolor::ColorChoice::Always));
    let result = check.check_release();
    ```
   - To:
     ```rust
     let mut config = GlobalConfig::new()
        .set_level(Some(log_level));
     config.set_color_choice(true); // or set_{err,out)_color_choice for finer-grained control

     let mut check = Check::new(rustdoc);
     check.with_packages(packages);
     let result = check.check_release(&mut config);
     ```
- Set color choice in one of three ways:
  - Change:
    ```rust
    let choice: termcolor::ColorChoice = ...;
    config = config.set_color_choice(choice);
    ```
   - To: one of the three following options, depending on desired granularity
     1. Set global color choice for all crates using the `anstream` library:
        ```rust
        let choice: anstream::ColorChoice = ...;
        // set *before* creating `GlobalConfig` instance (if using default stdout/stderr)
        // or before calling `GlobalConfig::set_std{out,err}` (if configuring them)
        choice.write_global(); // configures `choice` for all crates using `anstream`
        ```
      2. Set whether to use colors for both stdout and stderr in the `cargo-semver-checks` library:
          ```rust
          let use_colors: bool = ...;
          config.set_color_choice(use_colors);
          ```
      3. Set whether to use colors for a *specific stream* in the `cargo-semver-checks` library:
          ```rust
          let use_colors_for_stderr: bool = ...;
          let use_colors_for_stdout: bool = ...;
          config.set_err_color_choice(use_colors_for_stderr);
          config.set_out_color_choice(use_colors_for_stdout);
          ```

 - config structs (`Check`, `GlobalConfig`, `PackageSelection`) now have a consistent setter convention: `pub fn set_[property](&mut self, prop: T) -> &mut Self`:
   - Change:
     ```rust
     selection.with_excluded_packages(packages);
     check.with_selection(selection);
     ```
    - To:
      ```rust
      selection.set_excluded_packages(packages);
      check.set_package_selection(selection);
      ```
    - Change:
      ```rust
      config = config.set_level(level);
      ```
    - To:
      ```rust
      config.set_log_level(level); // now takes and returns &mut Self instead of owned Self
      ```
## New features

- Users can configure the stdout and stderr streams for `cargo-semver-checks` to be different from default:
  ```rust
  let stderr: Box<dyn std::io::Write> = Box::new(stderr_buffer);
  config.set_stderr(stderr);
  ```

## Minor changes
- `#[must_use]` added on `GlobalConfig::std{err,out}`
- `GlobalConfig::shell_print` takes an `anstyle::Color` instead of a `termcolor::Color`:
  - Change `config.shell_print(status, message, termcolor::Color::Red, justified)`
  - To `config.shell_print(status, message, anstyle::Color::Ansi(anstyle::AnsiColor::Red, justified)`
- To write colors to stdout/stderr, use `writeln!(config.stdout(), "{}styled text", anstyle::Style::new().bold().[...]);` instead of the `termcolor` API
- use `GlobalConfig::{out,err}_color_choice()` to query whether to print colors for the stdout/stderr stream instead of using `GlobalConfig::is_err_tty` as a heuristic
- `GlobalConfig::std{out,err}` returns an opaque `impl Write + '_` type instead of a `termcolor::StandardStream`

src/lib.rs
```diff
@@ -256,51 +256,51 @@ impl Check {
...
 
-    pub fn with_all_features(&mut self) -> &mut Self {
+    pub fn set_all_features(&mut self) -> &mut Self {
         self.current_feature_config.features_group = rustdoc_gen::FeaturesGroup::All;
         self.baseline_feature_config.features_group = rustdoc_gen::FeaturesGroup::All;
         self
     }
 
-    pub fn with_extra_features(
+    pub fn set_extra_features(
```
obi1kenobi, line 303: Hmm these new names are more confusing than the old names here.

For example, compare: "check with default features" vs "check set default features." The former sounds like what it does, the latter sounds like we're overriding which features the crate considers default, or something.

Would it be weird to leave the `impl Check` methods as `with_...()`, and make everything else `set_...()`? "Config set" makes sense, "package selection set" makes sense, "check set" does not really make sense.

suaviloquence: No, that definitely makes sense.  These aren't "setters" in the same sense as the other ones; they don't take an additional parameter.

obi1kenobi (approved):

obi1kenobi:
Great, ship it 🚀

Issue #767, opened 2024-04-19
`cargo semver-checks` takes much longer after upgrading to version 0.30.0 from 0.24.1


ysaito1001:
### Steps to reproduce the bug with the above code

In [smithy-rs](https://github.com/smithy-lang/smithy-rs) (the code generator for AWS SDK for Rust), we run `cargo semver-checks` as part of CI. As described [in this PR](https://github.com/smithy-lang/smithy-rs/pull/3594), we noticed that the execution time for a GitHub action step for semver-checks increased from 15 minutes to nearly 4 hours after we had upgraded the `semver-checks` version from 0.24.1 to 0.30.0.

We run semver-checks against 49 crates in CI, and it looks like only a few of them cause a long tail of the total execution time. One of them is `aws-sdk-ec2` (there could be offending crates but this happened to be the first I noticed).

Here are reproduction steps:

1. Clone the [aws-sdk-rust](https://github.com/awslabs/aws-sdk-rust) repository and run `cd aws-sdk-rust/sdk`.
2. Install `cargo semver-checks` version 0.24.1
```
cargo install cargo-semver-checks --locked --version 0.24.1
```
3. Run
```
cargo semver-checks check-release --baseline-rev main --manifest-path ec2/Cargo.toml -v -p aws-sdk-ec2 --all-features --release-type minor
```
This is pretty much how [our CI script runs](https://github.com/smithy-lang/smithy-rs/blob/13c04dafcf5d0f6d65a3edf54b78193f00b44265/tools/ci-scripts/codegen-diff/semver-checks.py#L50-L58) `semver-checks`.

4. Install `cargo semver-checks` version 0.30.0
```
cargo install cargo-semver-checks --locked --version 0.30.0
```
5. Run the same command as step 3

----

FYI, we are not blocked by this, since we can continue using `0.24.1` in our CI for now.

### Actual Behaviour

With version `0.24.1`, step 3 takes around 14 seconds:
```
   Completed [  13.446s] 45 checks; 45 passed, 6 unnecessary
```

With version `0.30.0`, step 5 does not return a result (at least not before I give and hit Ctrl+C):
```
    Starting 61 checks, 6 unnecessary on 10 threads
(stuck at this point)
```

### Expected Behaviour

I expect step 5 (with version `0.30.0`) to finish in a similar amount of time as version `0.24.1`.

#### Software version

cargo-semver-checks 0.30.0

#### Operating system

[...]

PR #768, opened 2024-04-20
Add lint pub_static_mut_now_immutable

### LINKED ISSUE #366: New schema and lints for const and static values

arpity22:
Lint for when a mutable static becomes immutable static

Issue mentioned in https://github.com/obi1kenobi/cargo-semver-checks/issues/366#issuecomment-1662820577

Resolves #366.


src/lints/pub_static_mut_now_immutable.ron
```diff
@@ -0,0 +1,50 @@
+SemverQuery(
+    id: "pub_static_mut_now_immutable",
+    human_readable_name: "pub static mut is now immutable",
+    description: "A mutable static became immutable and thus cannot be imported",
```
obi1kenobi, line 4: Is `imported` here intentional? I admit I haven't looked too closely at what the exact breakage is when a mutable static becomes immutable. I just want to make sure that it is indeed the `use` import statement that will trigger a compilation error if a static becomes immutable, since that seems a bit surprising to me.

src/lints/pub_static_mut_now_immutable.ron
```diff
@@ -0,0 +1,50 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+    },
+    error_message: "A mutable static is now immutable",
```
obi1kenobi, line 48: We may want to expand this error message a bit.

If you just ran `cargo-semver-checks` and it reported this problem, what information would you like to see printed as part of the error log? You'd like to know what happened, why it's a problem, and where to look for it -- in as few characters as possible. This is tough, but it's a skill like any other and practicing it will help.

obi1kenobi (commented):
Looks pretty good! Just a couple of polish items in the lint metadata and the user-facing diagnostic message.

Thanks for putting this together, much appreciated!

src/lints/pub_static_mut_now_immutable.ron
arpity22: I am not sure either. I just read the discussion on the issue and figured the user would not be able to use the static since they depended on it being mutable.

obi1kenobi: Ah okay. Let's step back for a sec, since I think the bigger picture is more important right now.

`cargo-semver-checks` exists as a tool because the rules for SemVer in Rust are very complex. We humans have a hard time learning them and applying them consistently, so we delegate some of that responsibility to this tool: it can be more consistent, and it also holds the distilled expertise of dozens of humans. Each person there dug into some possible SemVer problem, understood it in detail, figured out how to prevent it from ever happening again, and then wrote a lint that both catches and describes the problem to the end user.

If some of the distilled expertise is wrong, that violates users' trust. Why should they delegate trust and responsibility to a tool that violates their trust? They will stop using it.

This is an _existential threat_ to `cargo-semver-checks`. This is not exaggeration. Several prior attempts at a SemVer linter for Rust have been abandoned due to reporting incorrect information, such as false-positive lints. We may very well increment that number by one more if we aren't careful.

The _real_ value you are contributing in PRs that add lints is _not_ the lint queries, the test cases, etc. It is your expertise on that particular facet of the SemVer rules. It happens to be distilled into a lint file with its query and error message etc. but that's just the packaging.

The value is realized when users can say "ah, I would have made a terrible mistake here, but `cargo-semver-checks` caught it and explained it to me in a way that I understand the problem and its consequences." This doesn't happen if the lint query is wrong. It also doesn't happen if the error message or description. Or if any other metadata in the lint is wrong, like if we mis-categorized a lint as `Minor` instead of `Major`.

So please dig into the problem space, win that expertise, and contribute it as part of the lint as a whole. Otherwise, our users don't get any value — in fact, if we teach them something plainly incorrect, they may even be worse off overall!

obi1kenobi: Coming back to this specific instance, the user would not be able to use the static in the same way, since it isn't mutable. But I wouldn't expect the manner of "use" they can't do is "import it."

Obtaining the necessary expertise here is figuring out what manner of "use" is denied to the user that depended on a mutable static that became no longer mutable.

arpity22: I looked into the compile error and it seems compile error is only caused by assigning to the immutable static. So, I updated the error message and description accordingly.

P.S: I was looking through issue #767 and wanted to ask is there any documentation on lint optimization I should go through for future lints. Its nothing urgent and I doubt I would be able to go through them this week.

obi1kenobi: Great!

For lint optimization, there's not much one can do within the lints themselves. There's room for minor improvements like "order edges such that filtering happens as early as possible" but that won't make a big difference for cases like #767. Don't worry, you didn't cause that issue and neither did anyone else writing lints. It'll get resolved, and it won't require any changes to the lints themselves to do so.

The cause of that issue is a missing optimization in the Trustfall query engine adapter for rustdoc. If you are curious, you can read more about how I resolved a previous instance of the same issue here: https://predr.ag/blog/speeding-up-rust-semver-checking-by-over-2000x/

obi1kenobi (approved):
Awesome!

PR #776, opened 2024-05-10
Added lint trait_associated_type_now_doc_hidden

arpity22:
Lint for when an associated type in a public trait is now marked as #[doc(hidden)]

Issue mentioned in https://github.com/obi1kenobi/cargo-semver-checks/issues/578


test_crates/trait_associated_type_now_doc_hidden/old/src/lib.rs
```diff
@@ -0,0 +1,33 @@
+// Basic test case
+pub trait PUB_TRAIT_A {
+    type TYPE_A;
+    // Should not flag already #[doc(hidden)] type
+    #[doc(hidden)]
+    type DOC_HIDDEN_TYPE;
+}
```
obi1kenobi, line 7: nit: Could you rename the traits and associated types to the idomatic `CapitalizedSnakeCase` style instead of using `SCREAMING_SNAKE_CASE`? Only constants in Rust use that, so it's a bit disorienting to see it elsewhere.

obi1kenobi (commented):
It looks good except for a stylistic nitpick. If you could give that a quick look, I'd be happy to merge this immediately!

obi1kenobi (approved):
Awesome! 🎉

PR #781, opened 2024-05-27
fix: #735

jw013:
Remove id field from lint .ron's and set id based on file name passed to add_lints! macro instead.

Same idea as the original PR #739, just with the missing piece to set the id again.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -65,6 +65,8 @@ impl From<ReleaseType> for ActualSemverUpdate {
 #[non_exhaustive]
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct SemverQuery {
+    // this should be populated with the file name and not serialized
+    #[serde(skip)]
     pub id: String,
 
     pub(crate) human_readable_name: String,
@@ -98,8 +100,8 @@ pub struct SemverQuery {
 impl SemverQuery {
     pub fn all_queries() -> BTreeMap<String, SemverQuery> {
         let mut queries = BTreeMap::default();
-        for query_text in get_query_text_contents() {
-            let query: SemverQuery = ron::from_str(query_text).unwrap_or_else(|e| {
+        for (id, query_text) in get_queries() {
+            let mut query: SemverQuery = ron::from_str(query_text).unwrap_or_else(|e| {
                 panic!(
                     "\
 Failed to parse a query: {e}
@@ -108,6 +110,7 @@ Failed to parse a query: {e}
 ```"
                 );
             });
+            query.id = id.to_string();
             let id_conflict = queries.insert(query.id.clone(), query);
             assert!(id_conflict.is_none(), "{id_conflict:?}");
         }
@@ -462,10 +465,13 @@ macro_rules! add_lints {
             )*
         }
 
-        fn get_query_text_contents() -> Vec<&'static str> {
+        fn get_queries() -> Vec<(&'static str, &'static str)> {
             vec![
                 $(
-                    include_str!(concat!("lints/", stringify!($name), ".ron")),
+                    (
+                        stringify!($name),
+                        include_str!(concat!("lints/", stringify!($name), ".ron")),
+                    ),
                 )*
             ]
         }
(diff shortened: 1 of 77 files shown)


src/query.rs
```diff
@@ -65,6 +65,8 @@ impl From<ReleaseType> for ActualSemverUpdate {
 #[non_exhaustive]
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct SemverQuery {
+    // this should be populated with the file name and not serialized
+    #[serde(skip)]
```
obi1kenobi, line 69: I've never tried to combine `serde(skip)` of a non-`Option` field together with a `#[derive(Deserialize)]`.

Do you know what happens if someone tries to deserialize a `SemverQuery` type using that derived impl? I genuinely have no idea what the result would be, and I'm worried we might be leaving a footgun here.

src/query.rs
```diff
@@ -475,6 +481,8 @@ macro_rules! add_lints {
     }
 }
 
+// The following add_lints! invocation is programmatically edited by scripts/make_new_lint.sh
+// If you must manually edit it, be sure to read the "Requirements" comments in that script first
```
obi1kenobi, line 485: This comment is great and I'd love to merge it right away. If you'd like, I can merge it in its own PR if you were to open one for it.

obi1kenobi (commented):
Heads up that I'll be mostly offline for the next 2.5 weeks, so I might not be around very much. Wanted to let you know ahead of time so you don't think I'm ghosting you, because I really appreciate all your PRs!

src/query.rs
jw013: According to serde's [docs](https://serde.rs/field-attrs.html#skip), `#[serde(skip)]` should unconditionally fill it in with `Default::default()` (i.e. `String::new()`). It hadn't occurred to me before that this would constitute a change in behavior on a publicly visible field. 

Perhaps it would be better to use [`#[serde(default)]`](https://serde.rs/field-attrs.html#default) instead of skip. This would still deserialize the field if it exists just like before, the only observable change would be that if the `id` field is missing, deserialization attempts will now succeed whereas before they would return an error. It's conceivable that could break someone's workflow if they have their own tests relying on deserialization without id to fail but that seems unlikely to me.

Another option is to not touch the serde attributes on the struct at all. This would require putting the id field back in the lints/*.ron files, but we can use a throwaway placeholder value to indicate that it will be overwritten (with the filename derived id).

obi1kenobi: Taking a couple of steps back here: the underlying problem I think we're solving is the possibility that someone writing / editing a lint file might set the `id` incorrectly. Right?

The recommended workflow for making new lints is to use `./scripts/make_new_lint.sh`, which automatically fills in the `id` field. In that case, mistakes are unlikely: they'd require an explicit change to the `id` field.

If we apply either `serde` attribute, or override the field's value on deserialization, that seems to introduce quite a bit of non-obvious behavior and risks being a footgun.

An alternative option would be to add a test that enforces that all lint file names match the `id` string in the file. Normally, this would be annoying since new contributors might not know about this requirement or about the test. But we expect new lints to be created using the above workflow that automatically sets this field correctly, so this test should in practice never fail for new contributors _unless_ they mistakenly edited the `id` field by hand.

I'm inclined to think that adding such a test is the lesser inconvenience here, rather than a fancier solution with `serde` attributes or other magic that might backfire.

What do you think?

obi1kenobi:
Unrelated to this PR: I really appreciate your willingness to dig in and make substantive changes and improvements to `cargo-semver-checks`!

So far I've been a bit hesitant to create issues for more ambitious upgrades to `cargo-semver-checks` on the issue tracker. They are often places where code quality is crucial or else the long-term maintenance burden might be severe.

Your work so far has convinced me that I can trust your thoughtfulness and code quality, and if you're up for it, I'd be interested in working together with you more closely.

Is that something you might be interested in? I have some longer-term ideas on how `cargo-semver-checks` can be improved, and I'm sure you have ideas as well. Perhaps we should grab a virtual coffee and chat?

jw013:
> Is that something you might be interested in? I have some longer-term ideas on how `cargo-semver-checks` can be improved, and I'm sure you have ideas as well. Perhaps we should grab a virtual coffee and chat?

Sure, sounds good to me.

obi1kenobi:
Awesome! Let's move this chat off of GitHub — could you send me a note at the email [here](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/Cargo.toml#L5) and we can set up a time to chat?

jw013:
Sent you an email.

Closing PR as superseded by #783

PR #783, opened 2024-05-27
Check deserialized query id matches expected id

### LINKED ISSUE #735: Consider either testing that lint file names match their `id` fields, or removing the field entirely.

jw013:
Alternative approach to resolve #735 by testing for id consistency

(continuing discussion from #781)

My original line of thinking was that having the name of a lint repeated in the file name and contents and requiring them to match seems like a violation of DRY/"Don't Repeat Yourself", so it would be better to only store the name in one place. Comparing the two approaches though, I have to agree the checking approach does seem much simpler.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -98,7 +98,7 @@ pub struct SemverQuery {
 impl SemverQuery {
     pub fn all_queries() -> BTreeMap<String, SemverQuery> {
         let mut queries = BTreeMap::default();
-        for query_text in get_query_text_contents() {
+        for (id, query_text) in get_queries() {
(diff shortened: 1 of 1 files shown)


obi1kenobi (approved):

obi1kenobi:
Thanks!

PR #788, opened 2024-06-05
add query override data structures

suaviloquence:
adds the new members:
- `QueryOverride` to configure overridden default values for a lint
- `OverrideMap` (type alias) for a mapping of lint ids to `QueryOverride` structs.
- `OverrideStack`, a stack of `OverrideMap`s to store multiple levels of overrides, where items higher in the stack have higher precedence over lower ones when both are set. 

I kind of struggled in choosing names for these, especially `OverrideMap`, so if you have any ideas for that which better represent what it is, please let me know, thanks!

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeMap;
+use std::{collections::BTreeMap, sync::Arc};
 
 use serde::{Deserialize, Serialize};
 use trustfall::TransparentValue;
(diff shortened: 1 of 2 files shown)


src/query.rs
```diff
@@ -141,9 +141,80 @@ Failed to parse a query: {e}
...
+    /// the effective lint level.
+    pub lint_level: Option<LintLevel>,
+}
+
+/// A mapping of lint ids to configured values that override that lint's defaults.
+pub type OverrideMap = BTreeMap<String, QueryOverride>;
+
+/// Stores a stack of [`OverrideMap`] references such that items towards the top of
+/// the stack have *higher* precedence and override items lower in the stack if both
+/// are set.
```
obi1kenobi, line 165: Let's define what "top of the stack" is relative to our backing `Vec`.

I'm also not sure what `if both are set` refers to. Presumably the overriding behavior is only when two maps specify the same key, right? That would be my assumption if we don't say anything, so saying `if both are set` is making me think the answer is something more surprising instead..?
```suggestion
/// Stores a stack of [`OverrideMap`] references such that items towards the top of
/// the stack (later in the backing `Vec`) have *higher* precedence and override 
/// items lower in the stack.
```

src/query.rs
```diff
@@ -141,9 +141,80 @@ Failed to parse a query: {e}
...
+    /// Creates a new, empty [`OverrideStack`] instance.
+    #[must_use]
+    pub fn new() -> Self {
+        Self(Vec::new())
+    }
+
+    /// Inserts the given element at the top of the stack.
+    ///
+    /// The inserted overrides will take precedence over any lower item in the stack,
+    /// if both are set.
```
obi1kenobi, line 179: The `if both are set` is confusing again. The `both` seems to be referring to `items` or `overrides`, but it's unclear how they can be "set".

obi1kenobi (commented):
The data structures seem very sensible and are quite well-documented. Nicely done!

I got a bit tripped up on one of the phrases in the docs. I'm happy to let you take a look and merge immediately after, or even merge now and trust you to make the right tweaks in a subsequent PR if you'd prefer that. Your call!

src/query.rs
suaviloquence: Yes.  I wasn't sure to phrase this succinctly, but precedence happens when, for a given key in `{required_update, lint_level}` and a given lint `id`, if both maps have an entry for the lint with id `id` and `override.key` is not `None` for both of them, the override found at the top of the stack will take precedence.

```suggestion
/// Stores a stack of [`OverrideMap`] references such that items towards the top of
/// the stack (later in the backing `Vec`) have *higher* precedence and override items lower in the stack.
/// That is, when an override is set and not `None` for a given lint in multiple maps in the stack, the value
/// at the top of the stack will be used to calculate the effective lint level or required version update.  
```

Does this change communicate it more clearly?

src/query.rs
suaviloquence:
```suggestion
    /// if both maps have a not-`None` entry for a given lint.
```

Does this make sense?

src/query.rs
obi1kenobi: Perfect!

obi1kenobi (approved):

Issue #462, opened 2023-05-30
can fail when projects of same name are in a subdirectory


ehuss:
### Steps to reproduce the bug with the above code

1. `git clone https://github.com/rust-lang/mdbook.git`
2. `cd mdbook`
3. `git worktree add mdbook-0.2 v0.2.0`
4. `cargo semver-checks check-release`

### Actual Behaviour

    Updating index
     Parsing mdbook v0.2.0 (current)
Error: `cargo metadata` exited with an error:     Updating crates.io index
error: failed to select a version for the requirement `ammonia = "^1.1"`
candidate versions found which didn't match: 3.3.0, 3.2.1, 3.2.0, ...
location searched: crates.io index
required by package `mdbook v0.2.0 (/Users/eric/Temp/z10/mdbook/mdbook-0.2)`
    ... which satisfies path dependency `mdbook` of package `rustdoc v0.0.0 (/Users/eric/Temp/z10/mdbook/target/semver-checks/local-mdbook-0_2_0)`
perhaps a crate was updated and forgotten to be re-vendored?

### Expected Behaviour

It should run the semver checks for the package in the current directory. I would not expect it to go looking outside of the current directory for the project. It should follow the same behavior as cargo for finding the package to use.

(I likely don't know why it is walking the directory, but it is not clear to me why it would behave differently from other cargo commands.)

#### Software version

cargo-semver-checks 0.20.1

#### Operating system

macOS 13.4 (Darwin 22.5.0)

#### Command-line

```bash
/Users/eric/.cargo/bin/cargo-semver-checks semver-checks --bugreport
```

#### cargo version

```
> cargo -V
cargo 1.71.0-nightly (09276c703 2023-05-16)
```

#### Compile time information

- Profile: release
- Target triple: aarch64-apple-darwin
- Family: unix
- OS: macos
- Architecture: aarch64
- Pointer width: 64
- Endian: little
- CPU features: aes,crc,dit,dotprod,dpb,dpb2,fcma,fhm,flagm,fp16,frintts,jsconv,lor,lse,neon,paca,pacg,pan,pmuv3,ras,rcpc,rcpc2,rdm,sb,sha2,sha3,ssbs,v8.1a,v8.2a,v8.3a,v8.4a,vh
- Host: aarch64-apple-darwin

obi1kenobi:
Thanks for the bug report.

[...]

PR #799, opened 2024-06-19
read package/workspace config from Cargo manifest

suaviloquence:
Deserializes lint table from the relevant `[workspace.metadata]` and `[package.metadata]` tables, if they exist.

Notes:
- this is only possible when we have access to the `Cargo.toml` manifest and run the `cargo metadata` command - that is, we won't be able to configure lint levels when the rustdoc source is `Rustdoc | Revision | VersionFromRegistry` as of current without something like CLI flags:
  - [here's](https://github.com/suaviloquence/cargo-semver-checks/commit/9628e96b12cdb105e3d328aa57722be85b3aa3be) a shim of how I plan to pass this to the `run_check_release` function. takeaway is that when we don't have the manifest to read the metadata, we just pass an empty `OverrideStack` as overrides
- it's hard to test that we are generating the correct overrides at this state without the proposed new API.  Once we actually *use* the overrides, we can test whether they are applied correctly by parsing the output
  - [here](https://github.com/obi1kenobi/cargo-semver-checks/commit/cb853e694f4ba0acb370fddb51cd22e858801b4b)'s the beginnings of a test crate, although I just printed the passed overrides for now:
    <details>
    <summary>test output</summary>
    run with `cargo r -- semver-checks --baseline-root test_crates/workspace-overrides/baseline --manifest-path test_crates/workspace-overrides/current/pkg/Cargo.toml  -vvv`

    `overrides: OverrideStack([{"function_missing": QueryOverride { required_update: Some(Minor), lint_level: Some(Warn) }, "module_missing": QueryOverride { required_update: None, lint_level: Some(Allow) }}, {"function_missing": QueryOverride { required_update: None, lint_level: Some(Deny) }, "module_missing": QueryOverride { required_update: None, lint_level: Some(Warn) }}])`
       </details>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -486,8 +486,8 @@ note: skipped the following crates since they have no library target: {skipped}"
                 }
 
                 let workspace_overrides =
-                    manifest::deserialize_lint_table(metadata.workspace_metadata.clone())
-                        .context("[workspace.metadata.cargo-semver-checks] table is incorrect")?
+                    manifest::deserialize_lint_table(&metadata.workspace_metadata)
+                        .context("[workspace.metadata.cargo-semver-checks] table is invalid")?
                         .map(Arc::new);
 
                 selected
@@ -513,11 +513,12 @@ note: skipped the following crates since they have no library target: {skipped}"
                             Ok((crate_name.clone(), None))
                         } else {
                             let package_overrides =
-                                manifest::deserialize_lint_table(selected.metadata.clone())
+                                manifest::deserialize_lint_table(&selected.metadata)
                                     .with_context(|| {
                                         format!(
-                                    "{} [package.metadata.cargo-semver-checks] table is incorrect",
-                                    selected.name
+                                    "package `{}`'s [package.metadata.cargo-semver-checks] table is invalid (at {})",
+                                    selected.name,
+                                    selected.manifest_path,
                                 )
                                     })?;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,6 +22,7 @@ use trustfall_rustdoc::{load_rustdoc, VersionedCrate};
 use rustdoc_cmd::RustdocCommand;
 use std::collections::{BTreeMap, HashSet};
 use std::path::{Path, PathBuf};
+use std::sync::Arc;
 use std::time::Instant;
 
 pub use config::GlobalConfig;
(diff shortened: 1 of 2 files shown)


src/manifest.rs
```diff
@@ -113,6 +113,16 @@ impl From<OverrideConfig> for QueryOverride {
     }
 }
 
+/// Helper function to deserialize an optional lint table from a [`serde_json::Value`]
+/// into a [`OverrideMap`].  Returns an `Err` if the `cargo-semver-checks` table is present
+/// but invalid.  Returns `Ok(None)` if the table is not present.
+pub(crate) fn deserialize_lint_table(
+    metadata: serde_json::Value,
+) -> anyhow::Result<Option<OverrideMap>> {
+    let table: Option<LintTable> = serde_json::from_value(metadata)?;
```
obi1kenobi, line 122: We should be able to avoid the `.clone()` when calling this function by making it accept `&Value` and then using the trick described in this post: https://users.rust-lang.org/t/serde-from-value-moves-my-value-but-i-want-i-returned-to-me-on-err/69621

src/lib.rs
```diff
@@ -484,6 +485,11 @@ note: skipped the following crates since they have no library target: {skipped}"
                     );
                 }
 
+                let workspace_overrides =
+                    manifest::deserialize_lint_table(metadata.workspace_metadata.clone())
+                        .context("[workspace.metadata.cargo-semver-checks] table is incorrect")?
```
obi1kenobi, line 490: nit: word choice, here and below as well
```suggestion
                        .context("[workspace.metadata.cargo-semver-checks] table is invalid")?
```

src/lib.rs
```diff
@@ -506,6 +512,23 @@ note: skipped the following crates since they have no library target: {skipped}"
                             })?;
                             Ok((crate_name.clone(), None))
                         } else {
+                            let package_overrides =
+                                manifest::deserialize_lint_table(selected.metadata.clone())
+                                    .with_context(|| {
+                                        format!(
+                                    "{} [package.metadata.cargo-semver-checks] table is incorrect",
```
obi1kenobi, line 519: Idea: do we by any chance know the path to the `Cargo.toml` file here? If it isn't too difficult, it might be useful to not just say which crate had the invalid table but also give a path to it.

This is because in some cases, projects may contain more than one crate by the same name — for example, see #462. In such cases, knowing the `Cargo.toml` path would be very helpful to the user.

If it's too hard to do now, we can just open a `C-enhancement` issue for the idea and tackle it another time.

obi1kenobi (commented):
A couple of minor nits, looks good overall — should be able to merge on the next pass.

obi1kenobi (approved):

obi1kenobi:
Awesome! 🚀

PR #800, opened 2024-06-21
use configured required update version

suaviloquence:
* pass our deserialized overrides to the `run_check_release` function
* use the `effective_required_update` function in place of `query.required_update` whenever we previously accessed it to use the configured value
* add the beginnings of an integration test which tests precedence of workspace/package overrides (right now, it only tests required update, not lint level)

note: in `lib.rs:460`, when we don't have access to the current manifest, we can't read config at all, so we pass in an empty stack.  If we were to read CLI config, we could store that in `GlobalConfig` and put that at the bottom of the stack for both situations we pass overrides


test_crates/workspace_overrides/current/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[workspace]
+resolver = "2"
+members = ["pkg"]
+
+[workspace.metadata.cargo-semver-checks]
```
obi1kenobi, line 5: Might we want to add another layer of indirection here, like `workspace.metadata.cargo-semver-checks.lints`? With the current design, our entire metadata entry is reserved for lints and we can't add any other kind of config data at all.

tests/workspace_lint_config.rs
```diff
@@ -0,0 +1,26 @@
...
+    // right now because that is all that is currently implemented.
+    let mut cmd = Command::cargo_bin("cargo-semver-checks")
+        .expect("cargo semver-checks command should exist");
+
+    cmd.current_dir("test_crates/workspace_overrides").args([
+        "semver-checks",
+        "--baseline-root",
+        "baseline",
+        "--manifest-path",
+        "current/pkg/Cargo.toml",
```
obi1kenobi, line 15: Could you either make sure to either make a new subdirectory for all test cases for package/workspace config testing, or follow the existing norms in the directory where you're adding the test crates?

Using `baseline/current` instead of `old/new` will probably break some of the tooling we use, such as this script for example: https://github.com/obi1kenobi/cargo-semver-checks/blob/f8f89ecfa7f1e60b343e7b07beaf415393f244f8/scripts/regenerate_test_rustdocs.sh#L46-L48

tests/workspace_lint_config.rs
```diff
@@ -0,0 +1,26 @@
...
+    ]);
+
+    let assert = cmd.assert();
+
+    // 1 minor: function_missing, overridden in workspace and not overridden in package
+    // 1 major: module_missing, overridden in workspace (minor) then overridden in package (major)
+    //
```
obi1kenobi, line 22: Is this empty comment line intentional?

test_crates/workspace_overrides/baseline/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[workspace]
+resolver = "2"
+members = ["pkg"]
+
+# [workspace.metadata.cargo-semver-checks] should only be read from the
+# current Cargo.toml (this is the baseline)
```
obi1kenobi, line 6: This would be a great test case to add in a future PR!

test_crates/workspace_overrides/current/pkg/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+// pub fn function_missing() {}
+// pub mod module_missing {}
```
obi1kenobi, line 2: I'm generally not a fan of commented-out code by itself, without any explanation or context.

Consider adding a brief doc comment explanation to the `lib.rs` in both `current` and `baseline` that describes the test case: what it tries to catch and how, and mention why this code is commented out here.

That will be super useful in the future, if this test case starts failing or needs to be modified by a future PR. We certainly won't remember why this code is commented out, and it's much easier to take 30s to write down that context now than to spend half an hour reconstructing it from scratch later.

obi1kenobi (commented):
The implementation portion looks good. I have a couple of questions and concerns about some of the test setup. Nothing major, should be easy to sort out.

tests/workspace_lint_config.rs
suaviloquence: We can't use the `regenerate_test_rustdocs.sh` because we are passing the manifest path to `cargo-semver-checks`, so we don't want to generate rustdoc for these crates, which is why I didn't use new and old.  It does make more sense to move them into a subdirectory, though.

test_crates/workspace_overrides/current/Cargo.toml
suaviloquence: Sounds good.  I can open a separate PR for this.

tests/workspace_lint_config.rs
suaviloquence: Fixed, thanks!

tests/workspace_lint_config.rs
```diff
@@ -0,0 +1,26 @@
...
+    let mut cmd = Command::cargo_bin("cargo-semver-checks")
+        .expect("cargo semver-checks command should exist");
+
+    cmd.current_dir("test_crates/manifest_tests/workspace_overrides")
+        .args([
+            "semver-checks",
+            "--baseline-root",
+            "old", // should be the workspace root, not package root
+            "--manifest-path",
+            "new/pkg/Cargo.toml",
```
obi1kenobi, line 16: Possible other test: pass `--manifest-path new/Cargo.toml` -- the workspace manifest instead of the individual crate manifest -- and assert that the behavior is the same.

We want to make sure that crates are scanned with _their crate's_ settings, even when the scan is triggered at workspace level.

This is fine to add in a subsequent PR if that's easier.

tests/workspace_lint_config.rs
```diff
@@ -0,0 +1,26 @@
...
+        ]);
+
+    let assert = cmd.assert();
+
+    // 1 minor: function_missing, overridden in workspace and not overridden in package
+    // 1 major: module_missing, overridden in workspace (minor) then overridden in package (major)
```
obi1kenobi, line 22: Can you assert that these are indeed the minor and major checks that got triggered, and not e.g. vice versa? The current check could easily pass incorrectly, which would be a painful bug to track down.

In general, when you find yourself using a comment to describe something that should be true for a piece of code, assert it instead. Either the comment is accurate and the assert will pass, or it isn't and we need to know about it. The comment by itself risks being misleading: false, but pretending to be true. That tends to lead to lots of painful debugging.

test_crates/manifest_tests/README.md
```diff
@@ -0,0 +1,3 @@
+This subdirectory is for tests that use the Cargo.toml manifest as part of
+the test (for example, to test declared lint config), so
+`regenerate_test_rustdocs.sh` will not be run on test crates.
```
obi1kenobi, line 3: I assume you confirmed that `regenerate_test_rustdocs.sh` on this branch succeeds without errors, continues to generate all the expected rustdoc files, and does not run on `manifest_tests`?

I think this is probably true, but best to be sure!

obi1kenobi (commented):
If you could tweak the `workspace_lint_config.rs` check to be a bit more stringent, I feel good about the rest of this change and would be happy to merge it so we can move forward.

test_crates/manifest_tests/README.md
suaviloquence: :+1: It still works for me.  The script only goes through the top level of the test_crates directory to regenerate rustdocs, so by moving the new test into the manifest_tests directory, it doesn't run rustdoc.  And since manifest_tests doesn't contain child directories called `new` and `old`, it won't try to generate rustdocs for that folder either.

obi1kenobi (approved):

PR #804, opened 2024-06-25
Move lint configuration to [cargo-semver-checks.lints] table

suaviloquence:
Moves `[{package,workspace}.metadata.cargo-semver-checks]` to `[*.metadata.cargo-semver-checks.lints]` table, freeing up namespace if we want to add more config options later.

### DIFF
--- a/src/manifest.rs
+++ b/src/manifest.rs
@@ -64,7 +64,7 @@ pub(crate) fn get_project_dir_from_manifest_path(
 }
 
 /// A [package.metadata] or [workspace.metadata] table with
-/// `cargo-semver-checks` config entries stored in `config`
+/// `cargo-semver-checks` config entries stored in the `config` field below.
 #[derive(Debug, Clone, Deserialize)]
 pub(crate) struct MetadataTable {
     /// Holds the `cargo-semver-checks` table, if it is declared.
@@ -72,7 +72,8 @@ pub(crate) struct MetadataTable {
     pub(crate) config: Option<SemverChecksTable>,
 }
 
-/// A [cargo-semver-checks] config table in [package or [workspace.metadata].
+/// A `[cargo-semver-checks]` config table in `[package.metadata]` 
+/// or `[workspace.metadata]`.
 #[derive(Debug, Clone, Deserialize)]
 #[non_exhaustive]
 pub(crate) struct SemverChecksTable {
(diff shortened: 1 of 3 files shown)


src/manifest.rs
```diff
@@ -64,17 +64,28 @@ pub(crate) fn get_project_dir_from_manifest_path(
...
     #[serde(default, rename = "cargo-semver-checks")]
-    pub(crate) config: Option<BTreeMap<String, OverrideConfig>>,
+    pub(crate) config: Option<SemverChecksTable>,
 }
 
-impl LintTable {
-    pub fn into_overrides(self) -> Option<OverrideMap> {
-        self.config
-            .map(|config| config.into_iter().map(|(k, v)| (k, v.into())).collect())
+/// A [cargo-semver-checks] config table in [package or [workspace.metadata].
```
obi1kenobi, line 75:
```suggestion
/// A `[cargo-semver-checks]` config table in `[package.metadata]` 
/// or `[workspace.metadata]`.
```
Let's quote items in doc comments that are in square brackets. Otherwise rustdoc may try to introduce intra-doc links there, or readers may think they are intended to be such links (and are just broken).

src/manifest.rs
```diff
@@ -64,17 +64,28 @@ pub(crate) fn get_project_dir_from_manifest_path(
 }
 
 /// A [package.metadata] or [workspace.metadata] table with
-/// `cargo-semver-checks` lint entries stored in `config`
+/// `cargo-semver-checks` config entries stored in `config`
```
obi1kenobi, line 67: Let's clarify which of these refer to TOML file sections and which refer to Rust fields:
```suggestion
/// `cargo-semver-checks` config entries stored in the `config` field below.
```

src/manifest.rs
```diff
@@ -64,17 +64,28 @@ pub(crate) fn get_project_dir_from_manifest_path(
...
+/// A [cargo-semver-checks] config table in [package or [workspace.metadata].
+#[derive(Debug, Clone, Deserialize)]
+#[non_exhaustive]
+pub(crate) struct SemverChecksTable {
+    /// Holds the `lints` table, if it is declared.
+    pub(crate) lints: Option<LintTable>,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct LintTable(BTreeMap<String, OverrideConfig>);
```
obi1kenobi, line 84: Any particular reason this is `pub` not `pub(crate)` like the other types around it?

suaviloquence: Good catch, thanks. 
```suggestion
pub(crate) struct LintTable(BTreeMap<String, OverrideConfig>);
```

obi1kenobi (approved):

PR #805, opened 2024-06-26
Handle lint level in check_release

suaviloquence:
trying something new with smaller, annotated commits, lmk if it is helpful or not.

This PR handles lint level in `check_release`, especially differentiating warnings from errors.  It now skips running any checks marked as `allow`, prints errors and warnings differently, and the summary now reflects the required version bump (from error/deny checks) and the suggested version bump (from warning checks, if it is higher than the required bump).

<details>

<summary> example command output </summary>

```text
$ cargo r -- semver-checks --baseline-root old --manifest-path new/pkg/Cargo.toml -v
   Compiling cargo-semver-checks v0.32.0 (/home/m/dev/cargo-semver-checks)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4.72s
     Running `/home/m/dev/cargo-semver-checks/target/debug/cargo-semver-checks semver-checks --baseline-root old --manifest-path new/pkg/Cargo.toml -v`
     Parsing pkg v0.1.0 (current)
      Parsed [   0.472s] (current)
     Parsing pkg v0.1.0 (baseline)
      Parsed [   0.460s] (baseline)
    Checking pkg v0.1.0 -> v0.1.0 (no change)
    Starting 75 checks, 0 unnecessary on 16 threads
        PASS [   0.002s]       major            deny     auto_trait_impl_removed
        PASS [   0.001s]       major            deny     constructible_struct_adds_field
        PASS [   0.001s]       major            deny     constructible_struct_adds_private_field
        PASS [   0.001s]       major            deny     constructible_struct_changed_type
        PASS [   0.002s]       major            deny     derive_trait_impl_removed
        PASS [   0.001s]       major            deny     enum_marked_non_exhaustive
        PASS [   0.000s]       major            deny     enum_missing
        PASS [   0.001s]       minor            deny     enum_must_use_added
        PASS [   0.001s]       major            deny     enum_now_doc_hidden
        PASS [   0.002s]       major            deny     enum_repr_int_changed
        PASS [   0.001s]       major            deny     enum_repr_int_removed
        PASS [   0.001s]       major            deny     enum_repr_transparent_removed
        PASS [   0.001s]       major            deny     enum_struct_variant_field_added
        PASS [   0.001s]       major            deny     enum_struct_variant_field_missing
        PASS [   0.001s]       major            deny     enum_struct_variant_field_now_doc_hidden
        PASS [   0.001s]       major            deny     enum_tuple_variant_field_added
[...]

PR #808, opened 2024-07-05
annotate `unreachable!()` statements

suaviloquence:
from #805


suaviloquence:
this commit broke nightly, btw: https://github.com/rust-lang/rust/commit/7e8aac553e756b0eb03fe98e1a65ffc47836ec51

obi1kenobi (approved):
Awesome!

Thanks for the heads-up on the new rustdoc JSON format version. I think it's likely there's going to be another iteration on the format very soon, so I'll hold off for a few days to see if that gets merged and released quickly enough or not.

PR #809, opened 2024-07-05
add test for `allow` lint level

suaviloquence:
Tests that checks configured as `allow` are not run
and don't require a new semver update.


obi1kenobi (approved):
Great!

PR #811, opened 2024-07-10
Add `workspace` key to `[cargo-semver-checks.lints]` table

suaviloquence:
Doesn't do anything yet (changes are ready [here](https://github.com/suaviloquence/cargo-semver-checks/compare/opt-in-workspace-override...suaviloquence:cargo-semver-checks:implement-workspace-override), but wanted to split up PRs), but adds a `workspace` key to `[package.metadata.cargo-semver-checks.lints]` table for opting into workspace overrides.

I changed the internal crate API for deserializing the lints table a little bit to hold this key, but I can revert it/change it differently if it would be better.  With this change, we deserialize directly into an `OverrideMap` (so we no longer test the internal representation of it).  This still needs a few more changes to make it compliant with cargo's lint table, namely changing the `lint-level` key to `level` and making it required, as well as removing the shorthand for specifying version only, but those are easier changes and can be done in another PR.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -489,7 +489,7 @@ note: skipped the following crates since they have no library target: {skipped}"
                 let workspace_overrides =
                     manifest::deserialize_lint_table(&metadata.workspace_metadata)
                         .context("[workspace.metadata.cargo-semver-checks] table is invalid")?
-                        .map(Arc::new);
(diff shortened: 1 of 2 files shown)


obi1kenobi (approved):
Looks good!

PR #813, opened 2024-07-12
make specifying `workspace = false` a hard error

suaviloquence:
For [compatibility with cargo](https://rust-lang.zulipchat.com/#narrow/stream/421156-gsoc/topic/Project.3A.20Adding.20lint.20configuration.20to.20cargo-semver-checks/near/450316607).  This PR keeps the original API but validates that the user hasn't specified `workspace = false` while deserializing.

### DIFF
--- a/src/manifest.rs
+++ b/src/manifest.rs
@@ -165,7 +165,7 @@ where
     match option {
         Some(true) => Ok(true),
         None => Ok(false),
-        Some(false) => Err(serde::de::Error::custom("`lints.workspace = false` is not valid configuration.  Either set `lints.workspace = true` or omit the key entirely.")),
+        Some(false) => Err(serde::de::Error::custom("`lints.workspace = false` is not valid configuration. Either set `lints.workspace = true` or omit the key entirely.")),
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/manifest.rs
```diff
@@ -152,6 +152,23 @@ where
...
+fn deserialize_workspace_key<'de, D>(de: D) -> Result<bool, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let option = Option::<bool>::deserialize(de)?;
+
+    match option {
+        Some(true) => Ok(true),
+        None => Ok(false),
+        Some(false) => Err(serde::de::Error::custom("`lints.workspace = false` is not valid configuration.  Either set `lints.workspace = true` or omit the key entirely.")),
```
obi1kenobi, line 168: This isn't written down explicitly (and probably should be) but we generally single-space all our text, especially user-facing messages.
```suggestion
        Some(false) => Err(serde::de::Error::custom("`lints.workspace = false` is not valid configuration. Either set `lints.workspace = true` or omit the key entirely.")),
```

obi1kenobi (approved):
Very nice! Merging with a tweak to use single-spaced style in user-facing messages.

PR #814, opened 2024-07-12
Test that config is only read from new (i.e., current) manifest

suaviloquence:
Tests that config is only read from the new (current) manifest and not the old (baseline) manifest.


obi1kenobi (approved):
Great idea to add this, nicely done!

PR #818, opened 2024-07-15
Make lint table entries compatible with cargo `[lints]`

suaviloquence:
Removes `lint_name = "required-update"` shorthand and renames `lint-level` key to `level` in struct entry.

In cargo's `[lints]` table, the `level` key is required (for now).  We don't (in case someone wants to specify the required update and keep the default lint level), but this could cause problems if, eventually, we integrate with the lints table and cargo still keeps this behavior.  Do we want to make our `level` key required and then relax it later?

[context from zulip](https://rust-lang.zulipchat.com/#narrow/stream/421156-gsoc/topic/Project.3A.20Adding.20lint.20configuration.20to.20cargo-semver-checks/near/450027921)

### DIFF
--- a/src/manifest.rs
+++ b/src/manifest.rs
@@ -111,33 +111,37 @@ impl From<LintTable> for OverrideMap {
 
 /// Different valid representations of a [`QueryOverride`] in the Cargo.toml configuration table
 #[derive(Debug, Clone, Deserialize)]
-#[serde(untagged, rename_all = "kebab-case")]
+#[serde(untagged)]
 pub(crate) enum OverrideConfig {
     /// Specify members by name, e.g.
-    /// `lint_name = { lint-level = "deny", required-update = "major" }
+    /// `lint_name = { level = "deny", required-update = "major" }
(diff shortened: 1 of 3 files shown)


obi1kenobi:
I think it's fine if it's not required in our table and it remains required in cargo's.

If/when users migrate from our table to cargo's, they'll already be editing the manifest. If there are more than a couple of lint config lines to migrate, it should be quite easy to write a tool to automatically insert the "no-op" level setting for each lint.

I wouldn't make the ergonomics of our config table worse _today_ to avoid a hypothetical future problem that, even if it did happen, would be relatively painless to fix.

obi1kenobi (approved):
Excellent job on all the PRs lately. They've been super easy to review, since they are tightly scoped, well implemented and tested, and you've done a really good job providing appropriate context in the PR summary such that I have a very good idea of what I'm looking at.

Well done! It's a pleasure to have you here 🎉

suaviloquence:
Thank you!

PR #821, opened 2024-07-21
Inherent method now doc hidden

orblivion:
Doesn't work yet. Details inline.

### DIFF
--- a/src/lints/inherent_method_now_doc_hidden.ron
+++ b/src/lints/inherent_method_now_doc_hidden.ron
@@ -61,5 +61,5 @@ SemverQuery(
         "true": true
     },
     error_message: "A method or associated fn is now #[doc(hidden)], removing it from the crate's public API.",
-    per_result_error_template: Some("method or associated fn {{method_name}} in file {{span_filename}}:{{span_begin_line}}"),
+    per_result_error_template: Some("{{name}}::{{method_name}} in file {{span_filename}}:{{span_begin_line}}"),
 )
--- a/src/lints/inherent_method_now_doc_hidden.ron
+++ b/src/lints/inherent_method_now_doc_hidden.ron
@@ -60,7 +60,6 @@ SemverQuery(
         "public": "public",
         "true": true
     },
-
     error_message: "A method or associated fn is now #[doc(hidden)], removing it from the crate's public API.",
     per_result_error_template: Some("method or associated fn {{method_name}} in file {{span_filename}}:{{span_begin_line}}"),
 )
--- a/src/lints/inherent_method_now_doc_hidden.ron
+++ b/src/lints/inherent_method_now_doc_hidden.ron
@@ -1,0 +1,66 @@
+SemverQuery(
+    id: "inherent_method_now_doc_hidden",
+    human_readable_name: "inherent method #[doc(hidden)] added",
+    description: "A method or associated fn is now marked #[doc(hidden)] and is thus no longer part of the public API.",
+    required_update: Major,
+    lint_level: Deny,
+    reference_link: Some("https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#hidden"),
+    query: r#"
+    {
+        CrateDiff {
+            baseline {
+                item {
+                    ... on ImplOwner {
+                        visibility_limit @filter(op: "=", value: ["$public"]) @output
+
+                        importable_path {
+                            path @output @tag
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        inherent_impl {
+                            method {
+                                method_visibility: visibility_limit @filter(op: "=", value: ["$public"]) @output
+                                method_name: name @output @tag
+                                public_api_eligible @filter(op: "=", value: ["$true"])
+
+                                span_: span @optional {
+                                    filename @output
+                                    begin_line @output
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            current {
+                item {
+                    ... on ImplOwner {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        name @output
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        inherent_impl {
+                            method {
+                                method_visibility: visibility_limit @filter(op: "=", value: ["$public"])
+                                name @filter(op: "=", value: ["%method_name"])
+                                public_api_eligible @filter(op: "!=", value: ["$true"])
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true
+    },
+
+    error_message: "A method or associated fn is now #[doc(hidden)], removing it from the crate's public API.",
+    per_result_error_template: Some("method or associated fn {{method_name}} in file {{span_filename}}:{{span_begin_line}}"),
+)
--- a/test_crates/inherent_method_now_doc_hidden/new/src/doc_hidden_struct_inherent_method_now_doc_hidden.rs
+++ b/test_crates/inherent_method_now_doc_hidden/new/src/doc_hidden_struct_inherent_method_now_doc_hidden.rs
@@ -1,0 +1,17 @@
+// Tests false positives. If the struct is #[doc(hidden)], changing hidden state of methods should
+// have no effect.
+
+#[doc(hidden)]
+pub struct Foo;
+
+impl Foo {
+    #[doc(hidden)]
+    pub fn associated_fn(x: i64, y: i64) -> i64 {
+        x + y
+    }
+
+    #[doc(hidden)]
+    pub fn method(&self, x: i64) -> i64 {
+        x
+    }
+}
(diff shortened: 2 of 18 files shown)


src/lints/inherent_method_now_doc_hidden.ron
```diff
@@ -0,0 +1,68 @@
...
+                    ... on ImplOwner {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        name @output
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+			# TODO - why different from const?
```
orblivion, line 47: What I mean is, I'm not sure if I have to worry about all of this:

https://github.com/obi1kenobi/cargo-semver-checks/blob/main/src/lints/inherent_method_const_removed.ron#L48

test_crates/inherent_method_now_doc_hidden/new/src/doc_hidden_struct_inherent_method_now_doc_hidden.rs
```diff
@@ -0,0 +1,14 @@
+pub struct Foo;
+
+#[doc(hidden)]
```
orblivion, line 3: So, this false positive test fails! It's the one thing that fails.

I tried adding it to the `inherent_method_const_removed` test crate, lint and it also has no effect on output. Is the item `visibility_limit` restriction not working? Or did I do this wrong?

orblivion: The failure I get for this lint:

```
Differences (-expected|+actual):
             "method_visibility": String("public"),
             "name": String("Foo"),
             "path": List([
                 String("inherent_method_now_doc_hidden"),
+                String("doc_hidden_struct_inherent_method_now_doc_hidden"),
+                String("Foo"),
+            ]),
+            "span_begin_line": Uint64(5),
+            "span_filename": String("src/doc_hidden_struct_inherent_method_now_doc_hidden.rs"),
+        },
+        {
+            "method_name": String("method"),
+            "method_visibility": String("public"),
+            "name": String("Foo"),
+            "path": List([
+                String("inherent_method_now_doc_hidden"),
+                String("doc_hidden_struct_inherent_method_now_doc_hidden"),
+                String("Foo"),
+            ]),
+            "span_begin_line": Uint64(9),
+            "span_filename": String("src/doc_hidden_struct_inherent_method_now_doc_hidden.rs"),
+        },
+        {
+            "method_name": String("associated_fn"),
+            "method_visibility": String("public"),
+            "name": String("Foo"),
+            "path": List([
+                String("inherent_method_now_doc_hidden"),
                 String("enum_inherent_method_now_doc_hidden"),
                 String("Foo"),
             ]),
             "span_begin_line": Uint64(6),
```

orblivion: Maybe the struct was supposed to be doc hidden rather than the impl, but it still fails if I do that. Or if I do both, fwiw.

obi1kenobi: @DeluxeRaph and I were just chatting about this -- we aren't sure what the semantics of "there's `#[doc(hidden)]` on the _impl itself_" are like. To figure out what _should_ happen here, we need to see what HTML rustdoc generates for cases where the `impl` is hidden, and compare that to when its items are hidden.

I don't think any of our lints currently have tests for `#[doc(hidden)]` on the impl. We only test for `#[doc(hidden)]` on the struct/enum/union, or on the contents of the impl individually.

We don't currently have a lint for `#[doc(hidden)]` on the impl (and it isn't clear how we'd write one). I think it's fine to leave that as future work.

For the sake of this test, I would just move the `#[doc(hidden)]` onto the struct itself and remove it from the `impl` block. Based on your query, I think that will work correctly.

`visibility_limit` just refers to the struct's declaration directly: `pub struct Foo;` means the `visibility_limit` is `"public"`. It doesn't consider `#[doc(hidden)]` or anything else, and appears to be working correctly.

src/lints/inherent_method_now_doc_hidden.ron
obi1kenobi: No, you don't. This query looks correct as-is.

In the case of method removals, or removals of `const` on the method, the compiler may try to fall back to other definitions that match the name and/or const restriction as needed. Those other definitions may come from traits, which is why we need the logic you highlighted.

In this case, we don't care about that — we're looking for the exact same item, with the only change being that it's now `#[doc(hidden)]` and didn't use to be. The compiler doesn't avoid using `#[doc(hidden)]` items and wouldn't fall back to another definition of it, so this simpler logic is correct. The simplicity is why I recommended it as a starting lint :)

obi1kenobi (approved):
Nicely done, thanks for putting this together!

Everything looks good to me. I'd just move the one `#[doc(hidden)]` from the impl to the struct and update the expected outputs file, and this is good to merge.

Figuring out what to do when `impl` blocks themselves get hidden is a bigger issue, and we can figure it out separately from this PR. No reason to hold it up while we work that out.

test_crates/inherent_method_now_doc_hidden/new/src/doc_hidden_struct_inherent_method_now_doc_hidden.rs
orblivion: > Based on your query, I think that will work correctly.

Funny, I thought I had tried that and it failed (as I mentioned above). But I tried again and you're right. I must have forgotten to rebuild the docs the first time.

orblivion:
> I'd just move the one #[doc(hidden)] from the impl to the struct **and update the expected outputs file**, and this is good to merge.

Just to make sure everything is clear, the outputs file I initially committed was already what I was expecting without the false positive. Now that I moved the `#[doc(hidden)]`, the output matches the file.

orblivion:
One check failing. I think I overheard you saying something about an issue with CI?

I'm not sure if this repo enables squash merges so I'll push one more rebasing all the WIPs.

src/lints/inherent_method_now_doc_hidden.ron
```diff
@@ -0,0 +1,66 @@
...
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true
+    },
+
+    error_message: "A method or associated fn is now #[doc(hidden)], removing it from the crate's public API.",
+    per_result_error_template: Some("method or associated fn {{method_name}} in file {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 65: I'm a bit worried that the current wording might spill to more than one line, which can often look awkward in CLI output.

How about this tweak, which is a bit more compact while also listing the type on which the method/associated fn exists:
```suggestion
    per_result_error_template: Some("{{name}}::{{method_name}} in file {{span_filename}}:{{span_begin_line}}"),
```

test_crates/inherent_method_now_doc_hidden/old/src/doc_hidden_struct_inherent_method_now_doc_hidden.rs
```diff
@@ -0,0 +1,15 @@
+// Tests false positives. If the struct is #[doc(hidden)], changing hidden state of methods should
+// have no effect.
```
obi1kenobi, line 2: Another interesting test case variant, if you choose to add it, would be a further tweak of this one.

Instead of having the struct start off as `#[doc(hidden)]` originally in the `old` version, only make it `#[doc(hidden)]` in the `new` version alongside hiding its methods as well. That should _also_ not trigger this lint, though it will trigger the `struct_now_doc_hidden` lint instead.

obi1kenobi (approved):
This looks great! I came up with one more test case that could be worth adding if you're up for it, but I'm happy to add it separately if you'd prefer to merge this PR as-is instead.

Thanks for putting this together and testing it so thoughtfully!

obi1kenobi:
> One check failing. I think I overheard you saying something about an issue with CI?
> 
> I'm not sure if this repo enables squash merges so I'll push one more rebasing all the WIPs.

I'll resolve the nightly failure shortly, that's on me and not caused by this PR.

We squash-merge PRs by default, so don't worry about rewriting git history any particular way.

obi1kenobi:
I fixed the nightly build and merged `main` into your PR to make CI all green again.

We'll squash-merge the PR so don't worry about making it a linear history again.

test_crates/inherent_method_now_doc_hidden/old/src/doc_hidden_struct_inherent_method_now_doc_hidden.rs
orblivion: Just pushed, hopefully passes CI

obi1kenobi: Looks like it passed! I'm happy with this, and I can merge it as soon as you're happy with it too.

Thanks for putting it together!

orblivion: I'm ready to merge!

obi1kenobi:
Thanks for putting this together! 🚀

PR #824, opened 2024-07-22
Add explicit features for gix backend

ggwpez:
Update tame-index and integrate the latest change: https://github.com/EmbarkStudios/tame-index/pull/66

This merge request is adding two features to allow downstream developers to select which gix backend to use. Since Rust unifies the feature of a crate when building it, and gix has two *mutually exclusive* features, the only way here seems to propagate this oddity and expose it in the features of this crate as well.  

More concretely, this conflicts otherwise in our crate where we also pull in [cargo](https://github.com/rust-lang/cargo/blob/ea14e861c78cf5bb67e68ddeeea7ae9219a920a2/Cargo.toml#L49) which then always enables the curl backend.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ log = "0.4.17"
 # Note that `tame-index` and `gix` must be upgraded in lock-step to retain the same `gix`
 # minor version. Otherwise, one will compile `gix` two times in different minor versions.
 gix = { version = "0.63", default-features = false, features = ["max-performance", "revision"] }
-tame-index = { version = "0.12", features = ["git", "sparse"] }
+tame-index = { version = "0.12", features = ["sparse"] }
 
 human-panic = "1.0.3"
 bugreport = "0.5.0"
@@ -67,3 +67,13 @@ opt-level = 3
 debug-assertions = true
 overflow-checks = true
 codegen-units = 16
+
+[features]
+default = ["gix-reqwest"]
+
+# Gix has mutually exclusive features that are exposed through tame-index. Hard-coding either of
+# these features can lead to compile errors when another crate enables the other feature through
+# workspace feature unification. We therefore allow downstream users to choose which feature to
+# enable:
+gix-reqwest = ["tame-index/gix-reqwest"]
+gix-curl = ["tame-index/gix-curl"]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,16 +134,28 @@ jobs:
         with:
           workspaces: 'semver'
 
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -32,7 +32,7 @@ log = "0.4.17"
 # Note that `tame-index` and `gix` must be upgraded in lock-step to retain the same `gix`
 # minor version. Otherwise, one will compile `gix` two times in different minor versions.
 gix = { version = "0.63", default-features = false, features = ["max-performance", "revision"] }
-tame-index = { version = "0.12", features = ["git", "sparse"] }
```
obi1kenobi, line 35: Doesn't this PR end up removing the `git` feature that we depend on to support gitrev baselines?

Specifically, this CLI flag: ([docs](https://github.com/obi1kenobi/cargo-semver-checks?tab=readme-ov-file#does-the-crate-im-checking-have-to-be-published-on-cratesio))
```
--baseline-rev <REV>
    Git revision to lookup for a baseline
```

ggwpez: I checked the features here, and IIUC it should still be enabled by default, but will test it locally just in case:  
https://github.com/EmbarkStudios/tame-index/blob/9bc174aebc946d87dbb5a68ee88ccc21b6d2bc69/Cargo.toml#L15-L20

obi1kenobi: I just checked, and our CI already has a test for `--baseline-rev` which passed successfully.

I'm pushing a couple of minor fixes to your branch, and will then merge the PR & release in a new version shortly. Thanks for the PR, and if you're able to offer sponsorship funding, I'd really appreciate it!

obi1kenobi:
This functionality is now available in v0.33.0!

Lorak-mmk:
I think it is after this PR not possible to do `cargo install cargo-semver-checks --no-default-features` - it fails to compile. Is that expected? I can't find anything about this in release notes

obi1kenobi:
Oops! 😅

In retrospect, yes, that's a breaking change. I should have called it out in the release notes!

I didn't realize it was breaking and @ggwpez might not have noticed it either.

The irony of a breaking change linter having an unexpected breaking change is not lost on me 😅 This is why I want `cargo-semver-checks` lints that can check manifests: https://rust-lang.github.io/rust-project-goals/2024h2/cargo-semver-checks.html

obi1kenobi:
I have updated the release notes to mention this, and offer the updated command `cargo install cargo-semver-checks --no-default-features --features gix-reqwest` that reproduces the previous behavior.

Thanks for reporting this, and I'm sorry I broke your workflow!

Issue #633, opened 2024-01-13
New lints: Breaking changes related to `union` types


obi1kenobi:
Now that we've added `union` types to the schema we use for writing lints, we have a lot of `union`-related lints to implement!

Major lints:
- [x] pub union removed
  - make sure that converting a union to a struct or enum (`ImplOwner` type in our query schema) by the same name isn't flagged -- it's sometimes possible to change up types like this without a major breaking change
  - see the `struct_missing` lint's implementation, which has to do the same thing for the same reason
- [x] pub union's pub field removed / renamed (no longer present under its previous name, by whatever cause)
- [x] pub union is no longer `repr(C)`: #678 
- [x] pub union became `doc(hidden)` and is no longer public API (related to #578): #679
- [x] pub union's field became `doc(hidden)` and is no longer public API (also related to #578): #844
  - analogous lint for structs: https://github.com/obi1kenobi/cargo-semver-checks/blob/main/src/lints/struct_pub_field_now_doc_hidden.ron

Minor lints:
- [x] pub union `#[must_use]` added: #845
  - analogous lint for structs: https://github.com/obi1kenobi/cargo-semver-checks/blob/main/src/lints/struct_must_use_added.ron
  - remember to add the new lint to the README sections on configuring `#[must_use]` lints:
    - https://github.com/obi1kenobi/cargo-semver-checks?tab=readme-ov-file#common-configurations-make-must_use-lints-warn-only
    - https://github.com/obi1kenobi/cargo-semver-checks?tab=readme-ov-file#common-configurations-disable-must_use-lints-entirely

pksunkara:
> pub union is no longer `repr(C)`

@obi1kenobi Can I rename the existing lints? I am wondering if we can combine this to existing similar ones.

obi1kenobi:
You could combine it with the enum and the struct versions of that lint, and use the `ImplOwner` interface plus the `__typename` property to express them in one lint.

If you're interested in doing that, go for it. I don't have a strong preference between adding a separate lint or replacing them with one lint that works for `struct / enum / union` all in one.

obi1kenobi:
Nice job on this issue, everyone! We're two lints away from closing it completely! 🚀

obi1kenobi:
All done here! 🚀

PR #844, opened 2024-07-31
Add union pub field now doc hidden lint

PedroTurik:


src/lints/union_pub_field_now_doc_hidden.ron
```diff
@@ -0,0 +1,69 @@
...
+                    }
+                }
+            }
+            current {
+                item {
+                    ... on Union {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        union_name: name @output
+
+
```
obi1kenobi, line 34:
```suggestion

```

obi1kenobi (approved):
Excellent job! 🚀

obi1kenobi:
Great job here! We're now one lint away from completing all of #633. Would you care to write that one last lint so we can close it out?

After that, how would you feel about doing some work on the "opposite end" of the tool — near the CLI side? Suggesting it so that you start to get a more complete picture of `cargo-semver-checks` as a whole, instead of just working on one corner of it.

PR #845, opened 2024-07-31
Adds union_must_use_added lint

PedroTurik:


src/lints/union_must_use_added.ron
```diff
@@ -0,0 +1,65 @@
+SemverQuery(
+    id: "union_must_use_added",
+    human_readable_name: "union #[must_use] added",
+    description: "A union has been marked with #[must_use].",
+    required_update: Minor,
+    lint_level: Deny,
+
```
obi1kenobi, line 7: nit:
```suggestion
```

If I may recommend a tweak to your workflow that in my experience works wonders, consider the following.

When you're ready to open a pull request:
- open it as a draft first
- then, open it in the GitHub PR review UI, and _self-review_ it
- fix any issues you spot, then repeat
- once there are no more issues, mark it open and request a review

We're all so used to seeing the code in our editors that we often miss little things. But the change of UI will force your brain to see the code as if it were new, and you'll catch all sorts of issues that will then seem obvious — like this one.

If you catch and fix these issues before you ask for a review, that will massively speed up the rate at which your code will get merged while also requiring less maintainer time spent on reviews. It's a win-win!

src/lints/union_must_use_added.ron
```diff
@@ -0,0 +1,65 @@
...
+    }"#,
+    arguments: {
+        "public": "public",
+        "must_use": "must_use",
+        "true": true,
+        "zero": 0,
+    },
+    error_message: "A union is now #[must_use]. Downstream crates that did not use its value will get a compiler lint.",
+    per_result_error_template: Some("union {{name}} in {{span_filename}}:{{span_begin_line}}"),
+)
```
obi1kenobi, line 65:
```suggestion
)

```
Please make sure to tweak your editor configuration to add these trailing newlines at the end of file automatically. This ensures better compatibility with other CLI tools.

obi1kenobi (approved):
Nice job!

PR #846, opened 2024-08-02
Add snapshot testing with `insta`

suaviloquence:
Sets up infrastructure for snapshot testing with `insta`. Example coming in subsequent PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,8 @@ anstream = "0.6.13"
 assert_cmd = "2.0"
 similar-asserts = { version = "1.5.0", features = ["serde"] }
 predicates = "3.1.0"
+insta = { version = "1.39.0", features = ["yaml"] }
+serde = { version = "1.0.185", features = ["derive"] }
 
 # In dev and test profiles, compile all dependencies with optimizations enabled,
 # but still checking debug assertions and overflows.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,8 +50,7 @@ anstream = "0.6.13"
 assert_cmd = "2.0"
 similar-asserts = { version = "1.5.0", features = ["serde"] }
 predicates = "3.1.0"
-insta = { version = "1.39.0", features = ["yaml"] }
-serde = { version = "1.0.185", features = ["derive"] }
+insta-cmd = "0.6.0"
 
 # In dev and test profiles, compile all dependencies with optimizations enabled,
 # but still checking debug assertions and overflows.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ anstream = "0.6.13"
 assert_cmd = "2.0"
 similar-asserts = { version = "1.5.0", features = ["serde"] }
 predicates = "3.1.0"
-insta-cmd = "0.6.0"
+insta = { version = "1.39.0", features = ["ron"] }
 
 # In dev and test profiles, compile all dependencies with optimizations enabled,
 # but still checking debug assertions and overflows.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,6 +9,9 @@ mod rustdoc_gen;
 mod templating;
 mod util;
 
+#[cfg(test)]
+mod snapshot_tests;
+
 use anyhow::Context;
 use cargo_metadata::PackageId;
 use clap::ValueEnum;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,9 +9,6 @@ mod rustdoc_gen;
 mod templating;
 mod util;
 
-#[cfg(test)]
-mod snapshot_tests;
-
 use anyhow::Context;
 use cargo_metadata::PackageId;
 use clap::ValueEnum;
@@ -20,6 +17,7 @@ use itertools::Itertools;
 
 use check_release::run_check_release;
 use rustdoc_gen::CrateDataForRustdoc;
+use serde::Serialize;
 use trustfall_rustdoc::{load_rustdoc, VersionedCrate};
 
 use rustdoc_cmd::RustdocCommand;
@@ -36,7 +34,7 @@ pub use query::{
 
 /// Test a release for semver violations.
 #[non_exhaustive]
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Serialize)]
 pub struct Check {
     /// Which packages to analyze.
     scope: Scope,
@@ -54,7 +52,7 @@ pub struct Check {
 /// Affects which lints are executed.
 /// Non-exhaustive in case we want to add "pre-release" as an option in the future.
 #[non_exhaustive]
-#[derive(ValueEnum, Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(ValueEnum, Debug, Clone, Copy, PartialEq, Eq, Serialize)]
 pub enum ReleaseType {
     Major,
     Minor,
@@ -62,7 +60,7 @@ pub enum ReleaseType {
 }
 
 #[non_exhaustive]
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Serialize)]
 pub struct Rustdoc {
     source: RustdocSource,
 }
@@ -112,7 +110,7 @@ impl Rustdoc {
     }
 }
 
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Serialize)]
 enum RustdocSource {
     /// Path to the Rustdoc json file.
     /// Use this option when you have already generated the rustdoc file.
@@ -130,12 +128,12 @@ enum RustdocSource {
 }
 
 /// Which packages to analyze.
-#[derive(Default, Debug, PartialEq, Eq)]
+#[derive(Default, Debug, PartialEq, Eq, Serialize)]
 struct Scope {
     mode: ScopeMode,
 }
 
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Serialize)]
 enum ScopeMode {
     /// All packages except the excluded ones.
     DenyList(PackageSelection),
@@ -150,7 +148,7 @@ impl Default for ScopeMode {
 }
 
 #[non_exhaustive]
-#[derive(Default, Clone, Debug, PartialEq, Eq)]
+#[derive(Default, Clone, Debug, PartialEq, Eq, Serialize)]
 pub struct PackageSelection {
     selection: ScopeSelection,
     excluded_packages: Vec<String>,
@@ -171,7 +169,7 @@ impl PackageSelection {
 }
 
 #[non_exhaustive]
-#[derive(Default, Debug, PartialEq, Eq, Clone)]
+#[derive(Default, Debug, PartialEq, Eq, Clone, Serialize)]
 pub enum ScopeSelection {
     /// All packages in the workspace. Equivalent to `--workspace`.
     Workspace,
--- a/src/snapshot_tests.rs
+++ b/src/snapshot_tests.rs
@@ -131,7 +131,7 @@ impl fmt::Display for CommandResult {
 #[allow(dead_code)] // TODO
 fn assert_integration_test(test_name: &str, invocation: &[&str]) {
     // remove the backtrace environment variable, as this may cause non-
-    // reproducable snapshots.
+    // reproducible snapshots.
     std::env::remove_var("RUST_BACKTRACE");
 
     let stdout = StaticWriter::new();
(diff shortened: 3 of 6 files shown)


Cargo.toml
```diff
@@ -50,6 +50,8 @@ anstream = "0.6.13"
 assert_cmd = "2.0"
 similar-asserts = { version = "1.5.0", features = ["serde"] }
 predicates = "3.1.0"
+insta = { version = "1.39.0", features = ["yaml"] }
```
obi1kenobi, line 53: Personally, I somewhat prefer `ron` over `yaml` since it's more explicit and ergonomic around Rust (ADT-style) enums. What do you think?

tests/snapshot_tests.rs
```diff
@@ -0,0 +1,75 @@
+//! Snapshot tests of `cargo semver-checks` runs to ensure
```
obi1kenobi, line 1: Could we get away with keeping this inside a `#[cfg(test)]` within the crate itself, instead of as a separate module inside the top-level `tests`?

There's a substantial perf and ergonomics difference between the two: the former get compiled as part of the crate itself whereas the latter get compiled as a separate crate and are much slower to run and iterate on in my experience.

Cargo.toml
suaviloquence: That's fine with me, I was going off of the [docs](https://insta.rs/docs/serializers/):

> This serializer is recommended because YAML is human readable and excellent at diffing because it is line based.

I think ron would work (is it more source control friendly?) as well.

obi1kenobi: Ultimately, let's pick whatever looks the best for the job. I love that you opened a PR with a specific test case where we can look at the exact generated output and decide based on that.

tests/snapshot_tests.rs
suaviloquence: By "this" do you mean the tests themselves or just the infrastructure for snapshot testing?

obi1kenobi: The whole thing. If possible, ideally we wouldn't add anything to the top-level `tests/` at all.

suaviloquence: I can try but I think it depends on having the `cargo-semver-checks` binary already built, so I don't know if it would work.

obi1kenobi: If we could attach to "the layer underneath" then the only thing we lose out on testing is making sure clap correctly parsed our arguments. That doesn't seem _that_ valuable by comparison, especially if we get much better test perf and iteration speed.

If you don't think it's possible, I think this is fine to merge too. But it's worth a look at whether it's possible, just because the iteration speed would be different by multiples, not percentages.

suaviloquence: I think I was wrong - it seems to work anyway, so I moved it to lib tests.  Not sure if it will still be slower because it needs to build the binary? but if so I can do like you suggest.

obi1kenobi: This seems highly suspect.

`insta-cmd`'s [`get_cargo_bin` implementation](https://github.com/mitsuhiko/insta-cmd/blob/main/src/cargo.rs#L68-L73) uses [the `CARGO_BIN_EXE_<name>` env var](https://doc.rust-lang.org/cargo/reference/environment-variables.html) which is only set while building the binary or an integration test. I suspect the tests are falling through to the `insta-cmd` fallback path and reusing a prior (cached) binary which wouldn't be available in a clean `cargo test` run.

suaviloquence: Ah, that looks right, I deleted `target/debug/cargo-semver-checks` and it no longer works.  I think we can't use `insta_cmd` then and we have to make our own diff structure using `Check::check_release` (but that may be better in the long run with the speedups and new API).  Does that sound like a goal?

obi1kenobi:
Yeah, fair enough.

Better to have good integration tests now than perfect tests much later. We
can refactor and improve things down the line.

Thanks for checking and making sure!

suaviloquence:
Reverted for now.  I can try to work on a harness to lower it to lib tests, if it seems too complex maybe we can push that back for later.

obi1kenobi:
The harness might not be worth it right now, so feel free to leave it for later if it isn't doable quickly and easily.

suaviloquence:
This is a rough draft of the harness: [snapshot-lib-tests](https://github.com/suaviloquence/cargo-semver-checks/compare/add-no-lib-targets-test...suaviloquence:cargo-semver-checks:snapshot-lib-tests)

obi1kenobi:
That looks pretty decent, nice work! How do you feel about it?

suaviloquence:
I think it's mostly good, with a couple of disadvantages, but I think the speedup in development probably outweighs those.  Things to consider:

- It requires configuring the `Check` the same as `cargo-semver-checks`-binary does in [`impl From<CheckRelease> for Check`](https://github.com/obi1kenobi/cargo-semver-checks/blob/941e14c2c3ea62a0e9b87e0a5fcdfa85aec9ef1d/src/main.rs#L333)
- Similarly, the [exit code logic](https://github.com/obi1kenobi/cargo-semver-checks/blob/941e14c2c3ea62a0e9b87e0a5fcdfa85aec9ef1d/src/main.rs#L99) but that seems trivial.

I think the benefits probably outweigh the disadvantages though, so if you agree, I can port over the changes to this PR and move it to a lib-test.

obi1kenobi:
One option to address some of those disadvantages is to make `Check` serializable (at least in `cfg(test)`), then have two tests:
- An integration test that checks that a particular `cargo semver-checks` invocation produces the expected `Check` value.
- A lib test that checks the outcome of running a `Check` order to completion on a given workspace.

This is how Trustfall's own tests work as well. The idea there is that we write a query string, then record all intermediate results: the parse tree, the compiler's intermediate representation (IR), the execution steps ("trace"), as well as the ultimate outputs of the query. Each test stage then takes an input (either the query string, or the output of a prior stage), runs the code under test, and checks that the produced output matches the expected (snapshotted) value.

This is a pretty nice pattern, because it would make it easy to write new tests while also eliminating the risk that we misconfigure the `Check` in a way that doesn't match real life. But it's definitely a bit of work to set up, so your call on whether you want to go all the way with the extra effort or save it for a future iteration.

suaviloquence:
That's a good idea! I'm not sure how that would look in implementation, though, since we are testing the `Check` serialization on the library side and the integration/CLI tests on the binary side.  The only way I can think to do that right now is to duplicate the `Check` in both parts, and I'm not sure if `insta` could help with that.  Is that what you were thinking of or is there a different way to do this?

obi1kenobi:
Not sure what you mean, could you elaborate?

> Message ID: ***@***.***
> com>
>

What I was proposing might require moving a few types from the main.rs to
the lib.rs, which should be fine to do.

suaviloquence:
Oh, if we moved the clap structs to `lib.rs` (and the `Into<Check>` impl), that would work.  Is that what you meant?

obi1kenobi:
Yeah, let's do that for now.

Eventually we might want one more lib in between, but that's a later
problem :)

suaviloquence:
Actually, do you know if the test speedup from lib tests also applies to binary tests? (that is, `#[tests]` inside the `main.rs` (or linked modules), not integration tests in the `tests` folder).  If it does, I think it would be better to put them there.  If not, we have to make the CLI part of the library's public API (or #[doc(hide)] it at least), which seems like not the best solution (but it could work).

obi1kenobi:
I'm not sure I've ever tried putting tests in main.rs files. Try it and
see if it works — I trust your judgment!

suaviloquence:
Binary tests (timings from dev-desktop):
- rebuild tests after library change: 13.171s
- rebuild tests after binary change (in main.rs): 10.690s
- rebuild tests after change in test file: 12.512s
Library tests:
- rebuild tests after library change: 12.832s
- rebuild tests after change in main.rs: 0s (no need to rebuild)
- rebuild tests after change in test file: 13.242s

It looks like the timings are the same for library changes and test file changes, but for binary tests, we do have to rebuild the tests if the binary-side also changes (in main.rs).

The tradeoff for not having to spend that time is that we don't have to expose our CLI types in the public API.  What do you think?

src/lib.rs
```diff
@@ -33,7 +34,7 @@ pub use query::{
 
 /// Test a release for semver violations.
 #[non_exhaustive]
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Serialize)]
```
suaviloquence, line 37: One drawback to bin tests is that we need `Serialize` even in a non `#[cfg(test)]` environment.  I don't know how big a deal that is, though.

src/snapshot_tests.rs
```diff
@@ -0,0 +1,175 @@
...
+/// # Arguments
+///
+/// - `test_name` is the file prefix for snapshot tests to be saved in as
+///   `test_outputs/snapshot_tests/{test_name}-{"args" | "output"}.snap
+/// - `invocation` is a list of arguments of the command line invocation,
+///   starting with `["cargo", "semver-checks"]`.
+#[allow(dead_code)] // TODO
+fn assert_integration_test(test_name: &str, invocation: &[&str]) {
+    // remove the backtrace environment variable, as this may cause non-
+    // reproducable snapshots.
```
obi1kenobi, line 134:
```suggestion
    // reproducible snapshots.
```

src/snapshot_tests.rs
```diff
@@ -0,0 +1,175 @@
...
+
+    config.set_stdout(Box::new(stdout.clone()));
+    config.set_stderr(Box::new(stderr.clone()));
+    config.set_color_choice(false);
+
+    let check = Check::from(arguments.check_release);
+
+    let mut settings = insta::Settings::clone_current();
+    settings.set_snapshot_path("../test_outputs/snapshot_tests");
+    let _grd = settings.bind_to_scope();
```
obi1kenobi, line 152: `_grd` appears unused, are we using it for a side-effect from `bind_to_scope()`?

If so, would the usual `let _ = ...` pattern work instead?

It might also be a good idea to add a comment that explains what this does, for the benefit of readers not super familiar with `insta`.

src/lib.rs
obi1kenobi: Is that because `Serialize` is used by `cargo-insta` outside of `cfg(test)` somehow? Or is something else going on that I couldn't spot?

src/snapshot_tests.rs
```diff
@@ -0,0 +1,175 @@
...
+//! and `--baseline-root` arguments to specify the location of the current/baseline test crate path:
+//! in the `test_crates` directory if the test can use cached generated rustdoc files, or in the
+//! `test_crates/manifest_tests` directory if the test relies on `Cargo.toml` manifest files).
+//! Add a new function marked `#[test]` that calls [`assert_integration_test`] with
+//! the prefix (usually the function name) and the arguments to `cargo semver-checks`
+//!
+//! Then run `cargo test --bin cargo-semver-checks snapshot_tests`.  The new test should fail, as
+//! there is no snapshot to compare to.  Review the output with `cargo insta review`,
+//! and accept it when the captured behavior is correct. (see above if you can't use
+//! `cargo-insta`)
```
obi1kenobi, line 41: These docs are great! 💯

obi1kenobi (commented):
This looks really nice, well done! 🙌

Just a couple of minor questions, mostly from my unfamiliarity with `insta`, and I'm happy to merge this.

src/snapshot_tests.rs
suaviloquence: `bind_to_scope` returns a `Guard`, such that, as long as the guard is alive, the settings will be applied on the current thread (and are unapplied when the guard is `Drop`ped.)  We need to keep it alive for the whole function, so I don't think `let _ = ...` will work, but I will add a comment explaining this.

suaviloquence:
```suggestion
    // The `settings` are applied to the current thread as long as the returned
    //  drop guard  `_grd` is alive, so we use a `let` binding to keep it alive 
    // for the scope of the function.
    let _grd = settings.bind_to_scope();
```

src/lib.rs
suaviloquence: Since we're running the tests in the binary part of the crate, the library side is not compiled with `#[cfg(test)]` (when it's used as a dependency for this part), so it needs to be unconditionally `Serialize`.

src/snapshot_tests.rs
obi1kenobi: Interesting, does `let _ = ...` drop immediately?

I'm happy to merge it with your suggestion applied and figure out the `let _ = ...` drop rule separately.

obi1kenobi (approved):
I fixed the merge conflict I caused with another PR, and this is good to go!

src/snapshot_tests.rs
```diff
@@ -0,0 +1,178 @@
...
+    config.set_stdout(Box::new(stdout.clone()));
+    config.set_stderr(Box::new(stderr.clone()));
+    config.set_color_choice(false);
+
+    let check = Check::from(arguments.check_release);
+
+    let mut settings = insta::Settings::clone_current();
+    settings.set_snapshot_path("../test_outputs/snapshot_tests");
+    // The `settings` are applied to the current thread as long as the returned
+    //  drop guard  `_grd` is alive, so we use a `let` binding to keep it alive
```
obi1kenobi, line 153:
```suggestion
    // drop guard  `_grd` is alive, so we use a `let` binding to keep it alive
```

src/snapshot_tests.rs
suaviloquence: Yes it looks like it drops immediately - I did a quick test [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0b98f6d4662682de059f1f43000fcd0f)

obi1kenobi: Fascinating, TIL!

PR #847, opened 2024-08-02
add no lib targets test

suaviloquence:
Depends on #846.  Adds a test when there are no library targets in a workspace, which is (currently) an error.


suaviloquence:
Oops, I think I ran the test on nightly, which gave a backtrace. Will fix soon

suaviloquence:
Never mind, I read the diff wrong, need to figure out why 1.78 is giving different output.

suaviloquence:
(fixed in [0b7e2ee](https://github.com/obi1kenobi/cargo-semver-checks/pull/847/commits/0b7e2ee25f8fcab59875e2218ea869a7b4dc5b4e))

tests/snapshots/snapshot_tests__workspace_no_lib_targets_error.snap
```diff
@@ -0,0 +1,7 @@
+---
+source: tests/snapshot_tests.rs
+expression: "&output"
+---
+exit_code: 1
+stderr: "error: no crates with library targets selected, nothing to semver-check\nnote: only library targets contain an API surface that can be checked for semver\nnote: skipped the following crates since they have no library target: bin_only\n"
```
obi1kenobi, line 6: This isn't the most readable output: the explicit `\n` are quite annoying, as are the extremely long line lengths.

Consider individually snapshotting the stderr and stdout into separate files that are stored verbatim, like rustc does with its ui tests? For example: https://github.com/rust-lang/rust/blob/fd8d6fbe505ecf913f5e2ca590c69a7da2789879/tests/ui/nll/issue-50716.stderr

suaviloquence: Ron output:

```ron
CommandOutput(
  exit_code: Some(1),
  stderr: "error: no crates with library targets selected, nothing to semver-check\nnote: only library targets contain an API surface that can be checked for semver\nnote: skipped the following crates since they have no library target: bin_only\n",
  stdout: "",
)
```

I think we can use the default `insta::assert_snapshot!` on the stdout and stderr, which will use their `Display` impls, which should print new lines.

obi1kenobi: Consider doing a hybrid approach. Make the `CommandOutput` struct contain three fields: `stdout`, `stderr`, and "everything else" inside a single struct that can be ron-serialized and snapshotted that way.

Separately, consider looking at the "redactions" and "filters" features of `insta` to remove information that will be a source of non-determinism: https://insta.rs/docs/redactions/

suaviloquence: It looks like `insta` provides the `insta-cmd` crate which does exactly what we want.  I refactored to use that.

obi1kenobi: Wow, TIL! Nice find, that looks quite good ✨

suaviloquence:
[compare with `add-snapshot` branch](https://github.com/suaviloquence/cargo-semver-checks/compare/add-snapshot...suaviloquence:cargo-semver-checks:add-no-lib-targets-test)

src/snapshot_tests.rs
```diff
@@ -176,3 +175,23 @@ fn assert_integration_test(test_name: &str, invocation: &[&str]) {
 
     insta::assert_snapshot!(format!("{test_name}-output"), result);
 }
+
+/// [#163](https://github.com/obi1kenobi/cargo-semver-checks/issues/163)
+///
+/// Running `cargo semver-checks --workspace` on a workspace that has library
+/// targets should be an error.
```
obi1kenobi, line 182: typo, presumably?
```suggestion
/// Running `cargo semver-checks --workspace` on a workspace that
/// doesn't have any library targets should be an error.
```

test_crates/manifest_tests/no_lib_targets/new/new/Cargo.toml
obi1kenobi, line 1: Is the addition of files under `.../new/new/` intentional here?

obi1kenobi (commented):
A couple of nitpicks, broadly looks good.

obi1kenobi (approved):

PR #856, opened 2024-08-15
Allow generating test rustdoc JSON files without git dependency.

### LINKED ISSUE #855: `regenerate_test_rustdocs.sh` depends on Git
### Describe your use case

We use the GitHub archives of revisions to build `cargo-semver-checks`, and we want to run the tests in CI, which requires running `scripts/regenerate_test_rustdocs.sh`. Since the exported tarballs don’t have a `.git` directory, the `git rev-parse --show-toplevel` call fails.

Use a method of finding the source root that doesn’t depend on being run in a Git repository. There are various Bash constructions for finding the directory the currently‐running script is in, which should work fine for this. I suspect Cargo may also have a command that lets you find the workspace root, too.

We’ve previously been working around that by depending on Git for tests and initializing a blank repository so that the script can find the root. I have an open pull request to switch to patching the assignment in the script directly instead, but that’s not terribly elegant either.

### Additional Context

https://github.com/NixOS/nixpkgs/pull/334718

obi1kenobi:
Fixes #855.

### DIFF
--- a/scripts/regenerate_test_rustdocs.sh
+++ b/scripts/regenerate_test_rustdocs.sh
@@ -17,7 +17,13 @@ dir_is_newer_than_file() {
 
 export CARGO_TARGET_DIR=/tmp/test_crates
 RUSTDOC_OUTPUT_DIR="$CARGO_TARGET_DIR/doc"
-TOPLEVEL="$(git rev-parse --show-toplevel)"
+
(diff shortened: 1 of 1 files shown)


obi1kenobi:
No worries at all! I figured it was low enough risk so I went ahead and merged it :)

PR #859, opened 2024-08-15
Add test for when all members in a workspace are `publish = false`

suaviloquence:
Depends on #847; compare branches [here](https://github.com/suaviloquence/cargo-semver-checks/compare/add-no-lib-targets-test...suaviloquence:cargo-semver-checks:test-workspace-all-publish-false).

I don't know if this is the behavior that we *want* (exiting silently with exit code 0, not semver-checking `publish = false` targets), but it is the current behavior, so this PR adds a test to snapshot this behavior.


obi1kenobi (approved):
Awesome!

A neat variant we could try in a subsequent PR is this: "if the workspace contains a `publish=false` crate, but the `-p` option is used to specify it for checking explicitly, check it anyway despite `publish=false`"

suaviloquence:
> A neat variant we could try in a subsequent PR is this: "if the workspace contains a `publish=false` crate, but the `-p` option is used to specify it for checking explicitly, check it anyway despite `publish=false`"

I started to write this.  I found two different behaviors

- when called with `--workspace --package pkg`, the `--workspace` flag takes precedence and the package is not semver-checked
- when called with just `--package pkg` (but at the workspace root), the package is semver-checked.

Which (or both) of these were you thinking of adding?

obi1kenobi:
Oooh those are both good! It feels like we might also want to check the package in both cases, what do you think?

As an aside, this is _exactly_ the sort of edge case I was hoping this kind of testing capability would uncover, fix, and forever prevent regressions in.

PR #861, opened 2024-08-15
adds ids to text areas in bug reports

PedroTurik:
for auto github issues url generating in a future PR


obi1kenobi (approved):
Looks good!

Btw, if it makes the user experience better, I'm fine with having another issue template specifically for auto-generated bug reports. These templates are optimized for users going through the GitHub UIs reporting flow, and we don't have to try to put a square peg in a round hole if a custom template will be better.

PR #862, opened 2024-08-15
Bugreport auto issue

PedroTurik:
This PR automates the process of creating an issue and copying the output from --bugreport by generating a github new issue URL. It also includes build cargo build configuration

2 dependencies were added: urlencoding urlencoding = "2.1.3" and cargo-config2 = "0.1.26"

Old report:
image

New report:
image
image

New report with NO_COLOR env variable set
image

image

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -186,7 +186,6 @@ fn print_issue_url(config: &mut GlobalConfig) {
         }
     };
 
-
     writeln!(
         config.stdout(),
         "{bold_cyan}\
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,7 +2,7 @@
 
 use std::{env, path::PathBuf};
 
-use anstyle::{Color, Reset, Style};
+use cargo_config2::Config;
 use cargo_semver_checks::{
     GlobalConfig, PackageSelection, ReleaseType, Rustdoc, ScopeSelection, SemverQuery,
 };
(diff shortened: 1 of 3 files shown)


obi1kenobi:
Thanks for putting this together!

Can you put a little before & after showcase of the `--bugreport` output to make it easy to compare at a glance? It's a bit difficult to figure out what the user would see based on the code diff alone.

src/main.rs
```diff
@@ -143,6 +138,65 @@ fn configure_color(cli_choice: Option<clap::ColorChoice>) {
...
+
+    let bug_report_url = urlencoding::encode(&bug_report);
+
+    let cargo_config = match Config::load() {
+        Ok(c) => toml::to_string(&c).unwrap_or_else(|s| {
+            eprintln!("Error serializing cargo build configuration: {}", s);
+            String::default()
+        }),
+        Err(e) => {
+            eprintln!("Error loading cargo build configuration: {}", e);
```
obi1kenobi, line 171: The `eprintln!()` should probably be `writeln!(config.stderr(), ...)` now that we're plumbing the config through.

src/main.rs
```diff
@@ -143,6 +138,65 @@ fn configure_color(cli_choice: Option<clap::ColorChoice>) {
...
+fn print_issue_url(config: &mut GlobalConfig) {
+    use bugreport::{bugreport, collector::*, format::Markdown};
+    let other_bug_url: &str = "https://github.com/obi1kenobi/cargo-semver-checks/issues/new?labels=C-bug&template=3-bug-report.yml";
+
+    let bug_report = bugreport!()
+        .info(SoftwareVersion::default())
+        .info(OperatingSystem::default())
+        .info(CommandLine::default())
+        .info(CommandOutput::new("cargo version", "cargo", &["-V"]))
+        .info(CompileTimeInformation::default())
```
obi1kenobi, line 150: Is it possible to just `.clone()` here, so we can `.print()` once and `.format()` another time for the URL just to keep the code simpler?

Right now, we're mixing-and-matching formatting (e.g. on line 155), so if `bugreport` changes the output format in a future version, our output might look inconsistent.

src/main.rs
```diff
@@ -143,6 +138,65 @@ fn configure_color(cli_choice: Option<clap::ColorChoice>) {
...
+            .fg_color(Some(Color::Ansi(AnsiColor::Cyan))),
+        Reset
+    )
+    .expect("Failed to print bug report Cargo configuration to stdout");
+
+    let cargo_config_url: String = urlencoding::encode(&cargo_config).into_owned();
+
+    writeln!(
+        config.stdout(),
+        "{}Please file an Issue on github reporting your bug.\n\
```
obi1kenobi, line 190:
```suggestion
        "{}Please file an issue on GitHub reporting your bug.\n\
```

obi1kenobi (commented):
I like how this is shaping up! Just a couple of suggestions to polish it further for you to consider

obi1kenobi (commented):
Looks good! One last thing we need to confirm:

If you run the bug reporter code locally and click the link, does it open in GitHub? Are all the fields properly populated?

PedroTurik:
> Looks good! One last thing we need to confirm:
> 
> If you run the bug reporter code locally and click the link, does it open in GitHub? Are all the fields properly populated?

Yes

obi1kenobi:
Awesome, let's ship it!

PR #865, opened 2024-08-16
Forward verbosity and filter time out of snapshot tests

suaviloquence:
Improvements/fixes for the stdout/stderr of a snapshot test.

1. Forward verbosity argument from clap args into `GlobalConfig`
2. Filter `[   x.xxxs]` time strings from stdout and stderr for reproducible snapshots

<details>

<summary> Example output run on the `only_warnings` test crate </summary>

```
---
source: src/snapshot_tests.rs
expression: result
---
success: true
--- stdout ---

--- warning function_missing: pub fn removed or renamed ---

Description:
A publicly-visible function cannot be imported by its prior path. A `pub use` may have been removed, or the function itself may have been renamed or removed entirely.
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.34.0/src/lints/function_missing.ron

Failed in:
  function only_warnings::function_missing, previously in file /home/m/dev/cargo-semver-checks/test_crates/manifest_tests/only_warnings/old/src/lib.rs:4

--- stderr ---
     Parsing only-warnings v0.1.0 (current)
      Parsed [TIME] (current)
     Parsing only-warnings v0.1.0 (baseline)
      Parsed [TIME] (baseline)
    Checking only-warnings v0.1.0 -> v0.1.0 (no change)
     Checked [TIME] 80 checks: 79 pass, 0 fail, 1 warn, 0 skip

     Summary no semver update required
     Warning produced 1 major and 0 minor level warnings
             produced warnings suggest new major version
    Finished [TIME] only-warnings
```

</details>

This is technically two changes in a PR - I can break it up if you'd like, but both changes are very small.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1857,6 +1857,7 @@ dependencies = [
  "console",
  "lazy_static",
  "linked-hash-map",
+ "regex",
  "ron 0.7.1",
  "serde",
  "similar",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ anstream = "0.6.13"
 assert_cmd = "2.0"
 similar-asserts = { version = "1.5.0", features = ["serde"] }
 predicates = "3.1.0"
-insta = { version = "1.39.0", features = ["ron"] }
+insta = { version = "1.39.0", features = ["ron", "filters"] }
 fs-err = "2.11.0"
 
 # In dev and test profiles, compile all dependencies with optimizations enabled,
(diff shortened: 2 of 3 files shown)


obi1kenobi (approved):
This part might need additional filtering: `Checked [TIME] 80 checks: 79 pass, 0 fail, 1 warn, 0 skip`

Otherwise, each new added lint will break it since it'll increment the check count and the pass count. I think masking both of those out should be fine — tests care about the number of fail/warn/skip, not the total lint count.

It's fine to do this in a separate PR, I'll merge this since it's a good change already.

PR #867, opened 2024-08-16
Filter total # of checks and filenames in snapshots

suaviloquence:
Filters, for reproducibility:

- total number of checks run
- number of passed checks
- filenames in spans (not sure about the regex on this one, let me know what you think)

### DIFF
--- a/src/snapshot_tests.rs
+++ b/src/snapshot_tests.rs
@@ -151,6 +151,12 @@ fn assert_integration_test(test_name: &str, invocation: &[&str]) {
     settings.set_snapshot_path("../test_outputs/snapshot_tests");
     // Turn dynamic time strings like [  0.123s] into [TIME] for reproducibility.
     settings.add_filter(r"\[\s*[\d\.]+s\]", "[TIME]");
+    // Turn total number of checks into [TOTAL] to not fail when new lints are added.
+    settings.add_filter(r"\d+ checks", "[TOTAL] checks");
+    // Similarly, turn the number of passed checks to also not fail when new lints are added.
+    settings.add_filter(r"\d+ pass", "[PASS] pass");
+    // Redact filenames in spans when lints are triggered.
+    settings.add_filter(r"in file [^:]+:", "in file [PATH]:");
 
     // The `settings` are applied to the current thread as long as the returned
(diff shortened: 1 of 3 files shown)


src/snapshot_tests.rs
```diff
@@ -151,6 +151,12 @@ fn assert_integration_test(test_name: &str, invocation: &[&str]) {
     settings.set_snapshot_path("../test_outputs/snapshot_tests");
     // Turn dynamic time strings like [  0.123s] into [TIME] for reproducibility.
     settings.add_filter(r"\[\s*[\d\.]+s\]", "[TIME]");
+    // Turn total number of checks into [TOTAL] to not fail when new lints are added.
+    settings.add_filter(r"\d+ checks", "[TOTAL] checks");
+    // Similarly, turn the number of passed checks to also not fail when new lints are added.
+    settings.add_filter(r"\d+ pass", "[PASS] pass");
+    // Redact filenames in spans when lints are triggered.
+    settings.add_filter(r"in file [^:]+:", "in file [PATH]:");
```
obi1kenobi, line 159: I'm not sure about this filter 🤔 Could you walk me through it?

Right now it seems like it'll redact the name, but will leave:
- the line number
- the name of the type or function
- the number of instances of that lint

suaviloquence: Yes, my intention is to just redact the path of the file, as e.g., this was different in CI and locally for me.  With the same input, the line number, type name, etc. will all be deterministic and will be the same regardless of the execution environment, right? When that's the case, I think it makes sense to keep them in the snapshot.

obi1kenobi: Ah, I see. In that case, let's redact just the changed prefix. Here's one way to do it:

`let this_file = std::path::Path::new(file!()).canonicalize().expect("canonicalization failed");` will give us a canonical file path for the file where that line runs. We can then use `.parent()` calls to get to the repo root. Let's call this functionality `get_root_path()`.

We can then use `regex::escape(&get_root_path()).as_str()` to define a regex matcher and replace the root path with `[ROOT]` or something similarly obvious.

I haven't looked that closely into insta, so maybe there's an even better way. In the worst case, I think the above should work okay.

obi1kenobi (approved):
Excellent, ship it 🚀

Issue #870, opened 2024-08-18
New lints: Breaking changes in non-sealed trait items


obi1kenobi:
We can now query whether lints are sealed or not: https://github.com/obi1kenobi/trustfall-rustdoc-adapter/pull/343

```graphql
{
    Crate {
        item {
            ... on Trait {
                name @output
                sealed @output
            }
        }
    }
}
```
Many changes in non-sealed traits are breaking downstream, since any downstream implementations of that trait will need to be adjusted to match the trait item changes.

Querying the sealed status of traits was a prerequisite for a variety of SemVer major lints, which can now be implemented:
- [x] non-sealed trait added associated constant without default value: #875 
- [x] non-sealed trait added associated type without default value: #890
- [x] non-sealed trait removed the default value for an associated constant: #888
- [x] non-sealed trait removed the default value for an associated type: #889
- [x] trait newly became sealed, breaking any external implementors: #876
- [x] non-sealed trait added new supertraits (also tracked by #441): #892
- [x] non-sealed trait added method without default impl: #891
- [x] non-sealed trait removed a default impl for a trait method (also tracked by #294, claimed by @dmatos2012): #880

For guidance on how to write these lints, look at existing lints that check trait-related breakage. For example, here's the lint for "a non-sealed trait's method stopped being `unsafe`, so implementations must remove the `unsafe` keyword in their declarations": https://github.com/obi1kenobi/cargo-semver-checks/blob/661fcedaf8b1700959a61306433807f08edf457f/src/lints/trait_method_unsafe_removed.ron

Also make sure to check out our [contributing guide](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md), and take a look at what [prior lints' merged PRs looked like](https://github.com/obi1kenobi/cargo-semver-checks/pull/844) so you know what to expect. It's easier than you think!

dmatos2012:
Hi! I have checked the linked `PRs` and followed along and I would like start working on the item: 

- [ ]  trait newly became sealed (breaking any external implementors)

Thanks for the extremely well written out explanation, it really helps for newer contributors to get context of why/hows.

mrnossiom:
I would like to claim the following to start off
- [ ]  non-sealed trait added associated constant without default value

Is there a written convention on lints name? `non_sealed_trait_added_assoc_const_without_default` seems right to me yet quite lengthy.

obi1kenobi:
[...]

PR #873, opened 2024-08-19
fix: regenerate script toplevel resolution

mrnossiom:
Hi 👋🏻,

I wanted to check on #870. (EDIT: sorry for linking the issue without a real connection)

When loading the `regenerate_test_rustdocs.sh` script, it would just output `Generating rustdoc with: cargo 1.80.1 (376290515 2024-07-16)` and then exit promptly. Upon closer examination, the [`TOPLEVEL` variable](https://github.com/obi1kenobi/cargo-semver-checks/blob/661fcedaf8b1700959a61306433807f08edf457f/scripts/regenerate_test_rustdocs.sh#L26) would end up having two values, thus silently crashing the rest of the program.

Though my bash knowledge is limited, I think this is what is happening:

```bash
git rev-parse --show-toplevel 2>/dev/null || cd -- "$(dirname -- "${BASH_SOURCE[0]}" )" &>/dev/null && cd -- .. &>/dev/null && pwd
# is either (in git repo)
true || true && true && true # execs both git and pwd
false || true && true && true # execs only pwd
```

Wrapping the second part of the logical `OR` resolves the issue.

It seems that #856, introduced this and was recently committed.

### DIFF
--- a/scripts/regenerate_test_rustdocs.sh
+++ b/scripts/regenerate_test_rustdocs.sh
@@ -23,7 +23,7 @@ RUSTDOC_OUTPUT_DIR="$CARGO_TARGET_DIR/doc"
 # - Otherwise, navigate up relative to this script's path.
 #   This latter option is useful when building cargo-semver-checks for distribution with NixOS:
 #   https://github.com/obi1kenobi/cargo-semver-checks/issues/855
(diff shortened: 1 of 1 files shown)


obi1kenobi:
Yup, I definitely messed that up — thank you so much!

PR #874, opened 2024-08-20
`cargo update` to grab new rustdoc adapter with function body ability.

obi1kenobi:


obi1kenobi:
cc @mrnossiom after this merges, your new schema addition will be ready for use in lints!

PR #875, opened 2024-08-20
add lint `trait_associated_const_added`

mrnossiom:
Lint for "non-sealed trait added associated constant without default value"

See #870

This is my first `semver-checks` lint. I'm unsure how to proceed to check how a default value has disappeared for a given associated constant.

I think I need to check if the default value of the assoc const is still `null` for the `current` crate. Looking at other lints, I should be able to do that with `fold` and `filter`?

Could someone provide a little guidance, so I can better understand how diff queries works?

Thanks


obi1kenobi:
> a default value has disappeared for a given associated constant

Just to make sure, is this lint looking for "the const's default has disappeared" or "there's a new const that didn't exist before, and doesn't have a default"?

Your query is very close either way, and I'll be happy to point you in the right direction.

There are also a couple of filter operators that might be useful: `"is_null"` and `"is_not_null"`, which you can use like this:
```
property @filter(op: "is_null")
```

mrnossiom:
Lint is looking for "there's a new const that didn't exist before, and doesn't have a default".

How can check the addition of an assoc const?

obi1kenobi:
Try framing the query like this:
- in the new version, there exists a const without a default
- the old version has no const by that name on that trait

For the second part: "X does not exist" is a `@fold @transform(op: "count") @filter(op: "=", value: ["$zero"])` clause sequence on the edge that produces the `X`. We specifically want to say "the same trait exists but the named const doesn't exist" so be mindful of where those clauses go. This is important because we don't want the lint to trigger if the trait is new and didn't exist before — adding a new trait with a const is not a problem.

I think you already have the first part just about working in your PR.

So minimal changes needed — you're super close.

src/lints/trait_associated_const_added.ron
```diff
@@ -0,0 +1,64 @@
...
+    id: "trait_associated_const_added",
+    human_readable_name: "non-sealed trait added associated constant without default value",
+    description: "TODO",
+    required_update: Major,  // TODO
+    lint_level: Deny,  // TODO
+    reference_link: None,  // TODO
+    query: r#"
+    {
+        CrateDiff {
+            baseline {
```
obi1kenobi, line 11: Reminder that this edge is looking at the old version.

src/lints/trait_associated_const_added.ron
```diff
@@ -0,0 +1,64 @@
...
+
+                        associated_constant {
+                            associated_constant: name @output @tag
+                            public_api_eligible @filter(op: "=", value: ["$true"])
+                            default @filter(op: "=", value: ["$null"])
+                        }
+                    }
+                }
+            }
+            current {
```
obi1kenobi, line 30: And this edge is looking at the new version.

You may want to flip the order of `baseline` and `current` in the query for this lint. It's easier to say "new has X and old doesn't" than to say "old doesn't have a thing that we'll look up later." (Actually the second one might be impossible in Trustfall, or at least very non-obvious.)

obi1kenobi:
Lmk if you'd like more in-depth suggestions than I gave in my last post!

No pressure to move this forward, it's totally fine if you had other things to do etc. I just want to make sure you aren't blocked because I did a bad job of explaining stuff :)

Feel free to ping me with more questions anytime!

mrnossiom:
Just didn't have time these last days. Simply switch the query parts made the tests pass with logical results.

obi1kenobi:
Nice! The new query looks good.

A few remaining things to look into:
- Figure out the remaining TODOs in the query file.
- Consider using `@filter(op: "is_null")` and `@filter(op: "is_not_null")` instead of `@filter(op: "=", value: ["$null"])`.
- Add some more test cases, making sure that the lint always triggers when expected _and_ never triggers when not expected. For example:
  - if the trait was previously sealed, adding a const doesn't trigger the lint
  - if the trait was _not_ previously sealed, sealing it in the new version + adding a const still triggers the lint
  - adding a const with a default value doesn't trigger the lint, regardless of whether the trait was sealed
  - if the trait already had at least one const, adding a second one still triggers the lint when appropriate

mrnossiom:
I don't know where to look for the `reference_link`.

src/lints/trait_associated_const_added.ron
```diff
@@ -0,0 +1,62 @@
+SemverQuery(
+    id: "trait_associated_const_added",
+    human_readable_name: "non-sealed trait added associated constant without default value",
+    description: "A non-sealed trait has gained an associated constant without default value which break downstream implementations of the trait",
+    required_update: Major,
+    lint_level: Deny,
+    reference_link: None, // TODO
```
obi1kenobi, line 7:
```suggestion
    reference_link: Some("https://doc.rust-lang.org/cargo/reference/semver.html#trait-new-item-no-default"),
```

obi1kenobi (commented):
The cargo SemVer reference, and the [API evolution RFC 1105](https://rust-lang.github.io/rfcs/1105-api-evolution.html), are the two authoritative documents on SemVer in Rust.

I suggested a link to the "trait added non-defaulted item" section, which encompasses this lint.

obi1kenobi:
If a trait is being annoying in tests and you want to make it not object-safe when it doesn't have a `const` already, just make a method that returns `-> Self`. That also makes it not object-safe, and adding a `const` to it won't cause the extra SemVer breakage.

mrnossiom:
I would love to be able to bless tests

mrnossiom:
I'll remove the comments in `test_crates/trait_associated_const_added/new/src/lib.rs`, but can you check that's actually the trait that should trigger the lint?

test_crates/trait_associated_const_added/new/src/lib.rs
```diff
@@ -5,30 +5,44 @@ mod sealed {
 // trigger
 pub trait WillGainConstWithoutDefault {
     const BAR: bool;
+
+    fn make_me_object_safe() -> Self;
```
obi1kenobi, line 9: These are actually making the trait _stop_ being object-safe, just like having a `const` does.

Perhaps a better name:
```suggestion
    fn make_me_non_object_safe() -> Self;
```

test_crates/trait_associated_const_added/new/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+
+// trigger
+pub trait WillGainAnotherConstWithoutDefault {
+    const ONE: bool;
+    const TWO: bool;
+
+    fn make_me_object_safe() -> Self;
+}
+// trigger
+pub trait WillGainAnotherConstWithoutDefaultSealed: sealed::Sealed {
```
obi1kenobi, line 43: I think this shouldn't trigger, since in `old` the trait is sealed. Perhaps you didn't intend to seal it in the baseline?

mrnossiom: no you are right, it should not indeed trigger

src/lints/trait_associated_const_added.ron
```diff
@@ -0,0 +1,62 @@
+SemverQuery(
+    id: "trait_associated_const_added",
+    human_readable_name: "non-sealed trait added associated constant without default value",
+    description: "A non-sealed trait has gained an associated constant without default value which break downstream implementations of the trait",
```
obi1kenobi, line 4: typo fix:
```suggestion
    description: "A non-sealed trait has gained an associated constant without default value, which breaks downstream implementations of the trait",
```

src/lints/trait_associated_const_added.ron
```diff
@@ -0,0 +1,62 @@
...
+                        visibility_limit @filter(op: "=", value: ["$public"])
+
+                        importable_path {
+                            path @output @tag
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        associated_constant {
+                            associated_constant: name @output @tag
+                            public_api_eligible @filter(op: "=", value: ["$true"])
```
obi1kenobi, line 23: This raises an interesting point. I think we don't want this clause, actually.

For `public_api_eligible` to be `false` and get filtered out here, the new const needs to be `#[doc(hidden)]`. But if the trait is not sealed and someone downstream implemented it, adding a new `const` will break their impl _regardless_ of `#[doc(hidden)]` or not. Hiding the `const` will just result in the downstream user being more confused about why their impl is broken. It won't unbreak the impl.

Would you mind adding a test that catches this case and updating the lint?

src/lints/trait_associated_const_added.ron
```diff
@@ -0,0 +1,62 @@
...
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        associated_constant @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
+                            name @filter(op: "=", value: ["%associated_constant"])
+                            public_api_eligible @filter(op: "=", value: ["$true"])
+                            default @filter(op: "is_null")
```
obi1kenobi, line 48: I think we don't want either of these two clauses.

We're looking for newly-added constants by a given name.

But currently, this block says: "there are no associated constants by that name that are public API (so not `#[doc(hidden)]`) and don't have a default value." So if the const exists but had a default value, this block will match. Same if the constant exists but is `#[doc(hidden)]`. In neither of those cases is the constant newly-added.

Could you add test cases for each of these cases and update the lint?

src/lints/trait_associated_const_added.ron
```diff
@@ -0,0 +1,62 @@
...
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+        "zero": 0,
+    },
+    error_message: "A non-sealed trait has gained an associated constant without default value which break downstream implementations of the trait",
+    per_result_error_template: Some("associated constant {{associated_constant}} of trait {{join \"::\" path}} in file {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 61: slightly shorter to minimize the risk it'll wrap in the terminal:
```suggestion
    per_result_error_template: Some("trait constant {{join \"::\" path}}::{{associated_constant}} in file {{span_filename}}:{{span_begin_line}}"),
```

src/lints/trait_associated_const_added.ron
```diff
@@ -0,0 +1,62 @@
...
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+        "zero": 0,
+    },
+    error_message: "A non-sealed trait has gained an associated constant without default value which break downstream implementations of the trait",
```
obi1kenobi, line 60: typo fix:
```suggestion
    error_message: "A non-sealed trait has gained an associated constant without a default value, which breaks downstream implementations of the trait",
```

obi1kenobi (commented):
A couple of minor typo fixes, plus some edge cases with the lint I just noticed. Sorry I missed these earlier!

src/lints/trait_associated_const_added.ron
mrnossiom: Didn't know public_api_eligible would exclude doc hidden item

obi1kenobi (approved):
Awesome! 🚀

PR #876, opened 2024-08-20
Implement trait_newly_sealed new lint

dmatos2012:
Should contribute to  #870 by adding trait becoming newly sealed lint. Welcome any feedback as I am by no means a rust expert and first time contributor here. 

Edit: Not sure if there is a standard for the `reference_link`, as `sealed traits` didnt have a specific section. Also not sure if i am missing any cursed sealed impls, but welcome any newer ones that should be added. 

Thanks!

### DIFF
--- a/src/lints/trait_newly_sealed.ron
+++ b/src/lints/trait_newly_sealed.ron
@@ -1,0 +1,52 @@
+SemverQuery(
+    id: "trait_newly_sealed",
+    human_readable_name: "pub trait became sealed",
+    description: "A public trait became sealed, so users of this trait can not implement it anymore",
+    required_update: Major,
+    lint_level: Deny,
+    reference_link: Some("https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed"),
+    query: r#"
+    {
+        CrateDiff {
+            baseline {
+                item {
+                    ... on Trait {
+                        visibility_limit @filter(op: "=", value: ["$public"]) @output
+                        sealed @filter(op: "!=", value: ["$true"])
+
+                        importable_path {
+                            path @output @tag
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+                    }
+                }
+            }
+            current {
+                item {
+                    ... on Trait {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        sealed @filter(op: "=", value: ["$true"])
+                        name @output
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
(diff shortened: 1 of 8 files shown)


src/lints/trait_newly_sealed.ron
```diff
@@ -0,0 +1,52 @@
+SemverQuery(
+    id: "trait_newly_sealed",
+    human_readable_name: "pub trait became sealed",
+    description: "A public trait became sealed, so users of this trait can not implement it anymore",
+    required_update: Major,
+    lint_level: Deny,
+    reference_link: Some("https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed"),
```
obi1kenobi, line 7: This is a great reference link!

src/lints/trait_newly_sealed.ron
```diff
@@ -0,0 +1,52 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+    },
+    error_message: "A publicly-visible trait became sealed, so it is no longer allowed to be implemented",
```
obi1kenobi, line 49: tiny wording tweak to make the impact more explicit:
```suggestion
    error_message: "A publicly-visible trait became sealed, so downstream crates are no longer able to implement it",
```

obi1kenobi (approved):
Excellent work — merging right away 🚀

Thank you for contributing 🙌

obi1kenobi:
> Also not sure if i am missing any cursed sealed impls, but welcome any newer ones that should be added.

The current test cases are great. We really want to catch cases where the query in the lint might have a typo or a bug. If the engine running the query has misclassified a trait as sealed when it isn't (or vice versa), that's an upstream bug not a bug in the lint itself. We already have [a _large_ test suite](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/rustdoc-v33/test_crates/sealed_traits/src/lib.rs) of cursed sealed and "can't believe it's not sealed" trait patterns specifically to ensure our analysis of whether a trait is sealed remains accurate.

This is one of the advantages of using Trustfall for a linter: you don't have to be an expert in "what are all the ways to seal a trait" in order to lint for sealed traits! We've already plugged in that knowledge into the query graph, and now it's accessible at the low low cost of writing a query 😁

dmatos2012:
> Excellent work — merging right away 🚀
> 
> Thank you for contributing 🙌

Thanks so much for the encouragement on 'X', and the awesome code. Hope I can contribute further either here or `TrustFall` if there are issues which are not at "expert" levels. :D Looking foward !

PR #880, opened 2024-08-22
Add lint for trait default impl removed

### LINKED ISSUE #294: New lint: non-sealed `pub trait` method default implementation removed
When the `pub trait` method's default implementation is removed, the method is no more provided and turns into a declared method. This forces the new version to implement the now-declared method in all of the `impl` blocks which were deriving that trait with its previously-provided method in the old version.

Required new version: **Major**

This requires some schema additions:
- [x] adding the edge from `Trait` to its `Method`s [linked PR]
- [ ] enabling the `has_body` field for `Method` [linked PR]

dmatos2012:
Should add another lint for #870, particularly `non-sealed trait removed a default impl for a trait method`. Fixes #294.

I am not sure if the `object_safe @filter(op: "=", value: ["$true"])` line is needed, since otherwise it would raise another lint. I thought that it could be a `False Positive`, since at least if i try it locally, I only get the `warning` if it returns `Self`, like so:

```rust
struct Test;
pub trait Trait {
    // fn method(&self) -> Self; #cargo complains about trait obj safety
    fn method(&self); # cargo is fine with it
}
impl Trait for Test {}
```

Anyways, otherwise any feedback happy to change. Thanks!

### DIFF
--- a/src/lints/trait_default_impl_removed.ron
+++ b/src/lints/trait_default_impl_removed.ron
@@ -1,0 +1,61 @@
+SemverQuery(
+    id: "trait_default_impl_removed",
+    human_readable_name: "pub trait default method impl removed",
+    description: "A non-sealed public trait default method impl was removed.",
+    required_update: Major,
+    lint_level: Deny,
+    reference_link: Some("https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations"),
+    query: r#"
+    {
+        CrateDiff {
+            baseline {
+                item {
+                    ... on Trait {
+                        visibility_limit @filter(op: "=", value: ["$public"]) @output
+                        sealed @filter(op: "!=", value: ["$true"])
+                        name @output
+
+                        importable_path {
+                            path @output @tag
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+                        method {
+                            method_name: name @output @tag
+                            public_api_eligible @filter(op: "=", value: ["$true"])
+                            has_body @filter(op: "=", value: ["$true"])
+
+                        }
+                    }
+                }
+            }
+            current {
+                item {
+                    ... on Trait {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        sealed @filter(op: "!=", value: ["$true"])
+                        object_safe @filter(op: "=", value: ["$true"])
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+                        method {
+                            public_api_eligible @filter(op: "=", value: ["$true"])
+                            name @filter(op: "=", value: ["%method_name"])
+                            has_body @filter(op: "!=", value: ["$true"])
+                            span_: span @optional {
+                                filename @output
+                                begin_line @output
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "true": true,
+        "public": "public",
+    },
+    error_message: "A trait's method default impl was removed, breaking users relying on the default impl",
+    per_result_error_template: Some("trait method <{{join \"::\" path}}>::{{method_name}} in file {{span_filename}}:{{span_begin_line}}"),
+)
--- a/src/lints/trait_default_impl_removed.ron
+++ b/src/lints/trait_default_impl_removed.ron
@@ -58,5 +58,5 @@ SemverQuery(
         "public": "public",
     },
     error_message: "A method's default impl in an unsealed trait has been removed, breaking trait implementations that relied on that default",
-    per_result_error_template: Some("trait method <{{join \"::\" path}}>::{{method_name}} in file {{span_filename}}:{{span_begin_line}}"),
+    per_result_error_template: Some("trait method {{join \"::\" path}}::{{method_name}} in file {{span_filename}}:{{span_begin_line}}"),
 )
--- a/src/lints/trait_default_impl_removed.ron
+++ b/src/lints/trait_default_impl_removed.ron
@@ -32,7 +32,6 @@ SemverQuery(
                 item {
                     ... on Trait {
                         visibility_limit @filter(op: "=", value: ["$public"])
-                        sealed @filter(op: "!=", value: ["$true"])
 
                         importable_path {
                             path @filter(op: "=", value: ["%path"])
@@ -43,6 +42,7 @@ SemverQuery(
                             public_api_eligible @filter(op: "=", value: ["$true"])
                             name @filter(op: "=", value: ["%method_name"])
                             has_body @filter(op: "!=", value: ["$true"])
+
                             span_: span @optional {
                                 filename @output
                                 begin_line @output
@@ -57,6 +57,6 @@ SemverQuery(
         "true": true,
         "public": "public",
     },
-    error_message: "A trait's method default impl was removed, breaking users relying on the default impl",
+    error_message: "A method's default impl in an unsealed trait has been removed, breaking trait implementations that relied on that default",
     per_result_error_template: Some("trait method <{{join \"::\" path}}>::{{method_name}} in file {{span_filename}}:{{span_begin_line}}"),
 )
--- a/src/lints/trait_default_impl_removed.ron
+++ b/src/lints/trait_default_impl_removed.ron
@@ -1,7 +1,7 @@
 SemverQuery(
     id: "trait_default_impl_removed",
     human_readable_name: "pub trait default method impl removed",
-    description: "A non-sealed public trait default method impl was removed.",
+    description: "A non-sealed public trait default method impl was removed",
     required_update: Major,
     lint_level: Deny,
     reference_link: Some("https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations"),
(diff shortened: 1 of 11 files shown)


obi1kenobi:
This is very close, and I'm excited to merge it!

Re: the object safety issue you mentioned, I suspect you might accidentally have had the `-> Self` method in the `new` arm of the test, and a method that returns nothing in the `old` arm of the test. That is consistent with the object-safety lint you observed, and would have been eliminated by the extra clause you added in the lint.

The `object_safe` clause is not necessary, and in fact would prevent this lint from triggering in valid cases where a non-object-safe trait had a removed default impl.

A few action item suggestions:
- add a test case where a non-object-safe lint has the default impl removed, and make sure the lint triggers correctly
- re-read your PR in the GitHub code review UI, to help you spot and fix little typos that are easy to miss when we look at code in our own editor
- try to improve the readability and consistency of the lint query with other lint queries, with respect to spacing, trailing spaces, etc.

dmatos2012:
Yeah, thanks for those suggestions, my bad. I normally do these things, and tend to be careful but this time  I was actually in a rush to get it in, before I got busy with something else that I really overlooked doing those details

- I have added the test case for `obj_safe` to make sure it triggers properly
- Fix typos and spacing etc in lint.

Thanks again for feedback!

src/lints/trait_default_impl_removed.ron
```diff
@@ -0,0 +1,62 @@
...
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        method {
+                            public_api_eligible @filter(op: "=", value: ["$true"])
+                            name @filter(op: "=", value: ["%method_name"])
+                            has_body @filter(op: "!=", value: ["$true"])
+                            span_: span @optional {
```
obi1kenobi, line 46: minor personal preference thing: I like a blank line before non-trivial edges
```suggestion

                            span_: span @optional {
```

src/lints/trait_default_impl_removed.ron
```diff
@@ -0,0 +1,62 @@
...
+    lint_level: Deny,
+    reference_link: Some("https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations"),
+    query: r#"
+    {
+        CrateDiff {
+            baseline {
+                item {
+                    ... on Trait {
+                        visibility_limit @filter(op: "=", value: ["$public"]) @output
+                        sealed @filter(op: "!=", value: ["$true"])
```
obi1kenobi, line 15: Right now, none of the test cases include sealed traits so this is never tested. We could remove this clause and the corresponding clause in the `current` arm of the query, and all the tests would still pass. That's a bit scary!

Would you mind adding some test cases that ensure this lint only triggers when the trait wasn't sealed to start with?

test_crates/trait_default_impl_removed/new/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+// Default trait method impl is removed, should be reported
+pub trait TraitA {
+    fn method_default_impl_removed(&self);
+}
+
+// Default trait method becomes removed completely, should not be reported as it becomes
+// missing instead
+pub trait TraitB {}
+
+//Default trait method impl is removed, but its private, shouldnt be reported
```
obi1kenobi, line 10:
```suggestion
// Default trait method impl is removed, but its private, shouldn't be reported
```

test_crates/trait_default_impl_removed/old/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+    fn method_default_impl_removed(&self) {}
+}
+
+// Default trait method becomes removed completely, should not be reported as it becomes
+// missing instead
+pub trait TraitB {
+    fn method_becomes_removed(&self) {}
+}
+
+//Default trait method impl is removed, but its private, shouldnt be reported
```
obi1kenobi, line 12:
```suggestion
// Default trait method impl is removed, but it's private, shouldn't be reported
```

test_crates/trait_default_impl_removed/old/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+trait TraitC {
+    fn method_default_impl_removed(&self) {}
+}
+
+// Non-object safe trait has its default impl removed, should be reported
```
obi1kenobi, line 17: This comment is a bit misleading. It makes it sound like the trait is _currently_ not object-safe, whereas in reality the trait _becomes_ non-object-safe as part of the change in `old -> new`.

Which scenario was the test meant to catch? Let's make sure the description and the actual test are in sync.

src/lints/trait_default_impl_removed.ron
```diff
@@ -0,0 +1,62 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "true": true,
+        "public": "public",
+    },
+    error_message: "A trait's method default impl was removed, breaking users relying on the default impl",
```
obi1kenobi, line 60: wording tweak suggestion to add more context:
```suggestion
    error_message: "A method's default impl in an unsealed trait has been removed, breaking trait implementations that relied on that default",
```

src/lints/trait_default_impl_removed.ron
```diff
@@ -0,0 +1,62 @@
...
+                            has_body @filter(op: "=", value: ["$true"])
+                        }
+                    }
+                }
+            }
+            current {
+                item {
+                    ... on Trait {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        sealed @filter(op: "!=", value: ["$true"])
```
obi1kenobi, line 35: This clause poses an interesting question: if the trait newly became sealed _and_ a default impl was removed, is that two breaking changes or one? Should it trigger two lints, or just one?

My gut feeling is that it should trigger two lints, which would mean we don't actually want this clause: regardless of whether the _new_ trait is sealed, we want to complain about the missing method default impl.
```suggestion
```

What do you think?

P.S.: This edge case is super important to cover with tests! It's the sort of thing that can really sneak up on you!

obi1kenobi (commented):
We're super close here! 🙌

I thought of an interesting edge case while reviewing the lint query, and I think we want to make a small tweak in approach to handle it properly.

There are also a few more typos, test case improvements, and general little points of friction that you could practice catching on your own by self-reviewing your code in the GitHub UI. The fewer comments and fixes I need to apply in code review, the faster and more efficiently we'll both ship code. Fixing things in code review is very inefficient results in fewer/slower merged PRs, so it can be frustrating and wasteful for you, me, and the community as a whole.

That said, don't worry about things that have already happened, and just seek to improve going forward! Becoming better every day is the goal, everything else is icing on the cake. As long as you're interested in improving your skills, I'll be happy to point you in the right direction to do so quickly and efficiently.

Go team! I think we're one last iteration away from shipping this 🚀

dmatos2012:
I am 100% willing to keep improving my skills, so I  appreciate the feedback. I will be more careful on the next PRs , so that the content of discussion should be the technical sides, not the typos as i can imagine that's extremely frustrating. Funny because normally I contribute code, and here is mostly wording thus making it more difficult :stuck_out_tongue: 

But on test cases from your suggestions, I added:
- [ ] Trait becomes sealed and has default impl removed
- [ ] Trait is partially sealed and has default impl removed
- [ ] Trait is fully sealed and has default impl removed

I hope that this one should be the one. Thanks for the patience!!

src/lints/trait_default_impl_removed.ron
dmatos2012: Added the fully sealed case to test that case

src/lints/trait_default_impl_removed.ron
```diff
@@ -0,0 +1,62 @@
...
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "true": true,
+        "public": "public",
+    },
+    error_message: "A method's default impl in an unsealed trait has been removed, breaking trait implementations that relied on that default",
+    per_result_error_template: Some("trait method <{{join \"::\" path}}>::{{method_name}} in file {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 61: I'm not sure we use angle brackets to indicate traits in other places. Rust only uses angle brackets when attempting to use an underlying trait for a type, like `<Example as Iterator>::Item`, but we're referencing the trait itself here.

Probably best to remove them.
```suggestion
    per_result_error_template: Some("trait method {{join \"::\" path}}::{{method_name}} in file {{span_filename}}:{{span_begin_line}}"),
```

test_crates/trait_default_impl_removed/new/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// Default trait method impl is removed, and it's partially sealed, should be reported
+pub trait TraitF {
+    fn method_partially_sealed_has_default_impl_removed(&self, _: private::Token);
+}
+
+// Default trait method impl is removed, but it's sealed. The fact that it's sealed dominates,
+// thus having its default impl removed is not a problem since it was never possible to
+// implement it, should not be reported
```
obi1kenobi, line 37: This is an excellent explanation, well done!

obi1kenobi (approved):
Excellent work! Merging 🚀

obi1kenobi:
Btw, I recently created a Discord server to make it easier to coordinate `cargo-semver-checks` development. If you'd like to join, we'd love to have you there! Here's an invite link (active for 7 days max): https://discord.gg/s7Cfmp3P

PR #882, opened 2024-08-22
Add snapshot test for when a workspace member's baseline has a compile error.

suaviloquence:
When the user runs `cargo semver-checks --workspace` on a workspace that has a member with a baseline version with a compile error, but other non-error workspace members.

The current behavior is to semver-check the members it can, and then return an error at the end saying it couldn't compile the errored crate.

Potential other tests: 

- when the current version has a compile error
- when no crates in the workspace can be built because their current/baseline versions all have compile errors

### DIFF
--- a/src/snapshot_tests.rs
+++ b/src/snapshot_tests.rs
@@ -140,7 +140,7 @@ fn assert_integration_test(test_name: &str, invocation: &[&str]) {
     std::env::remove_var("RUST_BACKTRACE");
     // remove the cargo verbosity variable, which gets passed to `cargo doc`
     // and may create a nonreproducible environment.
-    std::env::remove_var("CARGO_TERM_VERBOSE");
+    std::env::set_var("CARGO_TERM_VERBOSE", "false");
 
     let stdout = StaticWriter::new();
     let stderr = StaticWriter::new();
--- a/src/snapshot_tests.rs
+++ b/src/snapshot_tests.rs
@@ -303,3 +303,25 @@ fn workspace_publish_false_workspace_flag() {
         ],
     )
 }
+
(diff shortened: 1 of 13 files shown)


obi1kenobi (approved):
Excellent idea!

suaviloquence:
Hmm.  We definitely need to redact the ` Blocking waiting for file lock on package cache` in the snapshots but I need to figure out why CI has a slightly different output when running `cargo doc`.

suaviloquence:
Okay, the CI sets the env var `CARGO_TERM_VERBOSE=true` which gets passed to `cargo doc`.  We probably want to unset this.  We could explicitly force `verbose` which could give more detailed snapshots, but we would have to redact more and I don't think it would be worth it.

suaviloquence:
Rebased on #883.

obi1kenobi:
I think this might be causing the problem: https://github.com/obi1kenobi/cargo-semver-checks/blob/main/.github/workflows/ci.yml#L18

suaviloquence:
I think so too, although I explicitly unset it (and then, in an experiment, set it to `false` outright).  I need to do some more digging to figure out where the verbosity is coming from.

suaviloquence:
I found the issue: the output for cargo 1.77 is different than in the newer version, and failing in 1.77 was cancelling the other checks so I didn't figure this out :expressionless:.  I added in a shim to skip if the `cargo` version is less than 1.78.  If you think this is too hacky, let me know.

obi1kenobi (approved):
Thanks for tracking the issue down! Skipping 1.77 seems reasonable, since it's the oldest version we still actively support.

PR #888, opened 2024-08-26
feat: add trait_associated_const_default_removed lint

mrnossiom:
Lint for "non-sealed trait removed the default value for an associated constant"

I still haven't fully made myself of a mental model of how to create queries. Here's a starting point. The query I made matches another test with incoherent results.

### DIFF
--- a/src/lints/trait_associated_const_default_removed.ron
+++ b/src/lints/trait_associated_const_default_removed.ron
@@ -1,0 +1,60 @@
+SemverQuery(
+    id: "trait_associated_const_default_removed",
+    human_readable_name: "non-sealed trait removed the default value for an associated constant",
+    description: "A non-sealed trait associated constant lost its default value, which breaks downstream implementations of the trait",
+    required_update: Major,
+    lint_level: Deny,
+    reference_link: None, // TODO
+    query: r#"
+    {
+        CrateDiff {
+            current {
+                item {
+                    ... on Trait {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+
+                        importable_path {
+                            path @output @tag
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        associated_constant {
+                            associated_constant: name @output @tag
+                            default @filter(op: "is_null") @output
+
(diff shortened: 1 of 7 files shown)


src/lints/trait_associated_const_default_removed.ron
```diff
@@ -0,0 +1,60 @@
...
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        associated_constant @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
+                            name @filter(op: "=", value: ["%associated_constant"])
+                            default @filter(op: "is_not_null")
+                        }
```
obi1kenobi, line 47: This clause says: "there are no associated constants such that both the name matches and there is a default value." So not having any associated constant at all is a match, as is having only constants with different names, or only having constants without default values.

We want something different: "there exists an associated constant with a matching name and with a default value." This way, not having the right constant name is not a match, and not having a default value is not a match.

The clause we want is actually the inverted version of what the query currently says. The query says "there does not exist" but we actually do want it to exist. Merely removing the "there does not exist" part of the query (the "count is zero" part) is sufficient to fix it:
```suggestion
                        associated_constant {
                            name @filter(op: "=", value: ["%associated_constant"])
                            default @filter(op: "is_not_null")
                        }
```

obi1kenobi (commented):
The query is very close! I've added an explanation for what is going wrong and why, hopefully that makes the problem clear.

Thanks for putting this together!

obi1kenobi (approved):
Excellent! 🚀

obi1kenobi:
Did my explanation of the query change make sense btw? Happy to answer any further questions you may have!

obi1kenobi:
Of course, makes sense. Happy to support you as you practice.

Btw I just made a Discord server for cargo-semver-checks and Trustfall, to make coordination a bit easier. You're welcome to join if you'd like — there are many other people in similar places in their learning journey there: https://discord.gg/s7Cfmp3P

PR #889, opened 2024-08-27
feat: add trait_associated_type_default_removed lint

mrnossiom:
Lint for "non-sealed trait removed the default value for an associated type"

It seems like "associated type defaults" are a nightly feature (see tracking issue https://github.com/rust-lang/rust/issues/29661). Does `cargo-semver-checks` support nightly features. Am I just wrong on what feature this lint should test?

### DIFF
--- a/test_crates/trait_associated_type_default_removed/new/src/lib.rs
+++ b/test_crates/trait_associated_type_default_removed/new/src/lib.rs
@@ -1,0 +1,25 @@
+mod sealed {
+    pub(crate) trait Sealed {}
+}
+
+pub trait WillLoseDefault {
+    type Foo;
+}
+
+pub trait WillLoseDefaultSealed: sealed::Sealed {
+    type Foo;
+}
+
+pub trait Unchanged {
+    type Foo = bool;
+}
+pub trait UnchangedSealed: sealed::Sealed {
+    type Foo = bool;
+}
+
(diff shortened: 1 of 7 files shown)


test_crates/trait_associated_type_default_removed/new/src/lib.rs
```diff
@@ -0,0 +1,25 @@
+mod sealed {
```
obi1kenobi, line 1: Associated type defaults are unstable, but can be used with a nightly compiler and by explicitly enabling the feature in the crate's `lib.rs`:
```suggestion
#![feature(associated_type_defaults)]

mod sealed {
```

test_crates/trait_associated_type_default_removed/old/src/lib.rs
```diff
@@ -0,0 +1,25 @@
+mod sealed {
```
obi1kenobi, line 1:
```suggestion
#![feature(associated_type_defaults)]

mod sealed {
```

obi1kenobi (commented):
You're looking at the right thing, the feature just needed to be explicitly enabled in the test crates' source and our test infra will take it from there.

cc @PedroTurik this might be of interest to you too, since you've claimed the other lint that relies on associated type defaults.

obi1kenobi (approved):
Nicely done 🚀

PR #890, opened 2024-08-27
non-sealed trait added associated type without default value lint

PedroTurik:
https://github.com/obi1kenobi/cargo-semver-checks/issues/870


test_crates/trait_associated_type_added/new/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    fn make_me_non_object_safe() -> Self;
+}
+
+pub trait TypeDocHidden {
+    #[doc(hidden)]
+    type Bar;
+
+    fn make_me_non_object_safe() -> Self;
+}
```
obi1kenobi, line 69: Let's make sure all files end in a trailing newline, for maximum compatibility with other tools. All editors have a setting to do this automatically, so I recommend enabling it. For example, here are the [instructions for VSCode](https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files)
```suggestion
}

```

test_crates/trait_associated_type_added/old/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    fn make_me_non_object_safe() -> Self;
+}
+
+pub trait TypeDocHidden {
+    #[doc(hidden)]
+    type Bar;
+
+    fn make_me_non_object_safe() -> Self;
+}
```
obi1kenobi, line 53:
```suggestion
}

```

src/lints/trait_associated_type_added.ron
```diff
@@ -0,0 +1,59 @@
+SemverQuery(
+    id: "trait_associated_type_added",
+    human_readable_name: "non-sealed public trait added associated type without default value",
+    description: "A non-sealed trait has gained an associated type without a default value, which breaks downstream implementations of the trait",
+    required_update: Major,
+    lint_level: Deny,
+    reference_link: Some("https://doc.rust-lang.org/reference/items/associated-items.html#associated-types"),
```
obi1kenobi, line 7: This is a better reference link, since it explains the "trait item without default" rule directly. The link that defines associated types doesn't mention that, so it's less useful.
```suggestion
    reference_link: Some("https://doc.rust-lang.org/cargo/reference/semver.html#trait-new-item-no-default"),
```

obi1kenobi (approved):
Merging with a couple of tweaks, thank you! 🚀

Issue #638, opened 2024-01-19
Tracking issue: Checking cross-crate items


obi1kenobi:
`cargo-semver-checks` currently only generates and analyzes a single crate's rustdoc JSON at a time. If that crate's public API exposes or relies on items from other crates, those foreign items are not present in the rustdoc JSON we get to see.

This can cause both false-positives (e.g. #609) and false-negatives (e.g. #629).

The purpose of this issue is to have a single place to link to when this conversation comes up, and to have a single location where interested folks can track progress.

The underlying technical reasons for the lack of cross-crate checking are the following:
- Rustdoc doesn't inline re-exported items across crates. This used to be the case, but was removed because it was causing lots of bugs and other kinds of unpleasantness.
- There's currently no reliable way to determine which dependency (including crate and exact version, to distinguish between multiple major versions) a cross-crate item came from. There has been interest in adding such a mechanism, though progress seems to have slowed recently: https://github.com/rust-lang/compiler-team/issues/635
- After such a mechanism is added, there will still be probably 3-6 person-months of work needed on the cargo-semver-checks side: we'll need to analyze each rustdoc JSON to find 3rd party crates whose rustdoc JSON we also need, and recursively generate more rustdoc JSON until we've resolved all items in the public API we're scanning. Such a large investment of time would require finding a more permanent source of funding for the project — [sponsoring me on GitHub](https://github.com/sponsors/obi1kenobi) is the way to do that. When I can cover rent with `cargo-semver-checks`, I'll be much more able to work on time-consuming improvements like this.

PR #892, opened 2024-08-28
feat: add `trait_added_supertrait`

### LINKED ISSUE #441: New lint: non-sealed trait added supertrait
Semver-major, because 3rd party types that used to impl the trait in the last version might not impl the new supertrait imposed in the new version.

Example breaking change in the real-world: https://github.com/awslabs/smithy-rs/pull/2577/files#diff-82252c5a4ad7e9d34be9254db1afe172d0f01c067b121de20ff7f7f29d7cf223L24

mrnossiom:
Lint for "non-sealed trait added new supertraits"

Closes #441 

I blessed the tests, but it makes duplicate reports for `trait_newly_sealed` (as sealing a trait requires adding a supertrait) which may be annoying.

### DIFF
--- a/src/lints/trait_added_supertrait.ron
+++ b/src/lints/trait_added_supertrait.ron
@@ -1,0 +1,58 @@
+SemverQuery(
+    id: "trait_added_supertrait",
+    human_readable_name: "non-sealed trait added new supertraits",
+    description: "A non-sealed trait added one or more supertraits, which breaks downstream implementations of the trait",
+    required_update: Major,
+    lint_level: Deny,
+    reference_link: None, // TODO
+    query: r#"
+    {
+        CrateDiff {
+            current {
+                item {
+                    ... on Trait {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+
+                        importable_path {
+                            path @output @tag
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        supertrait {
+                            supertrait: name @output @tag
+                        }
+
+                        span_: span @optional {
+                           filename @output
+                           begin_line @output
+                        }
+                    }
+                }
+            }
+            baseline {
+                item {
+                    ... on Trait {
+                        visibility_limit @filter(op: "=", value: ["$public"]) @output
+                        sealed @filter(op: "!=", value: ["$true"])
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        supertrait @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
+                            name @filter(op: "=", value: ["%supertrait"])
+                        }
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+        "zero": 0,
+    },
+    error_message: "A non-sealed trait added one or more supertraits, which breaks downstream implementations of the trait",
+    per_result_error_template: Some("trait {{join \"::\" path}} has gained {{supertrait}} in file {{span_filename}}:{{span_begin_line}}"),
+)
--- a/src/lints/trait_added_supertrait.ron
+++ b/src/lints/trait_added_supertrait.ron
@@ -4,7 +4,7 @@ SemverQuery(
     description: "A non-sealed trait added one or more supertraits, which breaks downstream implementations of the trait",
     required_update: Major,
     lint_level: Deny,
-    reference_link: Some("doc.rust-lang.org/cargo/reference/semver.html#generic-bounds-tighten"),
+    reference_link: Some("https://doc.rust-lang.org/cargo/reference/semver.html#generic-bounds-tighten"),
     query: r#"
     {
         CrateDiff {
--- a/src/lints/trait_added_supertrait.ron
+++ b/src/lints/trait_added_supertrait.ron
@@ -4,7 +4,7 @@ SemverQuery(
     description: "A non-sealed trait added one or more supertraits, which breaks downstream implementations of the trait",
     required_update: Major,
     lint_level: Deny,
-    reference_link: None, // TODO
+    reference_link: Some("doc.rust-lang.org/cargo/reference/semver.html#generic-bounds-tighten"),
     query: r#"
     {
         CrateDiff {
@@ -54,5 +54,5 @@ SemverQuery(
         "zero": 0,
     },
     error_message: "A non-sealed trait added one or more supertraits, which breaks downstream implementations of the trait",
-    per_result_error_template: Some("trait {{join \"::\" path}} has gained {{supertrait}} in file {{span_filename}}:{{span_begin_line}}"),
+    per_result_error_template: Some("trait {{join \"::\" path}} gained {{supertrait}} in file {{span_filename}}:{{span_begin_line}}"),
 )
--- a/src/query.rs
+++ b/src/query.rs
@@ -1077,6 +1077,7 @@ add_lints!(
     struct_pub_field_now_doc_hidden,
     struct_repr_transparent_removed,
     struct_with_pub_fields_changed_type,
+    trait_added_supertrait,
     trait_associated_const_added,
     trait_associated_const_default_removed,
     trait_associated_const_now_doc_hidden,
--- a/test_crates/trait_added_supertrait/new/Cargo.toml
+++ b/test_crates/trait_added_supertrait/new/Cargo.toml
@@ -1,0 +1,7 @@
+[package]
+publish = false
+name = "trait_added_supertrait"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
--- a/test_crates/trait_added_supertrait/new/src/lib.rs
+++ b/test_crates/trait_added_supertrait/new/src/lib.rs
@@ -19,3 +19,5 @@ pub trait WillGainStdTwo: core::fmt::Debug {}
 pub trait WillGainStdThree: PartialEq {
     fn make_me_non_object_safe() -> Self;
 }
+
+pub trait WillChangeStdToCore: core::fmt::Debug {}
--- a/test_crates/trait_added_supertrait/new/src/lib.rs
+++ b/test_crates/trait_added_supertrait/new/src/lib.rs
@@ -1,15 +1,21 @@
+pub trait TraitOne {}
+pub trait TraitTwo {}
+
 mod sealed {
     pub trait Sealed {}
 }
 
-pub trait TraitOne {}
-pub trait TraitTwo {}
-
 pub trait Unchanged {}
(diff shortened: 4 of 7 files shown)


src/lints/trait_added_supertrait.ron
```diff
@@ -0,0 +1,58 @@
+SemverQuery(
+    id: "trait_added_supertrait",
+    human_readable_name: "non-sealed trait added new supertraits",
+    description: "A non-sealed trait added one or more supertraits, which breaks downstream implementations of the trait",
+    required_update: Major,
+    lint_level: Deny,
+    reference_link: None, // TODO
```
obi1kenobi, line 7: What do you think about citing this portion of the reference here? It's not an exact match, but it's along parallel lines (we're tightening the bounds on which the trait can be implemented) and something is better than nothing...

https://doc.rust-lang.org/cargo/reference/semver.html#generic-bounds-tighten

src/lints/trait_added_supertrait.ron
```diff
@@ -0,0 +1,58 @@
...
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+        "zero": 0,
+    },
+    error_message: "A non-sealed trait added one or more supertraits, which breaks downstream implementations of the trait",
+    per_result_error_template: Some("trait {{join \"::\" path}} has gained {{supertrait}} in file {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 57: saving a few characters to minimize the risk of spilling to a new line in the user's terminal
```suggestion
    per_result_error_template: Some("trait {{join \"::\" path}} gained {{supertrait}} in file {{span_filename}}:{{span_begin_line}}"),
```

obi1kenobi (commented):
This looks good, and don't worry about the matches in other crates. That's very common — we even actively encourage it in cases where it helps us ensure that precisely *one* lint fires per discovered problem.

For example, if a pub struct gets deleted, we don't also want to complain that each of its public fields and methods got deleted too. The test for "what triggers the struct fields deleted" lint then contains instances where the entire struct is deleted, and we use the tests to ensure we *don't* match it.

Assuming you feel the reference link I suggested is good, or you can find a better one, this is good to merge!

src/lints/trait_added_supertrait.ron
mrnossiom: I agree that it's better than nothing, but it may be really confusing. Maybe add a comment on top?

obi1kenobi: I'm open to it! Feel free to add a comment.

mrnossiom: Couldn't find a way to add value with a comment

mrnossiom:
Added tests with `std` supertraits, but the current query doesn't catch them. Do you have any idea why?

suaviloquence:
Another test case to add, maybe:

given:

```rust
pub mod one {
    /// Trait #1
    pub trait MyTrait {}
}

pub mod two {
    /// trait #2
    pub trait MyTrait {}
}
```

old:

```rust
use one::MyTrait;

pub trait Subtrait: MyTrait {}
```

new:

```rust
use two::MyTrait;

pub trait Subtrait: MyTrait {}
```

where `one::MyTrait` and `two::MyTrait` are different traits, even though they have the same name.

src/lints/trait_added_supertrait.ron
```diff
@@ -0,0 +1,58 @@
+SemverQuery(
+    id: "trait_added_supertrait",
+    human_readable_name: "non-sealed trait added new supertraits",
+    description: "A non-sealed trait added one or more supertraits, which breaks downstream implementations of the trait",
+    required_update: Major,
+    lint_level: Deny,
+    reference_link: Some("doc.rust-lang.org/cargo/reference/semver.html#generic-bounds-tighten"),
```
suaviloquence, line 7:
```suggestion
    reference_link: Some("https://doc.rust-lang.org/cargo/reference/semver.html#generic-bounds-tighten"),
```

This is more likely to make it clickable in terminals

suaviloquence:
> Added tests with `std` supertraits, but the current query doesn't catch them. Do you have any idea why?

To me, it looks like a bug in the trustfall rustdoc adapter: it doesn't seem to show supertraits from other crates.

Querying the `itertools` crate for [traits with supertraits](https://play.predr.ag/rustdoc#?f=2&q=*3-Traits-with-supertraits-%2B-where-each-of-them-is-defined*lquery---0Crate---2item---4*E-Trait---6name-*o*l*l--_6*3-Get-the-info-where-this-trait-is-defined.*l--_6span---8filename-*o*l--_8first_line*B-begin_line-*o*l--_6*J*l*l--_6supertrait---8trait---acanonical_path---csupapath*B-path-*o-*l--_a*J--_0*l--_8--*6--*4--*2--*0*J*l*J&v=*C*l*l*J&crate=itertools-0.10.4), it only returns `HomogenousTuple: itertools::tuple_impl::TupleCollect`.  However, this is certainly not all of them, take the [`Itertools`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html) trait which is `: core::iter::Iterator`, but it doesn't come up in the query.

src/lints/trait_added_supertrait.ron
obi1kenobi: Yes, we definitely want the `https://` in front. In fact, it would be great to have a test for this, since we don't have this policy written down anywhere.

obi1kenobi:
Yes, _in general_ we can't see traits from other crates. Recall that rustdoc JSON only contains data for the crate being analyzed, so other crates' items (including traits) are not present in the JSON. This unfortunately includes Rust's built-in crates `std` and `core`.

To work around this limitation until we have the ability to connect multiple rustdoc JSONs, we have [a hardcoded list of traits](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/92a98688eb68115fcc5dc10a519ef7b845c5a9c9/src/indexed_crate.rs#L420-L494) from Rust's builtins that we pretend are part of the JSON and should be able to lint.

The code to pretend those traits exist when they are implemented by a type is [here](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/92a98688eb68115fcc5dc10a519ef7b845c5a9c9/src/adapter/edges.rs#L403-L416). Unfortunately, it appears that the implementation of traits' `supertrait` edge [has neglected to add an equivalent block](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/92a98688eb68115fcc5dc10a519ef7b845c5a9c9/src/adapter/edges.rs#L476-L479). This is a bug.

Thanks for catching it! It should be an easy fix, if you'd like to try your hand at writing some Trustfall adapter code @mrnossiom! We'd want to update [this test](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/92a98688eb68115fcc5dc10a519ef7b845c5a9c9/src/adapter/tests.rs#L122-L175) and its corresponding test crate to contain some `std` bounds, and add a small block of code to look them up like in the other place.

obi1kenobi:
@mrnossiom no pressure, just checking in to see if you're interested in putting together the fix for `trustfall-rustdoc-adapter` to unblock this PR, or if I or someone else should take over either just that component or this PR as a whole. It's no trouble either way, and it's totally fine if you're busy and would like to step away completely. I just wanted to get your preferences on how you'd like this PR to proceed, since you did a lot of work on it already.

mrnossiom:
Hi Predrag, I'm afraid I don't currently have the time to make the fix myself. But I'll be happy to finish the work of this PR's scope once the fix is merged.

obi1kenobi:
No worries at all, I can take care of it and unblock this PR.

obi1kenobi:
The fix is in and I've updated the adapter library versions here, so if you pull the issue should go away.

While working on the fix, I thought of another interesting edge case we want to ensure isn't going to cause a false-positive:
```
pub trait Example: std::fmt::Debug {}
// becomes
pub trait Example: core::fmt::Debug {}
```
This shouldn't get flagged since the name change is essentially only cosmetic — we don't currently track `std` requirements, especially not within any given lint. But in principle, some possible implementations of the supertrait name property _might_ have considered this a name change, and we want to protect against a bad implementation there causing us a false-positive here.

obi1kenobi:
The fix is in and I've updated the adapter library versions here, so if you pull the issue should go away.

While working on the fix, I thought of another interesting edge case we want to ensure isn't going to cause a false-positive:
```rust
pub trait Example: std::fmt::Debug {}
// becomes
pub trait Example: core::fmt::Debug {}
```
This shouldn't get flagged since the name change is essentially only cosmetic — we don't currently track `std` requirements, especially not within any given lint. But in principle, some possible implementations of the supertrait name property _might_ have considered this a name change, and we want to protect against a bad implementation there causing us a false-positive here.

mrnossiom:
The current lint doesn't flag the edge case you mentioned, but I feel like it should. If you have any idea why tell me else I'll look further into it this weekend.

obi1kenobi:
> The current lint doesn't flag the edge case you mentioned, but I feel like it should.

This message was referencing this edge case, right?
```rust
pub trait Example: std::fmt::Debug {}
// becomes
pub trait Example: core::fmt::Debug {}
```

If so, the correct behavior is to _not_ flag this — we want the test case as a guard against false-positives here. The reason is that `std::fmt::Debug` is just a pub re-export of `core:fmt::Debug`, so those two bounds are equivalent from a supertraits perspective. The only difference is that one of them is usable in a `no_std` context and the other is not, but that's not the subject of this lint.

So I believe everything is in order with this lint and I think it's safe to merge. Do you agree? Happy to merge it if so, just wanted to confirm with you in case I've missed something :+1:

mrnossiom:
It's just that there is no special handling for `std` to `core` changes. This means that if `std::fmt::Debug` to `core::fmt::Debug` doesn't trigger, it also means that switching `lib_foo::Trait` to `lib_bar::Trait` would not trigger either.

I don't know if that's acceptable, else I have no other concern for this PR.

obi1kenobi:
You're right, there's definitely an issue there, though not with `std::fmt::Debug` and `core::fmt::Debug` specifically (the `std` one is a re-export of the `core` one) but with `lib_foo::Trait` to `lib_bar::Trait`. This is recorded in issue #922.

However, because of #638, bounds like `lib_foo::Trait` currently won't show up at all, which massively diminishes the surface area of the problem. Technically #638 prevents us from resolving even common built-in Rust traits like `Debug` and `Clone`, but we've special-cased a number of them in particular to work around this temporarily.

There's another (quite remote) false-positive issue here, around blanket impls: adding a supertrait bound can be "non-binding" (essentially, irrelevant and non-breaking) if there's already a blanket impl for that supertrait over all types that could impl the trait. This is #923.

We will resolve all three of these eventually, and the name-matching (#922) is a much bigger concern than blanket impls (#923). But given that #638 prevents the vast majority of the cases where the name-matching would be a problem, I think we'll help vastly more people by shipping this lint as-is and improving it later.

Issue #884, opened 2024-08-24
New lints: an enum variant changes from unit/tuple/struct to another kind. 


suaviloquence:
### Describe your use case

We have lints for when a `struct` changes "kind": 

- [`tuple_struct_to_plain_struct`](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/src/lints/tuple_struct_to_plain_struct.ron)
- [`unit_struct_changed_kind`](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/src/lints/unit_struct_changed_kind.ron)

However, currently `cargo-semver-checks` does not check when an enum variant changes kind like this.

from: 

```rust
pub enum Enum { 
    WasTuple(),
    WasStruct {},
    WasUnit,
    TupleItem (()),
    StructItem { a: (), },
    Unit,
}
```

to:

```rust
pub enum Enum {
    WasTuple {},
    WasStruct,
    WasUnit(),
    TupleItem { a: () },
    StructItem,
    Unit(()),
}
```

the first three variants would break with:

```rust
let _ = Enum::WasTuple;
// since it's now a tuple variant, this still gives us a value,
// but it's a `fn(()) -> Enum` tuple constructor, not an `Enum` instance.
let _: Enum = Enum::WasStruct;
let _ = Enum::WasUnit;
```

the second three also have breakage in a similar fashion, and `cargo-semver-checks` currently does not catch the
fields_missing/removed lints because these variants changed kind

### Describe the solution you'd like

Thus, it would be helpful to have lint(s) that detect this change for **constructible** enum variants - if a tuple/struct variant is marked `#[non_exhaustive]` in the baseline version, this is not a breaking change because it was not possible to construct these variants in the baseline version.

PR #894, opened 2024-08-29
Enum unit variant changed kind

PedroTurik:
solves one lint from https://github.com/obi1kenobi/cargo-semver-checks/issues/884

Applied the same logic from enum_tuple_variant_field_added and similar lints for whether or not newly doc(hidden) and newly non_exhaustive variants should report for both. 

Also, not sure about the reference link. Maybe https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute would be better?

### DIFF
--- a/src/lints/enum_unit_variant_changed_kind.ron
+++ b/src/lints/enum_unit_variant_changed_kind.ron
@@ -1,0 +1,69 @@
+SemverQuery(
+    id: "enum_unit_variant_changed_kind",
+    human_readable_name: "An enum unit variant changed kind",
+    description: "A public enum unit variant that isn't #[non_exhaustive] changed kind",
+    required_update: Major,
+    lint_level: Deny,
+    reference_link: None,
+    query: r#"
+    {
+        CrateDiff {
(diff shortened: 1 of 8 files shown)


src/lints/enum_unit_variant_changed_kind.ron
```diff
@@ -0,0 +1,71 @@
...
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        name @filter(op: "=", value: ["%enum_name"])
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        # Don't check for exhaustiveness and public API eligibility here.
+                        # Variants that changed kind and gained #[doc(hidden)] should
```
obi1kenobi, line 47:
```suggestion
                        # Variants that changed kind and gained `#[doc(hidden)]` should
```

obi1kenobi (approved):
Awesome!

PR #896, opened 2024-08-31
Functionality for unstable feature flags and options

suaviloquence:
*This got big, so if you'd like me to break it up into (feature_flags, unstable_options), I'd be happy to do that.*

This PR adds the functionality to set unstable feature flags (`-Z flag`) as well as feature-flag-gated unstable CLI options (e.g., `-Z unstable-flags --witness-hints`), as well as help messages and functionality when the feature flags become stable.

Design note: the `unstable_options` feature is a little hacky - I don't know if `clap` derive pattern provides all the expressiveness to make it (i.e., arguments only valid when a `-Z flag is passed) work (but I'm not a clap expert).  It's currently implemented to warn if a user passes any flags that alter the unstable options from its `Default` implementation, and reset them to `Default` when `-Z unstable-options` is not passed.  

It may be possible to get the desired functionality of feature flag gating without `unstable-options`, but it will make the transition from feature flag -> CLI option less smooth.  What do you think?


src/config.rs
```diff
@@ -284,6 +287,83 @@ impl GlobalConfig {
...
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct FeatureFlag {
+    /// `kebab-case` identifier for this feature flag.
+    pub id: &'static str,
+    /// Optional help message for this flag, separated by line breaks.  The first line
+    /// should be able to serve as a one-line short help message.
+    pub help: Option<&'static [&'static str]>,
```
obi1kenobi, line 326: To make sure I understand this right, could you elaborate what the elements in the slice should be? The element at `[0]` is the one-line short help message, how about the rest? Do we ever expect more than 1-2 elements? How would they be printed?

src/main.rs
```diff
@@ -16,19 +16,77 @@ mod snapshot_tests;
...
+
+    if config.feature_flag_enabled(FeatureFlag::HELP) {
+        config
+            .log_info(|config| {
+                let header = Style::new()
+                    .bold()
+                    .fg_color(Some(Color::Ansi(AnsiColor::Cyan)));
+                let option = Style::new().bold();
+
+                writeln!(config.stderr(), "{header}Unstable options{header:#}\n")?;
```
obi1kenobi, line 38: Whoa fascinating, what's `{header:#}` here? Never seen that before, is that a fancy way to undo the style by using the alternate representation?

src/main.rs
```diff
@@ -16,19 +16,77 @@ mod snapshot_tests;
...
+
+        write!(
+            config.stderr(),
+            "{}",
+            HelpPrinter::command().render_long_help()
+        )
+        .expect("print failed");
+
+        std::process::exit(0);
+    } else if args.bugreport {
```
obi1kenobi, line 85: Is there a way we can refactor this so that the `args.bugreport` branch happens as early as possible?

Two reasons:
- `--bugreport` should dominate all other flags. We don't want to skip reporting a bug by printing unstable features' help text.
- Every line of code we run before `args.bugreport` is another opportunity for a bug to prevent the bug report from happening. We want as few of those as possible.

obi1kenobi (commented):
This looks really nice, excellent job!

Before we merge, do you think it'd be feasible to add a couple of snapshot tests to make sure that `-Z unstable-options --unstable-help` and similar commands are correctly parsed and produce the expected terminal output? The code here is non-trivial, and as we add unstable options I expect folks who aren't experts in clap will have to update it. It'd be nice to give them both a well-tested starting point and a place to draw inspiration from while designing their own tests.

src/main.rs
suaviloquence: Yeah, I was looking through `cargo` code and they use this way of using (semantically) named styles, which let you use the alternate formatting flag to render a reset code.

obi1kenobi: Very cool, TIL!

src/main.rs
obi1kenobi: Now that I think about it, it's probably a good idea to have a test for `--bugreport` actually ... printing a report 😁 The exact system details in the report aren't relevant and can be masked out, we just care that the command has an effect and something vaguely of the right shape is printed.

suaviloquence:
> Before we merge, do you think it'd be feasible to add a couple of snapshot tests to make sure that `-Z unstable-options --unstable-help`

We can do snapshots for parsing the flags, but because the actual printing of the help happens in the `main` function, we would need to do that in a `tests/*` integration test, not with our current snapshot test infrastructure.  This does come with the compile time penalty we talked about earlier, but I can add it if you think it'd be helpful.

src/config.rs
suaviloquence: That's what I intended, I think `clap` does this behavior internally.  I changed it to print the whole long help message (formatted by lines) with `-Z help`, but we do provide the `help` string to `clap` (although it's currently hidden), so I added a doc-comment that the first line should be able to be used as a short help message.

obi1kenobi: I'm just not sure how the remaining elements in the slice get printed. I wouldn't know whether I'm expected to offer exactly two elements here, or offer one element per line, or one element per paragraph.

Could you elaborate the doc comment on those points, for the benefit of future code reviewers' ability to review `help` text for any future options being added here.

obi1kenobi:
It's probably worth the compilation time penalty to test this. Thanks for bringing it up, the distinction hadn't occurred to me.

src/config.rs
obi1kenobi: Ah I see you've already changed it :+1:

src/config.rs
```diff
@@ -284,6 +287,119 @@ impl GlobalConfig {
...
+/// to stable feature, mark the associated constant `#[deprecated]`, but don't remove it.  This will warn downstream
+/// code and `cargo-semver-checks` binary users that the flag has been stabilized and may be removed in a
+/// future release, and hide it from `-Z help`, without breaking downstream code and binary users right
+/// when it is stabilized.
+///
+/// When you stabilize a flag and mark it as `#[deprecated]`, remove any `if` blocks that feature-gate code.
+/// Testing if a flag is enabled with [`GlobalConfig::feature_flag_enabled`] will always return
+/// `true` when a flag has been stabilized.
+///
+/// ## See also
```
obi1kenobi, line 347: Try to keep line lengths under 100 chars whenever it's reasonably practical, just for convenience of allowing multiple editors side-by-side on a small screen. clippy won't lint for this AFAICT, and `cargo fmt` won't reformat docstrings, so this is a bit of a manual step. I have rulers set up in my editor to help me not exceed 100 chars, and I highly recommend that approach.

src/main.rs
```diff
@@ -16,19 +16,22 @@ mod snapshot_tests;
...
     human_panic::setup_panic!();
 
-    let Cargo::SemverChecks(args) = Cargo::parse();
+    let Cargo::SemverChecks(mut args) = Cargo::parse();
+
+    let feature_flags = HashSet::from_iter(args.unstable_features);
 
     configure_color(args.color_choice);
     let mut config = GlobalConfig::new();
+    config.set_log_level(args.check_release.verbosity.log_level());
```
obi1kenobi, line 25: This line appears to be duplicated below the big if/elif block. Is that intentional?

src/main.rs
```diff
@@ -491,6 +590,41 @@ impl From<CheckRelease> for cargo_semver_checks::Check {
...
+    if !config.feature_flag_enabled(FeatureFlag::UNSTABLE_OPTIONS)
+        && unstable_options != &UnstableOptions::default()
+    {
+        config
+            .shell_warn(
+                "unstable options were passed without `-Z unstable-options`. They will be ignored.",
+            )
+            .expect("print failed");
+
+        *unstable_options = UnstableOptions::default();
```
obi1kenobi, line 624: Do `cargo` and `rustdoc` simply ignore unstable options, or do they exit non-zero? It feels like the latter might be safer, especially for a tool that's intended to be used in a CI environment where ~nobody reads the logs...

src/main.rs
```diff
@@ -82,6 +85,65 @@ fn main() {
             }
             Ok(())
         });
+        std::process::exit(0);
+    } else if config.feature_flag_enabled(FeatureFlag::HELP) {
```
obi1kenobi, line 89: Could you add comments inside these blocks mentioning what command the user would have run to end up inside them?

The difference between `config.feature_flag_enabled(FeatureFlag::HELP)` vs `args.unstable_options.unstable_help` is not immediately obvious to me right now.

src/main.rs
suaviloquence: Oh, you're right, at least `cargo` exits with an error instead.  Let's use that behavior.

src/main.rs
suaviloquence: it's because the verbosity is not a global option, it's in the `CheckRelease` struct, so it could either be in the `SemverChecks.check_release` from `cargo semver-checks -v ...` or in the subcommand `cargo semver-checks check-release -v ...` - currently these are distinct.  The first line is from before we know whether we are running `check_release` (e.g., for bug report or lint info or now unstable help), but we still want to have access to the log level.  We could also move the verbosity field to the global options to get rid of the duplicated lines, but that would be a slightly bigger change.

src/main.rs
suaviloquence: I annotated the code paths, but it also makes more sense at this point to just have one `-Z help` option for both feature flags and unstable options.

suaviloquence:
> It's probably worth the compilation time penalty to test this

Great, I'll add some tomorrow.

src/main.rs
obi1kenobi: Ah that's interesting and I think unintentional. If you want to open a separate PR to move it to the global scope, I'd happily merge that.

src/main.rs
```diff
@@ -84,6 +94,66 @@ fn main() {
...
+                            writeln!(config.stderr(), "{first}")?;
+
+                            for line in lines {
+                                writeln!(config.stderr(), "{:<20}{line}", "")?;
+                            }
+                        }
+                    } else {
+                        writeln!(config.stderr())?;
+                    }
+                }
```
obi1kenobi, line 129: What do you think about printing these to stdout instead of stderr?

It might be confusing to run a command and have your desired output (the help text) appear on stderr instead of stdout. But if cargo prints to stderr then maybe I'm thinking about this wrong.

src/main.rs
```diff
@@ -491,6 +624,53 @@ impl From<CheckRelease> for cargo_semver_checks::Check {
...
+            ))
+            .expect("printing failed");
+    }
+
+    if !config.feature_flag_enabled(FeatureFlag::UNSTABLE_OPTIONS) {
+        let non_default_options = unstable_options.non_default();
+
+        if !non_default_options.is_empty() {
+            let mut message = String::from(
+                "The following options are not supported without `-Z unstable-options`:\n",
```
obi1kenobi, line 657: I think our current `anyhow::bail!()` messages try to be lowercase, to match cargo and rustc. I'm assuming this will get printed as `error: The following ...` right?

If so, consider:
```suggestion
                "the following options are not supported without `-Z unstable-options`:\n",
```

suaviloquence:
Should be ready now.  I added snapshot tests for `-Z help` and for when an unstable option is passed without `-Z unstable-options`.

suaviloquence:
When working on using unstable options in the witness branch, it makes more sense to use unstable options in the `CheckRelease` set of arguments versus global `SemverChecks` (similarly, but opposite, to the `verbosity` change). This would prevent us from having unstable global options, but I can't currently think of a use case for that.  I can also change it so have `UnstableGlobalOptions` and `UnstableCheckReleaseOptions` if you think that would be helpful, but we could also just implement the global options functionality when (or if) it is needed.  What do you think?

obi1kenobi:
The `check-release` subcommand is essentially deprecated: we want people to use c-s-c as `cargo semver-checks` only. There are no plans to add other subcommands parallel to `CheckRelease`, and we want anything that works after the `check-release` CLI argument to also work at global level.

So we want users to be able to use the unstable flags directly off of `cargo semver-checks`, as opposed to having to always write `cargo semver-checks check-release` to use them. As long as that's the case, I'm fine with whichever placement you feel is better.

I know it's super early in the morning for you right now, so I'll hold off on merging this until you've had a chance to read this and make a call on what you think is best here.

src/main.rs
```diff
@@ -487,6 +617,55 @@ impl From<CheckRelease> for cargo_semver_checks::Check {
...
+        if !non_default_options.is_empty() {
+            let mut message = String::from(
+                "the following options are not supported without `-Z unstable-options`:\n",
+            );
+
+            for option in non_default_options {
+                use std::fmt::Write as _;
+
+                // writes to strings are infallible.
+                let _ = writeln!(&mut message, " - `{option}`");
```
obi1kenobi, line 659: Can a write cause a reallocation which then fails?

I'd feel better about this being an `.expect("writes to strings are infallible")` than a comment. I've been on many a wild goose chase as a result of someone using a "comment assert" instead of a real assert, so I've learned to find them suspicious by default.
```suggestion
                writeln!(&mut message, " - `{option}`").expect("writes to strings are infallible");
```

suaviloquence: I think (confusingly) it is an [API guarantee](https://doc.rust-lang.org/std/fmt/struct.Error.html) that writes to strings are fine, but I don't think this error check would impact performance at all.

suaviloquence:
> So we want users to be able to use the unstable flags directly off of `cargo semver-checks`, as opposed to having to always write `cargo semver-checks check-release` to use them. As long as that's the case, I'm fine with whichever placement you feel is better.

Yes, you would be able to use them with both e.g.,  `cargo semver-checks -Z unstable-options --witness-hints` and `cargo semver-checks -Z unstable-options check-release --witness-hints`.  Moving them to the `CheckRelease` struct just makes it available in the `impl From<CheckRelease> for cargo_semver_checks::Check` impl, so we can store check-related configuration there (which makes it reproducible e.g. in snapshot tests)

PR #911, opened 2024-09-03
Release v0.35.0.

obi1kenobi:


obi1kenobi:
@suaviloquence apparently our snapshots hardcode the cargo-semver-checks version string as part of the "impl" field in the diagnostic we print. I'll update them to get this release unblocked. If you could add an escape to them when you get a chance, that'd be lovely.

Issue #898, opened 2024-09-01
New lints: Changes to enum variant discriminants


obi1kenobi:
Background on discriminants: https://doc.rust-lang.org/reference/items/enumerations.html#discriminants

We're looking for three lints total:
- [x] a public API enum without an explicit `repr()` has a variant that used to have a well-defined discriminant, but no longer does:
  - We only check the "without explicit `repr()`" case here. This is because if the enum has a `repr()`, then its variants always have a well-defined discriminant so there's nothing to lint. The case when the `repr()` is removed (and therefore might un-define some discriminants) is handled by another lint.
- a public API enum's variant had its discriminant has changed from its previous value: @dmatos2012 
  - [x] the enum has an explicit representation, like `#[repr(u8)]` or `#[repr(i64)]`: #1036
  - [x] the enum does not have an explicit representation: #912 

We distinguish between the repr/no-repr cases so that we can offer better error messages. When an enum has an explicit repr, that imposes an ABI (application binary interface) constraint in addition to the API constraint. Without a repr, the *numeric* value of the discriminant is fixed but its *binary representation* is not and could be anything.

For examples of querying the `repr` properly, please look at the `enum_repr_int_changed` lint. Make sure to include test cases where the repr is `#[repr(C, i64)]` or `#[repr(u8, C)]` since both variants are allowed too. The lints must appropriately detect the integer repr in both these cases and more straightforward ones like `#[repr(i8)]`.

dmatos2012:
Ill take the the 2nd bullet point, `a public API enum's variant had its discriminant has changed from its previous value ` lint, which will cover the two bullet points you mention :)

dmatos2012:
> Cool, got you marked as owning that. Are you still working on #885 as well, or is that blocked on something?

[answered](https://github.com/obi1kenobi/cargo-semver-checks/issues/885#issuecomment-2325289242)

obi1kenobi:
Ah @dmatos2012 we actually want four lints for this issue, not two. The second bullet point describes two lints, since we want to lint for "you're breaking ABI" (such as for FFI use cases) separately from "you are breaking API." That will let us make stronger statements than "may break FFI", since hedging is generally bad when reporting errors.

PR #912, opened 2024-09-03
Add enum variant discriminant changed lint

dmatos2012:
Adds lint `a public API enum's variant had its discriminant has changed from its previous value` from #898. 

PS: Hope no typos or spacing issues this time, only technical things. Thanks :)


obi1kenobi:
Oh wow, we ... might have found a breaking change in clap? 😮

@dmatos2012 could you dig into the source code of clap at the version used in CI, and compare it to the previous clap release to see if the enum discriminant value indeed changed?

If it didn't, that sounds like a false-positive we should investigate. If it did, we'll need to adjust what the CI test expects.

dmatos2012:
@obi1kenobi Indeed, there are bunch of discriminant changes from `clap-v3.1.18` to `clap-v3.2.0`. For example, the first one:
```
Failed in:
  variant AppSettings::AllowInvalidUtf8ForExternalSubcommands 13 -> 14
  
  ```
  
  #clap-v3.1.18
  ```
  pub enum AppSettings {
    IgnoreErrors,
    WaitOnError,
    AllowHyphenValues,
    AllowNegativeNumbers,
    AllArgsOverrideSelf,
    AllowMissingPositional,
    TrailingVarArg,
    DontDelimitTrailingValues,
    InferLongArgs,
    InferSubcommands,
    SubcommandRequired,
    SubcommandRequiredElseHelp,
    AllowExternalSubcommands,
    AllowInvalidUtf8ForExternalSubcommands,
}
  ```
 
 Then clap-v3.2.0 
 ```
 pub enum AppSettings {
    IgnoreErrors,
    WaitOnError,
    AllowHyphenValues,
    AllowNegativeNumbers,
    AllArgsOverrideSelf,
    AllowMissingPositional,
    TrailingVarArg,
    DontDelimitTrailingValues,
    InferLongArgs,
    InferSubcommands,
    SubcommandRequired,
    SubcommandRequiredElseHelp,
    AllowExternalSubcommands,
    Multicall,
    AllowInvalidUtf8ForExternalSubcommands,
}
 ```
 
 So in this case `  AllowInvalidUtf8ForExternalSubcommands` moved from pos 13 to 14 as the magic error indicates:)
 
 Because `MultiCall` went from :
 ```
     #[deprecated(
        since = "3.1.0",
        note = "Replaced with `Command::multicall` and `Command::is_multicall_set`"
    )]
    #[cfg(feature = "unstable-multicall")]
    Multicall,


```

to 
```
    /// Deprecated, replaced with [`Command::multicall`] and [`Command::is_multicall_set`]
    #[deprecated(
        since = "3.1.0",
        note = "Replaced with `Command::multicall` and `Command::is_multicall_set`"
    )]
    Multicall,

```

That being said, I guess should I adjust the `CI` test?
  
  TLDR: `MultiCall` was enabled only under `unstable-multicall` feature, but its on by default on `clap-v3.2.0`

obi1kenobi:
Nice, that looks legit.

Do you think you can edit the clap CI test to expect both lints to trigger, or would you like some help/suggestions?

src/lints/enum_variant_discriminant_changed.ron
```diff
@@ -0,0 +1,65 @@
+SemverQuery(
+    id: "enum_variant_discriminant_changed",
+    human_readable_name: "Public enum's variant had its discriminant changed from its previous value",
```
obi1kenobi, line 3: This field is intended to be a name — it's printed in the headline of the CLI diagnostics we emit when breakage is found — so it should be quite short and it's okay if it doesn't have all the details.

Check out how these fields are used in the CLI ("what gets printed where") if you have a moment. I think that'll give you a better sense of what to put in the various fields going forward.
```suggestion
    human_readable_name: "enum variant had its discriminant change value",
```

dmatos2012:
Applied the change and changed the `CI` test as well

obi1kenobi:
Just making sure you've seen this comment: https://github.com/obi1kenobi/cargo-semver-checks/issues/898#issuecomment-2327511560

We should split this lint into two separate lints based on whether there's an ABI concern or just an API concern. This will let us offer specific feedback to the user that is stronger than "may break ABI." That's important because our users likely have a general expectation that `cargo-semver-checks` should be the expert in most cases, not them. We want to be able to tell them "you broke ABI compatibility" or "you didn't," not "learn about SemVer of variants and discriminants, then decide if you broke the ABI or not."

Sorry I didn't catch this earlier, hope you don't mind making the tweaks!

dmatos2012:
Yeah no problem, I dont mind making the tweaks. I have changed everything slightly, so it might as well be kinda like a new lint.  I hope that I got all the details correctly, and if I am missing something, feedback is welcomed :) 

To reiterate, this is targeting `a public API enum's variant had its discriminant has changed from its previous value`, on the API only brekage(thus no repr). Thank you!

.github/workflows/ci.yml
```diff
@@ -1329,7 +1329,7 @@ jobs:
       - name: Check output
         run: |
           cd semver
-          EXPECTED="$(echo -e "--- failure auto_trait_impl_removed: auto trait no longer implemented ---")"
+          EXPECTED="$(echo -e "--- failure auto_trait_impl_removed: auto trait no longer implemented ---\n--- failure enum_no_repr_variant_discriminant_changed: enum variant had its discriminant change value ---")"
           RESULT="$(cat output | grep failure)"
```
obi1kenobi, line 1333: See if you can squeeze a `sort` call after `grep failure` here? I'm not sure if we guarantee (or want to guarantee) a particular order of lint outputs, and flaky tests are annoying so better avoid the risk whenever possible.
```suggestion
          EXPECTED="$(echo -e "--- failure auto_trait_impl_removed: auto trait no longer implemented ---\n--- failure enum_no_repr_variant_discriminant_changed: enum variant had its discriminant change value ---")"
          RESULT="$(cat output | grep failure | sort)"
```

src/lints/enum_no_repr_variant_discriminant_changed.ron
```diff
@@ -0,0 +1,78 @@
...
+                item {
+                    ... on Enum {
+                        visibility_limit @filter(op: "=", value: ["$public"]) @output
+                        enum_name: name @output @tag
+
+                        attribute @optional {
+                            content {
+                                base @filter(op: "!=", value: ["$repr"])
+                            }
+                        }
```
obi1kenobi, line 22: I don't think we want `@optional` here. Try adding an enum that has two non-repr attributes on it (e.g. `#[non_exhaustive]` and `#[allow(dead_code)]`) to see why.

There's a subtle difference in Trustfall between "if it has an attribute, it isn't repr" (AKA `@optional` and `@filter` like here) and "it doesn't have repr" (AKA `@fold @transform(op: "count") ...`). The difference is around what happens on edges with cardinality greater than 1: in the former case we have a "left outer join" in relational terms, and in the latter (because of the `@fold` aggregation), we don't.

src/lints/enum_no_repr_variant_discriminant_changed.ron
```diff
@@ -0,0 +1,78 @@
...
+
+                        importable_path {
+                            path @output @tag
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        variant {
+                            variant_name: name @output @tag
+                            discriminant @optional {
```
obi1kenobi, line 31: Why `@optional` here?

src/lints/enum_no_repr_variant_discriminant_changed.ron
```diff
@@ -0,0 +1,78 @@
...
+                item {
+                    ... on Enum {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        name @filter(op: "=", value: ["%enum_name"])
+
+                        attribute @optional {
+                            content {
+                                base @filter(op: "!=", value: ["$repr"])
+                            }
+                        }
```
obi1kenobi, line 48: It shouldn't matter if the enum _now_ gained a repr, the API is still broken.

src/lints/enum_no_repr_variant_discriminant_changed.ron
```diff
@@ -0,0 +1,78 @@
...
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "repr": "repr",
+        "true": true,
+    },
+    error_message: "The public enum's variant had its discriminant changed from its previous value. This breaks implementors that relied on the discriminant value.",
```
obi1kenobi, line 76: Let's make sure to be very specific about what broke, and precise with terminology. Enums aren't traits so they don't have implementors.
```suggestion
    error_message: "The enum's variant had its discriminant value change. This breaks downstream code that used its value via a numeric cast like `as isize`.",
```

src/lints/enum_no_repr_variant_discriminant_changed.ron
```diff
@@ -0,0 +1,78 @@
+SemverQuery(
+    id: "enum_no_repr_variant_discriminant_changed",
```
obi1kenobi, line 2: Naming is hard. This is not the best name, and not the worst. Nothing better comes to my mind, but if you have other suggestions, it would be good to consider them.

obi1kenobi:
Thanks for working on this tricky lint!

dmatos2012:
Hopefully this addresses all the changes requested :)

 But there is still a pending task, the renaming of the actual lint. Ill make the renames in the code once we come up with a good one, but for now i left since didnt want to rename everything too many times.
 
 Possible ones:
 * enum_variant_discriminant_changed_api_only_breakage
 * enum_variant_discriminant_changed_no_abi_changes
 * enum_variant_discriminant_changed_api_only
 
 I dont know as you said, naming is hard and particularly this one. Specially since we would like to keep naming consistent with the other lints.

src/lints/enum_no_repr_variant_discriminant_changed.ron
```diff
@@ -0,0 +1,78 @@
...
+
+                            discriminant {
+                                new_value: value @output @filter (op: "!=", value: ["%old_value"])
+                            }
+                        }
+
+                        span_: span @optional {
+                            filename @output
+                            begin_line @output
+                        }
```
obi1kenobi, line 63: Can we grab the span of the variant instead of the span of the enum? It's nicer to point to `Example::Variant` instead of to `enum Example`.

test_crates/enum_no_repr_variant_discriminant_changed/new/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+// Discriminant changed values. Having #[non_exhaustive] on the enum should not have any effect
+// on the *API* breakage, should still be reported.
+#[non_exhaustive]
+pub enum DiscriminantIsChanged {
+    First = 1,
+}
+
+// Discriminant changed values and the variant is also marked as `non_exhaustive`.
+// This now implies that the enum is no longer `well-defined`, which means that a numeric
+// cast is no longer possible on the enum, should not be reported.
```
obi1kenobi, line 18:
```suggestion
// This now implies that the discriminant is no longer `well-defined`, which means that a numeric
// cast is no longer possible on the enum. Should not be reported.
```
Please try to use precise, clear language and good grammar as much as possible. Long sentences with lots of commas are hard to read.

test_crates/enum_no_repr_variant_discriminant_changed/new/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+// This means that the enum is already not `well-defined`, and the numeric cast
+// was never possible, should not be reported.
+#[non_exhaustive]
+pub enum DiscriminantIsChangedButAlreadyNonExhaustive {
+    First,
+    #[non_exhaustive]
+    Second = 2,
+}
+
+// Discriminant changed to be doc hidden and explicit. Being doc hidden is not relevant
```
obi1kenobi, line 37: Some more imprecise language:
```suggestion
// Variant became doc(hidden) while variant became explicit. Being doc(hidden) is not relevant
```

test_crates/enum_no_repr_variant_discriminant_changed/new/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+// Implicit discriminant value changed by becoming explicit, should not be reported as it involves
+// ABI breakage as well, and this is linted at *API* only breakage.
+#[repr(u8)]
+pub enum FieldlessChangesToExplicitDescriminant {
+    Tuple(),
+    Struct {},
+    Unit = 5,
+}
+
+// Variant `Struct` acquires a field. By doing so, it makes the enum not `well-defined`,
```
obi1kenobi, line 72:
```suggestion
// Variant `Struct` acquires a field. By doing so, it makes the discriminant not `well-defined`,
```

test_crates/enum_no_repr_variant_discriminant_changed/new/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+    Second = 2,
+}
+
+// Discriminant changed values, but the variant is already `non_exhaustive`.
+// This means that the enum is already not `well-defined`, and the numeric cast
+// was never possible, should not be reported.
+#[non_exhaustive]
+pub enum DiscriminantIsChangedButAlreadyNonExhaustive {
+    First,
+    #[non_exhaustive]
```
obi1kenobi, line 33: Could you add a test case where another variant is `#[non_exhaustive]`, not the one that changes discriminant values?

I think there's a bug in the lint query related to it: the lint is checking for "is the variant with the changed discriminant marked `#[non_exhaustive]`" when it should be checking for "does the enum have any `#[non_exhaustive]` variants" — not quite the same thing.

obi1kenobi (commented):
This lint name is fine, let's keep it. I spent a bunch of time thinking about alternatives, and couldn't come up with anything that didn't have worse downsides.

This PR is super close, it needs maybe one more round of polishing. I think there might be a bug in the query (which might be hidden by a feature by the underlying rustdoc plugin). There's also some imprecise language and use of terminology. Otherwise, good to go! 🚀

A quick meta suggestion outside this PR: consider going through past PRs, making a list of the changes I've suggested in reviews in the past, and using it as a checklist for things to self-review more closely before opening a new PR. With that list, you could catch pretty much everything I've commented on so far — and that's a great way to improve your skills and get more PRs done in less time.

dmatos2012:
Thanks for the detailed feedback, is greatly appreciated. I did the following:
* Point `span` to the actual variant
* Add two new test cases to cover the bug you discovered with the `non_exhaustive` variants
* Fix the suggestions on clarity. 

Regarding clarity or precision in language: Every new PR either on `trustfall` itself, or the `RustLang` nuances, I learn a ton. This means that sometimes I am not clear with the use of terminology, but this might come from the fact that I am not sure myself either, and end up using it plainly wrong or making it unclear. This is where your expertise helps me, so that I can get those right for future PRs and contributions.

Thanks again for the detailed feedback and *HOPE* that this time is good. This has been a very tricky lint for me, specially if compared with my first two lints :)

obi1kenobi (approved):
Go team! Excellent work, I'm thrilled to see you keep improving with every new PR 🚀

PR #918, opened 2024-09-05
Add support for specifying features with commas `feat1,feat2`

### LINKED ISSUE #915: Supplying comma-separated features in the CLI doesn't work
CLI args like `--baseline-features 24bits,50bits,libm` or `--current-features 24bits,50bits,libm` are interpreted as one long feature named `"24bits,50bits,libm"` instead of a comma-separated list of features.

I believe this is a divergence from how the cargo CLI handles features. Specifying the features one-at-a-time with a long CLI flag like `--baseline-features` is very clunky. We should support lists, and we should add CLI tests using the new testing framework to ensure this doesn't regress.

Bug noticed in https://github.com/obi1kenobi/cargo-semver-checks/issues/914#issuecomment-2329356250

dmatos2012:
Fixes #915  by enabling support for specifying features via commas. Specifically it adds support for the flags:
* --baseline-features
* --current-features
* --features

Before:
`cargo semver-checks --baseline-features foo,bar`
Raised:
```
warning: Feature `std,multicall` is not present in the baseline.
```

Now:
`cargo semver-checks --baseline-features foo,bar`
Succeeds fine. 

Thank you! :)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -484,17 +484,31 @@ struct CheckRelease {
     /// Add a feature to the set of features being checked.
     /// The feature will be used in both the baseline and the current version
     /// of the crate.
-    #[arg(long, value_name = "NAME", help_heading = "Features")]
+    #[arg(
+        long,
+        value_delimiter = ',',
+        value_name = "NAME",
+        help_heading = "Features"
+    )]
     features: Vec<String>,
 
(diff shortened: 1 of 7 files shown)


obi1kenobi (approved):
Thank you!

PR #926, opened 2024-09-11
Add some lint group infra + 'all' lint group

suaviloquence:
Adds the lint pseudo-group `all` to configure all lints in `cargo-semver-checks`.  Mainly adds the beginnings of infrastructure to specify lint groups:
- adds the `priority` field to TOML like the `cargo` lints table
  - makes it an error if there are conflicting entries for the same lint at the same priority level (this is by field)
  - reorganized schema to encode the fact that `{}` or `{priority = ...}` is an error
- changes up data structures to support lint groups:
  - `OverrideMap` is now a middle-level structure `Identifier -> Overrides` (where identifier is group, pseudo-group, or lint name) that compiles to a `String (lint name) -> Overrides`, and makes it an error if there was a conflict
    - in the future `compile` may need to take an `&OverrideStack` to calculate e.g. the `warnings` group but that can be added later

When we add lint groups, the lint group names will need to be statically known so we can deserialize correctly in `Identifier`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,7 +23,6 @@ use trustfall_rustdoc::{load_rustdoc, VersionedCrate};
 use rustdoc_cmd::RustdocCommand;
 use std::collections::{BTreeMap, HashSet};
 use std::path::{Path, PathBuf};
-use std::sync::Arc;
 use std::time::Instant;
 
 pub use config::{FeatureFlag, GlobalConfig};
@@ -492,7 +491,7 @@ note: skipped the following crates since they have no library target: {skipped}"
                 let workspace_overrides =
                     manifest::deserialize_lint_table(&metadata.workspace_metadata)
                         .context("[workspace.metadata.cargo-semver-checks] table is invalid")?
-                        .map(|table| Arc::new(table.inner));
+                        .map(|table| table.into_stack());
 
                 selected
                     .iter()
@@ -535,12 +534,18 @@ note: skipped the following crates since they have no library target: {skipped}"
 
                             if lint_workspace_key || metadata_workspace_key {
                                 if let Some(workspace) = &workspace_overrides {
-                                    overrides.push(Arc::clone(workspace));
+                                    for level in workspace {
+                                        overrides.try_push(level).context("configuration conflict found in workspace overrides.\n\
+                                            use the `priority` field in the manifest to set precedence.")?;
+                                    }
                                 }
                             }
 
                             if let Some(package) = package_overrides {
-                                overrides.push(Arc::new(package.inner));
+                                for level in package.into_stack() {
+                                overrides.try_push(&level).with_context(|| format!("configuration conflict found in package {} overrides.\n\
+                                    use the `priority` field in the manifest to set precedence.", selected.name))?;
+                                }
                             }
 
                             let start = std::time::Instant::now();
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,8 +27,8 @@ use std::time::Instant;
 
 pub use config::{FeatureFlag, GlobalConfig};
 pub use query::{
-    ActualSemverUpdate, LintLevel, OverrideMap, OverrideStack, QueryOverride, RequiredSemverUpdate,
-    SemverQuery,
+    ActualSemverUpdate, CompiledOverrideMap, Identifier, LintLevel, OverrideMap, OverrideStack,
+    QueryOverride, RequiredSemverUpdate, SemverQuery,
 };
 
 /// Test a release for semver violations.
--- a/src/manifest.rs
+++ b/src/manifest.rs
@@ -3,7 +3,7 @@ use std::collections::BTreeMap;
 use anyhow::Context;
 use serde::Deserialize;
 
-use crate::{LintLevel, OverrideMap, QueryOverride, RequiredSemverUpdate};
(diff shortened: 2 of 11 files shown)


src/manifest.rs
```diff
@@ -99,61 +99,102 @@ pub(crate) struct LintTable {
...
 #[derive(Debug, Clone, Deserialize)]
 #[serde(untagged)]
 pub(crate) enum OverrideConfig {
-    /// Specify members by name, e.g.
+    /// Specify both lint level and required update by name, e.g.
     /// `lint_name = { level = "deny", required-update = "major" }
-    /// Any omitted members will default to `None`
     #[serde(rename_all = "kebab-case")]
-    Structure {
+    Both {
```
suaviloquence, line 168: this is split into three states so the `None, None` state is unrepresentable

test_outputs/snapshot_tests/cargo_semver_checks__snapshot_tests__all_lint_group-output.snap
```diff
@@ -0,0 +1,148 @@
...
+  mod all_lint_group::module_missing, previously in file [ROOT]/test_crates/manifest_tests/all_lint_group/old/src/lib.rs:4
+
+--- stderr ---
+     Parsing all_lint_group v0.1.0 (current)
+      Parsed [TIME] (current)
+     Parsing all_lint_group v0.1.0 (baseline)
+      Parsed [TIME] (baseline)
+    Checking all_lint_group v0.1.0 -> v0.1.0 (no change)
+    Starting [TOTAL] checks, 0 unnecessary on [THREADS] threads
+        PASS [TIME]       major        auto_trait_impl_removed
```
obi1kenobi, line 55: This list will break as soon as we merge any new lints.

src/lib.rs
```diff
@@ -535,12 +534,18 @@ note: skipped the following crates since they have no library target: {skipped}"
...
+                                    }
                                 }
                             }
 
                             if let Some(package) = package_overrides {
-                                overrides.push(Arc::new(package.inner));
+                                for level in package.into_stack() {
+                                overrides.try_push(&level).with_context(|| format!("configuration conflict found in package {} overrides.\n\
+                                    use the `priority` field in the manifest to set precedence.", selected.name))?;
+                                }
```
obi1kenobi, line 548: Could you add tests that trigger each of these errors? I'm curious to see how they are displayed to users, since I suspect they'll be somewhat common as soon as we add more groups — nobody will remember to add the priority key *until* they see one of these errors.

obi1kenobi: As a side note, maybe these should say "`priority` field in the manifest's lint configuration"? Or is that too long or otherwise undesirable?

src/manifest.rs
```diff
@@ -99,61 +99,102 @@ pub(crate) struct LintTable {
...
 }
 
-impl From<LintTable> for OverrideMap {
-    fn from(value: LintTable) -> OverrideMap {
-        value.inner
+impl LintTable {
+    /// Converts this into a stack of `OverrideMap`s, where the top of the stack (later
+    /// indices in the `Vec`) has higher priority than the bottom.
+    pub(crate) fn into_stack(self) -> Vec<OverrideMap> {
+        // use a priority -> OverrideMap BTreeMap, which will be sorted by priority
```
obi1kenobi, line 110: Are the two uses of "priority" between the doc comment and the subsequent regular comment inside the function using opposite definitions of "higher priority" here, in terms of whether that signifies precedence vs lack thereof?

If so, that's going to be extremely confusing!

src/manifest.rs
```diff
@@ -99,61 +99,102 @@ pub(crate) struct LintTable {
...
+                    },
+                ),
+            };
+
+            map.entry(priority).or_default().insert(id, overrides);
+        }
+
+        // this will be sorted by key `priority` in ascending order.
+        // to get the stack semantics of later elements take priority,
+        // we need to reverse this.
```
obi1kenobi, line 156: consistency nit:
```suggestion
        // This will be sorted by key `priority` in ascending order.
        // To get the stack semantics of later elements take priority,
        // we need to reverse this.
```

src/manifest.rs
```diff
@@ -99,61 +99,102 @@ pub(crate) struct LintTable {
...
+    /// `lint_name = { level = "deny" }
+    #[serde(rename_all = "kebab-case")]
+    LintLevel {
+        level: LintLevel,
         #[serde(default)]
-        required_update: Option<RequiredSemverUpdate>,
+        priority: i64,
+    },
+    /// Specify just required update by name, with optional priority.
+    /// `lint_name = { level = "deny" }
```
obi1kenobi, line 187: This appears to be accidental copy-pasta since it references level instead of required update.

obi1kenobi (commented):
This PR is a bit too large, and I think it's a bit too risky (in terms of bugs slipping in) to merge as-is. Would you be able to try splitting it up into three PRs:
- priority changes and their tests;
- infra for having lint groups, and
- the special "all" group?

I think that will help both of us review it and spot any bugs that might be present. There's no rush so feel free to take it slow — I'm assuming your semester might have started already?

suaviloquence:
> Would you be able to try splitting it up into three PRs:

Working on this now. I'll open a new PR for priority first when it's ready.

obi1kenobi:
Awesome, thank you!

PR #932, opened 2024-09-17
Handle `priority` in the lints table

suaviloquence:
Like the cargo `[lints]` table, the `priority` key in a lint entry is optional (defaults to 0), and a lower (more negative) `priority` overrides a higher (more positive) entry that configures the same lint (which lets us define how to handle a lint group containing a lint, and a lint group).  Changing `OverrideStack` from holding `Arc<OverrideMap>`s to just `OverrideMap`s is setting up for the upcoming lint group PR where each `OverrideMap` will be compiled based on lower items in the `OverrideStack`, so shared ownership doesn't make sense to use anymore.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,7 +23,6 @@ use trustfall_rustdoc::{load_rustdoc, VersionedCrate};
 use rustdoc_cmd::RustdocCommand;
 use std::collections::{BTreeMap, HashSet};
 use std::path::{Path, PathBuf};
-use std::sync::Arc;
 use std::time::Instant;
 
 pub use config::{FeatureFlag, GlobalConfig};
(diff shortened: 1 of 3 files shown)


obi1kenobi (approved):
This looks great.

In a subsequent PR, could you think about a way to warn users if a lint could hit multiple configs that affect it with the same priority level? That's analogous to [this clippy lint for regular cargo lints](https://rust-lang.github.io/rust-clippy/master/index.html#/lint_groups_priority?groups=cargo,complexity,correctness,deprecated,nursery,pedantic,perf,restriction,style,suspicious).

If you can't think of an easy and elegant way to do it, it's fine to just open an issue about it and move on — we shouldn't let it block other work if it's more than a 15min thing to build.

PR #933, opened 2024-09-18
Document contributor process for adding witness hint templates

suaviloquence:
Adds information about the `witness` field and when and how to add a hint template.  Also sets the `witness` to `None` in `make_new_lint.sh`, with a link to this new section of the contributor docs.

[Rendered view](https://github.com/suaviloquence/cargo-semver-checks/blob/new-witness-doc/CONTRIBUTING.md)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -210,6 +210,124 @@ the "actual" output, then re-run `cargo test` and make sure everything passes.
 
 Congrats on the new lint!
 
+#### Adding a witness
+
+**Witnesses** are a new, unstable feature of `cargo-semver-checks` that let us create witness
+code, a minimal compilable example of potential downstream breakage\. They are configured via the
+`witness` field in the lint file `src/lints/<lint_name>.ron`:
+
+If it is `None` (or the field is omitted entirely), `cargo-semver-checks` will not be able
+to generate witness code for this lint\. This can be the right choice, sometimes:
+for example, if this lint is a warn- or allow-by-default lint that hints at potential
+breakage, but won't cause breaking changes directly\. Additionally, if it's not currently
+possible to write a Trustfall query that gets the necessary information to generate
+witnesses, leave this field as `None`, but leave a `// TODO` comment explaining
+what would unblock this lint from being able to generate a witness.
+
+##### Hint templates
+
+When the `witness` field is not `None`, it must have the `hint_template` field\. This is a
+`handlebars` template that generates a small (1-3 line) human-readable message that
+explains the idea of how downstream code would break\. For example, for the `function_missing` lint:
+
+```rust
+use {{join "::" path};
+{{name}}(...);
+```
+
+which could render to something like:
+
+```ron
+witness: (
+  hint_template: r#"use function_missing::will_be_removed_fn;\n\
+  will_be_removed_fn(...);"#
+)
+```
+
+A witness hint like this may not be buildable (e.g., we elide the function arguments here,
+and the function call is not inside a block), but a witness hint should be a distilled
+example of breakage, and should not require extra information beyond the original
+lint's query.
+
+##### Templating
+
+We use the `handlebars` crate for writing these templates\. [More information
+more information about the syntax can be found here](https://handlebarsjs.com/guide/#simple-expressions),
+and [here is where `cargo-semver-checks` defines custom helpers
+](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/src/templating.rs).
+
+All fields marked with `@output` in the `query` in `<lint_name>.ron` are available
+to access with `{{name}}` in the `hint_template`, like in the example above.
+
+##### Testing
+
+When the `witness` field is not `None`, `cargo-semver-checks` tests the witness generation
+of the lint similarly to how it tests the `query` itself\. If you run `cargo test` after adding
+a witness for the first time, the test will fail, because it expects a file containing the
+expected witness output.
+
+To solve this, create a file called `test_outputs/witnesses/<lint_name>.output.ron`. Make the
+contents `{}` currently, but we will change this\. Run `cargo test` again\. The test
+`<lint_name>` should fail (because we told it we expect no witnesses to be generated,
+and this is no longer true).
+
+The failed test should output a message like:
+
+```
+---- query::tests_lints::function_missing stdout ----
+--- actual output ---
+{
+    // cut down for readibility
+    "./test_crates/function_missing/": [
+        (
+            filename: "src/lib.rs",
+            begin_line: 1,
+            hint: "use function_missing::will_be_removed_fn;\nwill_be_removed_fn(...);",
+        ),
+        (
+            filename: "src/lib.rs",
+            begin_line: 4,
+            hint: "use function_missing::pub_use_removed_fn;\npub_use_removed_fn(...);",
+        ),
+    ],
+}
+thread 'query::tests_lints::function_missing' panicked at src/query.rs:751:17:
+Witness output for function_missing did not match expected values:
+Differences (-expected|+actual):
+-{}
++{
++    // cut down for readibility
++    "./test_crates/function_missing/": [
++        (
++            filename: "src/lib.rs",
++            begin_line: 1,
++            hint: "use function_missing::will_be_removed_fn;\nwill_be_removed_fn(...);",
++        ),
++        (
++            filename: "src/lib.rs",
++            begin_line: 4,
++            hint: "use function_missing::pub_use_removed_fn;\npub_use_removed_fn(...);",
++        ),
++    ],
++}
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -210,6 +210,124 @@ the "actual" output, then re-run `cargo test` and make sure everything passes.
 
 Congrats on the new lint!
 
+#### Adding a witness
+
+**Witnesses** are a new, unstable feature of `cargo-semver-checks` that let us create witness
+code, a minimal compilable example of potential downstream breakage\. They are configured via the
```
obi1kenobi, line 216: Let's explain the new term directly instead of making a circular definition first.
```suggestion
**Witnesses** are a new, unstable feature of `cargo-semver-checks` that let us create
minimal compile-able examples of potential downstream breakage\. They are configured via the
```

CONTRIBUTING.md
```diff
@@ -210,6 +210,124 @@ the "actual" output, then re-run `cargo test` and make sure everything passes.
...
+When the `witness` field is not `None`, it must have the `hint_template` field\. This is a
+`handlebars` template that generates a small (1-3 line) human-readable message that
+explains the idea of how downstream code would break\. For example, for the `function_missing` lint:
+
+```rust
+use {{join "::" path};
+{{name}}(...);
+```
+
+which could render to something like:
```
obi1kenobi, line 238: "Render" is a bit ambiguous here.
```suggestion
which in the RON format could be represented like:
```

CONTRIBUTING.md
```diff
@@ -210,6 +210,124 @@ the "actual" output, then re-run `cargo test` and make sure everything passes.
...
+breakage, but won't cause breaking changes directly\. Additionally, if it's not currently
+possible to write a Trustfall query that gets the necessary information to generate
+witnesses, leave this field as `None`, but leave a `// TODO` comment explaining
+what would unblock this lint from being able to generate a witness.
+
+##### Hint templates
+
+When the `witness` field is not `None`, it must have the `hint_template` field\. This is a
+`handlebars` template that generates a small (1-3 line) human-readable message that
+explains the idea of how downstream code would break\. For example, for the `function_missing` lint:
```
obi1kenobi, line 231:
```suggestion
When the `witness` field is not `None`, it must have the `hint_template` field\. This is a
`handlebars` template that generates a small (1-3 line) human-readable message that
explains the idea of how downstream code would break.

This example code is meant to be small and illustrative, and does not have to pass `cargo check`. It should give the reader a sense of the kind of breakage in one glance.

For example, for the `function_missing` lint, a witness template may look like this:
```

CONTRIBUTING.md
```diff
@@ -210,6 +210,124 @@ the "actual" output, then re-run `cargo test` and make sure everything passes.
...
+
+```rust
+use {{join "::" path};
+{{name}}(...);
```
obi1kenobi, line 235: This is actually broken in the same way the enum variants example was, it will fail under renaming =/

```suggestion
{{join "::" path}(...);
```

CONTRIBUTING.md
```diff
@@ -210,6 +210,124 @@ the "actual" output, then re-run `cargo test` and make sure everything passes.
...
+```
+
+which could render to something like:
+
+```ron
+witness: (
+  hint_template: r#"use function_missing::will_be_removed_fn;\n\
+  will_be_removed_fn(...);"#
+)
+```
```
obi1kenobi, line 245: It might be worth illustrating multi-line templates in a separate example, after going through the basic example end-to-end first.

CONTRIBUTING.md
```diff
@@ -210,6 +210,124 @@ the "actual" output, then re-run `cargo test` and make sure everything passes.
...
+
+A witness hint like this may not be buildable (e.g., we elide the function arguments here,
+and the function call is not inside a block), but a witness hint should be a distilled
+example of breakage, and should not require extra information beyond the original
+lint's query.
```
obi1kenobi, line 250:
```suggestion
This hint will not pass `cargo check` (e.g. the function call arguments are elided), 
and that's okay. The hint is a distilled example of breakage, and shouldn't require
additional information beyond what the lint query retrieved.
```

CONTRIBUTING.md
```diff
@@ -210,6 +210,124 @@ the "actual" output, then re-run `cargo test` and make sure everything passes.
...
+
+##### Templating
+
+We use the `handlebars` crate for writing these templates\. [More information
+more information about the syntax can be found here](https://handlebarsjs.com/guide/#simple-expressions),
```
obi1kenobi, line 255: duplicated words
```suggestion
about the syntax can be found here](https://handlebarsjs.com/guide/#simple-expressions),
```

CONTRIBUTING.md
```diff
@@ -210,6 +210,124 @@ the "actual" output, then re-run `cargo test` and make sure everything passes.
...
+and [here is where `cargo-semver-checks` defines custom helpers
+](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/src/templating.rs).
+
+All fields marked with `@output` in the `query` in `<lint_name>.ron` are available
+to access with `{{name}}` in the `hint_template`, like in the example above.
+
+##### Testing
```
obi1kenobi, line 262:
```suggestion
##### Testing witnesses
```

CONTRIBUTING.md
```diff
@@ -210,6 +210,124 @@ the "actual" output, then re-run `cargo test` and make sure everything passes.
...
++    ],
++}
+
+Update the `test_outputs/witnesses/function_missing.output.ron` file if
+                    the new test results are correct.
+```
+
+Check the actual result under `--- actual ---` and make sure that it outputted
+the correct witness hints\. Note that it may contain output for other test crates - this
+not necessarily an error: see the troubleshooting section for more info\. Once you've
```
obi1kenobi, line 320:
```suggestion
is not necessarily an error: see the troubleshooting section for more info\. Once you've
```

CONTRIBUTING.md
```diff
@@ -210,6 +210,124 @@ the "actual" output, then re-run `cargo test` and make sure everything passes.
...
+verified that the `--- actual ---` results are as expected, copy the actual results
+into the `test_outputs/witnesses/<lint_name>.output.ron` file and save\. Running
```
obi1kenobi, line 322: Is there a viable way to `insta` this? 🤔

obi1kenobi (commented):
Looks good! Just some wording nitpicks here and there.

CONTRIBUTING.md
suaviloquence: Totally! I was matching the behavior of the basic lint tests, but we could definitely use `insta` here, and it would probably be better for contributor experience (but less consistent)

CONTRIBUTING.md
suaviloquence: Will fix in #934

CONTRIBUTING.md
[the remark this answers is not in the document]
suaviloquence: Just a bad regex find-and-replace for replacing `.  ` with `. ` (extra spaces between periods). :sweat_smile:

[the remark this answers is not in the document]
obi1kenobi: Ah okay, no worries. You might need to edit my suggestions post-merge then, since I mirrored the `\.` style.

CONTRIBUTING.md
obi1kenobi: I imagine we'll eventually move the existing non-insta steps to insta, so probably best to start standardizing toward the new behavior we'll want eventually anyway. If someone is annoyed at the inconsistency, we can tell them we'd love a PR to move everything to insta 😅

suaviloquence: OK, I'll open a PR to use insta for the tests instead, then update the docs here.

suaviloquence: #935

CONTRIBUTING.md
```diff
@@ -210,6 +210,85 @@ the "actual" output, then re-run `cargo test` and make sure everything passes.
...
+witnesses, leave this field as `None`, but leave a `// TODO` comment explaining
+what would unblock this lint from being able to generate a witness.
+
+##### Hint templates
+
+When the `witness` field is not `None`, it must have the `hint_template` field. This is a
+`handlebars` template that generates a small (1-3 line) human-readable message that
+explains the idea of how downstream code would break.
+
+This example code is meant to be small and illustrative, and does not have to pass `cargo check`. It should give the reader a sense of the kind of breakage in one glance.
```
obi1kenobi, line 233:
```suggestion
This example code is meant to be small and illustrative, and does not have to pass `cargo check`.
It should give the reader a sense of the kind of breakage in one glance.
```

obi1kenobi (approved):
Awesome 🚀

PR #936, opened 2024-09-19
Reduce size of uploaded crate

DaniPopes:
Currently, the compressed size of the `cargo-semver-checks` crate on crates.io is 1.6MiB, as listed on crates.io and lib.rs. This is mostly attributed to the brand and test directories, which are not essential for the crate to compile. We can avoid uploading these directories by explicitly excluding them in the manifest using `package.exclude`.


obi1kenobi (approved):
Awesome, thank you!

Issue #879, opened 2024-08-22
Consider using `margo` for integration tests involving registries.


suaviloquence:
### Describe your use case

As part of the effort to write snapshot tests for `cargo-semver-checks` edge cases (see #846), we need to use a Cargo registry for some edge cases, for example when the baseline version is yanked as in #275.  It is probably not ideal both in terms of stress on crates.io and testing speed to use the official crates.io registry.

### Describe the solution you'd like

One alternative registry is [margo](https://github.com/integer32llc/margo), which is a tool that creates a (sparse) registry of static files, that can be hosted on any HTTP server.  This would be better than using crates.io because it is a sparse registry, so we don't have to download the index, and it can also be hosted locally because it's just static files.

It looks like `margo` added [support for yanking](https://github.com/integer32llc/margo/pull/11), so it would at least be sufficent for testing that edge case.

### Alternatives, if applicable

There may be similar projects/alternative registries.  We will also need to determine whether to host a test registry publicly (e.g., on GitHub pages, which would make our tests require internet access), locally (ad-hoc server on test run, which would be more complex), or a mix of the two.

obi1kenobi:
Using a GitHub pages test registry in a GitHub Actions CI job would be totally fine. I wouldn't add it in the suite that `cargo test` runs by default. But we have plenty of dedicated CI jobs that require internet access already and test for various regressions, and it's worked out reasonably so far.

PR #945, opened 2024-09-25
Add third party registry support

drewkett:
This adds a `--registry` command line flag to the check release capabilities. The argument to the flag is the name of the registry to use to lookup the crates when using the default behavior or --baseline-version. If left unset, the default behavior remains of --checking crates.io.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -16,6 +16,8 @@ pub struct GlobalConfig {
     stdout: AutoStream<Box<dyn Write + 'static>>,
     stderr: AutoStream<Box<dyn Write + 'static>>,
     feature_flags: HashSet<FeatureFlag>,
+    /// Registry name to look up crates in
+    registry: Option<String>,
 }
 
 impl Default for GlobalConfig {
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -16,6 +16,8 @@ pub struct GlobalConfig {
     stdout: AutoStream<Box<dyn Write + 'static>>,
     stderr: AutoStream<Box<dyn Write + 'static>>,
     feature_flags: HashSet<FeatureFlag>,
+    /// Registry name to look up crates in
+    registry: Option<String>,
```
drewkett, line 20: I wasn't sure if this was the best place to stash the registry flag from the cli. It seems like its okay, but I can't say I did a deep dive of the code to fully understand all the components.

suaviloquence:
> As far as third party registries, one thought I had was to setup a static file based registry with `margo`. That does require at least spinning up a local host based web server, but I think that wouldn't be too much trouble to make work.

See also #879 for some discussion about using `margo` for testing.

fujiapple852:
Rebased and reworked this PR against the latest `main` and also added an integration test: https://github.com/obi1kenobi/cargo-semver-checks/pull/1048

PR #951, opened 2024-09-27
Further insta

CommanderStorm:
I know that during https://github.com/obi1kenobi/cargo-semver-checks/pull/688 you said that waiting is appropriate.
Thing is that this is a really low ammount of work which is still left (given that insta is already in use in other parts)
Am I missing something?

Hope that @vasucp1207 does not mind too much ^^

For reviewing of this PR:
- I have split up the commits to be easy to review. There are a few formatting differences between `insta` and the previous technique.
  => Please review this commit by commit to make this reviewable
- Most of the changes are whitespace
  => please press the ignore whitespace button in the review window

> [!NOTE]
> (unless that is not something you care about for these files 😉): If you decide to merge, please don't squash it to prevent the git history of the files getting lost. (moving and editing in 2 commits is history preventing, I don't know about squashing)

Resolves https://github.com/obi1kenobi/cargo-semver-checks/issues/634


obi1kenobi:
Hey! Thanks for doing all the work to make this happen.

An unfortunate part about reviewing individual commits is that the GitHub UI doesn't allow suggesting changes at the individual commit level. So when I make comments, you'll have to do a bunch of rebasing to apply the change to the appropriate commit. Are you sure you're okay with that?

We've generally preferred squash-merging so that there's less ceremony around making trivial updates to PRs, and so that every commit on `main` is valid, buildable, and passes tests. Also, so that we can revert PRs in one command, instead of having to revert all the commits separately. But if you don't mind the overhead of all those rebases (and we'll need a couple of iterations here), I don't mind doing per-commit reviews here.

Just one thing before we get started: could you separate out the "unrelated snapshot fixes" commit into its own PR, since it doesn't seem related to the rest of the work here?

CONTRIBUTING.md
```diff
@@ -111,6 +111,38 @@ To generate this data, please run `./scripts/regenerate_test_rustdocs.sh`.
 To use a specific toolchain, like beta or nightly, pass it as
 an argument: `./scripts/regenerate_test_rustdocs.sh +nightly`.
 
+## What are those `.snap` or `.snap.new` files generated via `cargo test`
```
obi1kenobi, line 114: Does it make sense to make CI fail if `.snap.new` files are checked into the repo? Will it do that already?

CONTRIBUTING.md
```diff
@@ -111,6 +111,38 @@ To generate this data, please run `./scripts/regenerate_test_rustdocs.sh`.
 To use a specific toolchain, like beta or nightly, pass it as
 an argument: `./scripts/regenerate_test_rustdocs.sh +nightly`.
 
+## What are those `.snap` or `.snap.new` files generated via `cargo test` 
+
+Next to the regular testcase strategies, we also utilise snapshot testing.
```
obi1kenobi, line 116:
```suggestion
As part of our overall testing strategy, we use a technique called "snapshot testing."
```

CONTRIBUTING.md
```diff
@@ -111,6 +111,38 @@ To generate this data, please run `./scripts/regenerate_test_rustdocs.sh`.
...
+   ```text
+   INSTA_UPDATE=always cargo test
+   ```
+3. **Manually**: If you can't (or don't want to) use `cargo-insta`, you can verify the snapshot
+   file manually.  There should be a file called `test_outputs/witnesses/<lint_name>.snap.new`.
+   Open it, and verify that the witnesses generated as expected.  Once you've checked it, move it
+   to `test_outputs/witnesses/<lint_name>.snap` (remove the `.new`)
+
+
+
```
obi1kenobi, line 145: Please try to review your own PRs using the GitHub interface before marking them ready for review, so that you can catch and fix little issues like trailing whitespace or other things that become obvious in the diff viewer. It'll save time and will mean we can merge your PRs faster.
```suggestion
```

CONTRIBUTING.md
```diff
@@ -269,18 +301,13 @@ time, run `cargo test` to start generating the snapshots.  The first time you ru
...
 
 We use `insta` for snapshot testing witness results, so after adding/changing a witness, we need
-to update the test outputs. Note that it may contain output for other test crates - this
-is not necessarily an error: see the troubleshooting section for more info.
+to update the test outputs.
 
-There are two ways to update the output:
+> [!TIP]
+> It may contain output for other test crates - this is not necessarily an error:
+> See the [troubleshooting section](#troubleshooting) for more info.
```
obi1kenobi, line 308: Nice callout!

obi1kenobi:
Overall this is quite close to merging — thank you for putting it together!

I would just like to remove the changes from the integration test snapshots from this PR and make them into their own PR if they are necessary. With that + the minor nitpicks above, I think this can merge 🎉

CONTRIBUTING.md
CommanderStorm: It will somewhat already do that (unless one creates a `.snap.new` and the same `.snap` file, will have to test that part)

CommanderStorm:
(continuing the discussion from #957)

About the different file suffixes, that is is as far as I read into it not configurable.
This makes sense as the files contain the metadata and the snapshot => would not be valid `.ron` anyway => no syntax highlighting..

Here are the two current usages using `.snap`
- https://github.com/obi1kenobi/cargo-semver-checks/blob/main/test_outputs/witnesses/function_missing.snap 
- https://github.com/obi1kenobi/cargo-semver-checks/blob/main/test_outputs/integration_snapshots__bugreport.snap

=> Not using `.snap` as a suffix would also be a bit confusing 😅

CONTRIBUTING.md
obi1kenobi: Yes, having both a `.snap.new` and a `.snap` file was what I would love to make sure we prevent. Because if that happens, it will pollute the repo and confuse the local setups of everyone else, since everyone's insta may think there are altered snapshots.

CONTRIBUTING.md
```diff
@@ -111,6 +111,36 @@ To generate this data, please run `./scripts/regenerate_test_rustdocs.sh`.
...
+The tool we use for this ([`insta`](https://insta.rs/docs/)) is user friendly and has mutliple ways to interact with it:
+
+These snapshots are by default written to `.snap.new` files (because `INSTA_UPDATE` explained below defaults to `auto`) if they differ and fail the testcase.
+Reviewing them is possible via these options:
+
+1. **With `cargo-insta`**: If you install (or have already installed) the `insta` CLI with
+   `cargo install --locked cargo-insta`, you can run `cargo insta review`. Check that the
+   new output is what you expect, and accept it in the TUI.
+2. **Without `cargo-insta`**:
+   From [`insa`s docs](https://insta.rs/docs/quickstart/#tests-without-insta):
```
obi1kenobi, line 126:
```suggestion
   From [`insta`s docs](https://insta.rs/docs/quickstart/#tests-without-insta):
```

CONTRIBUTING.md
```diff
@@ -269,18 +299,13 @@ time, run `cargo test` to start generating the snapshots.  The first time you ru
 it will fail, because there's no expected result to compare to.  Let's make the test pass:
 
 We use `insta` for snapshot testing witness results, so after adding/changing a witness, we need
-to update the test outputs. Note that it may contain output for other test crates - this
-is not necessarily an error: see the troubleshooting section for more info.
+to update the test outputs.
 
-There are two ways to update the output:
+> [!TIP]
+> It may contain output for other test crates - this is not necessarily an error:
```
obi1kenobi, line 305: this is just me nitpicking on typography, don't mind me 😅
```suggestion
> It may contain output for other test crates — this is not necessarily an error:
```

test_outputs/integration_snapshots__unstable_options_without_flag.snap
```diff
@@ -5,6 +5,9 @@ info:
   args:
     - semver-checks
     - "--witness-hints"
+  env:
+    CARGO_TERM_COLOR: never
+    RUST_BACKTRACE: "0"
```
obi1kenobi, line 10: Let's remove all such unrelated snapshot test changes from this PR.

CONTRIBUTING.md
```diff
@@ -111,6 +111,36 @@ To generate this data, please run `./scripts/regenerate_test_rustdocs.sh`.
...
+   > - `no`: does not update snapshot files at all (just runs tests)
+
+   Thus, if you run the following command, you can accept the current snapshots after reviewing the `.snap.new` files.
+   ```text
+   INSTA_UPDATE=always cargo test
+   ```
+3. **Manually**: If you can't (or don't want to) use `cargo-insta`, you can verify the snapshot
+   file manually.  There should be a file called `test_outputs/witnesses/<lint_name>.snap.new`.
+   Open it, and verify that the witnesses generated as expected.  Once you've checked it, move it
+   to `test_outputs/witnesses/<lint_name>.snap` (remove the `.new`)
```
obi1kenobi, line 142: We don't just use snapshots for witnesses, so let's generalize the advice to anything contributors may run into.
```suggestion
   file manually.  There should be a file called `test_outputs/<some_path>/<lint_name>.snap.new`.
   Open it, and verify that its contents match what you expected: all expected data is present, and no unexpected data is included. 
   Once you've checked it, remove the `.new` suffix so that the file's new path
   is `test_outputs/<some_path>/<lint_name>.snap`
```

obi1kenobi (commented):
A few remaining TODOs before we can merge:
- Please remove the changes in the unrelated snapshot files, since we'll want to merge them in a separate PR if they are needed.
- Please read through the `Adding a new lint` section and update the process description as needed to fit the new workflow. For example, [it currently still references `.output.ron` files](https://github.com/obi1kenobi/cargo-semver-checks/pull/951/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R169).
- Please update the `make_new_lint.sh` script as needed. I'm not sure if it's better to generate an empty snapshot file, or to not generate a snapshot at all — what do you think?

obi1kenobi:
> (continuing the discussion from #957)
> 
> About the different file suffixes, that is is as far as I read into it not configurable. This makes sense as the files contain the metadata and the snapshot => would not be valid `.ron` anyway => no syntax highlighting..
> 
> Here are the two current usages using `.snap`
> 
> * https://github.com/obi1kenobi/cargo-semver-checks/blob/main/test_outputs/witnesses/function_missing.snap
> * https://github.com/obi1kenobi/cargo-semver-checks/blob/main/test_outputs/integration_snapshots__bugreport.snap
> 
> => Not using `.snap` as a suffix would also be a bit confusing 😅

All this seems reasonable. In #957 I brought up the `.snap` suffix since IIRC we were just renaming `.ron` to `.snap` without using insta's format which is where the metadata gets added. When we're using insta, I think it's reasonable to do things insta's way :)

CommanderStorm:
> I'm not sure if it's better to generate an empty snapshot file, or to not generate a snapshot at all — what do you think?

I don't think this terribly matters.
Added a file as this way users know where to look, but 🤷🏻

obi1kenobi:
> > I'm not sure if it's better to generate an empty snapshot file, or to not generate a snapshot at all — what do you think?
> 
> I don't think this terribly matters. Added a file as this way users know where to look, but 🤷🏻

Once we merge this, it would be a great idea if you were to run through the lint-writing process end-to-end and experience it as a new user would. That way, any issues that come up can be fixed while all this is still fresh in our minds and before any new contributors run into problems.

When this PR is ready for another review, let me know using the "re-request review" button (circular arrows) in the reviewers section of the right sidebar area!

CommanderStorm:
> Yes, having both a `.snap.new` and a `.snap` file was what I would love to make sure we prevent

When the testcase sees such a case, it will pass without failiour and remove `.snap.new` if the content is identical.
If not, it will fail and show the diff.

I added a simple testcase to check for the existance of `.snap.new` and show an appropriate errormessage.

image

CommanderStorm:
> it would be a great idea if you were to run through the lint-writing process end-to-end and experience it as a new user would. That way, any issues that come up can be fixed while all this is still fresh in our minds and before any new contributors run into problems.

Shure can try myself the weekend.
Do you think this one is a good fit as a simple, straigtforeward lint?

- `a public API enum's variant used to have a well-defined discriminant, but no longer does` https://github.com/obi1kenobi/cargo-semver-checks/issues/898

obi1kenobi (changes requested):

obi1kenobi:
> Shure can try myself on the weekend. Do you think one of these is a good fit as a simple, straigtforeward lint?
> 
> * `a public API enum's variant used to have a well-defined discriminant, but no longer does` [New lints: Changes to enum variant discriminants #898](https://github.com/obi1kenobi/cargo-semver-checks/issues/898)
> * [New lint: public API struct became enum or union #954](https://github.com/obi1kenobi/cargo-semver-checks/issues/954)

Either of those lints is fine, but only if you are willing and able to thoroughly self-review your code in GitHub's UI before marking the PR as open for reviews. Please don't make me point out obvious issues with the code that you could have found & fixed in less time than it'll take me to write a review comment.

CONTRIBUTING.md
```diff
@@ -171,42 +202,76 @@ and probably isn't working quite right.
...
+thread 'query::tests_lints::enum_struct_variant_field_added' panicked at /home/frank/.cargo/registry/src/index.crates.io-6f17d22bba15001f/insta-1.40.0/src/runtime.rs:548:9:
 ```
 
-Inspect the "actual" output:
+Inspect the generated "actual" output in the `.snap.new` file:
 - Does it report the semver issue your lint was supposed to catch? If not, the lint query
   or the test crates' code may need to be tweaked.
 - Does it report correct span information? Is the span as specific as possible, for example
-  pointing to a struct's field rather than the whole struct if the lint refers to that field?
+  pointing to a `struct`'s field rather than the whole struct if the lint refers to that field?
```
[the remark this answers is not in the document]
CommanderStorm: yea, sorry thought of this as a typo-fix

CONTRIBUTING.md
```diff
@@ -171,42 +202,76 @@ and probably isn't working quite right.
...
 - Does the output also report any code from test crates other than `test_crates/<lint_name>`?
   If so, ensure the reported code is indeed violating semver and is not being flagged
   by any other lint.
 
-If everything looks okay, edit your `test_outputs/<lint_name>.output.ron` file adding
-the "actual" output, then re-run `cargo test` and make sure everything passes.
+If everything looks okay, either run `cargo insta review` (see the [snapshot instructions](#what-are-those-snap-or-snapnew-files-generated-via-cargo-test-) for context) or manually move `test_outputs/query_execution/<lint_name>.snap.new` to  `test_outputs/query_execution/<lint_name>.snap`.
```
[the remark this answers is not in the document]
CommanderStorm: Interesting, seems like github and rustrover handle this part of markdown differently
(github removes trailing dashes, rustrover wants them)

image
image

tests/integration_snapshots.rs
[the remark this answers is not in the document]
CommanderStorm: I have moved the testcase to `src/snapshot_tests`
The file seemed like the best place to put this testcase as I only searched in the `tests` directory for tests. Creating another module for this seemed not quite something that I guessed you'd approve 😓

The tight formatting requirements around matching other code are really hard to match, given that there are a lot of unwritten rules for this.
This is something which seems obvious to you (given that you have seen all of the codebase) but does not look wrong when I review my code, sorry 😅 

And no, this is just me (maybe a bit tired, but that is not a good excuse), no AI.
Maybe I should let PRs rest and look at them with a fresh head on the next day to prevent this, but I cannot guarantee that I would have caught on to the differenece in for example `.unwrap` vs `.expect("failed to..` in the file..
Likely that is just experience of working with rust. Sorry

obi1kenobi (approved):
Thanks, this is much better. There's still room for growth, and as long as you are working hard on improving your skills, I'm happy to keep pointing you in the right direction.

PR #952, opened 2024-09-27
docs: add `<details>` to the `CONTRIBUTING.md` to make some content visually shorter

CommanderStorm:
During editing the contribution guide, I noticed that some parts of the contribution guide are likely not immediately necessary for all users. Let's hide some content (where it makes sense).

[Rendered](https://github.com/obi1kenobi/cargo-semver-checks/blob/f6f608c2c688d3762dc24406be7a68bd60402457/CONTRIBUTING.md)

What do you think about this?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -168,9 +168,8 @@ If it didn't fail, then your lint didn't report any semver issues in the test cr
 and probably isn't working quite right.
 (Did any other lints' tests fail also? See the [troubleshooting item](#troubleshooting-other-lints-tests-failed-too) below.)
 
-<details><summary>For a lint named <code>enum_struct_variant_field_added</code>, you'll probably see its test fail with
-a message similar to this</summary>
-
+For a lint named `enum_struct_variant_field_added`, you'll probably see its test fail with
+a message similar to this:
 ```
 Query enum_struct_variant_field_added produced incorrect output (./src/lints/enum_struct_variant_field_added.ron).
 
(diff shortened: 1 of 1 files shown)


obi1kenobi:
Thanks for suggesting it, it's a good idea!

Details sections have a tendency to be "not obviously clickable" when considering a broad range of contributor skill levels, which in turn makes the contributing doc less contributor-friendly. I think with small tweaks, we can largely avoid this risk and end up with a better contributing file.

I don't like hiding the test output since it's necessary for people to see. I'd prefer we leave that out of a details section.

I'm open to using details blocks to hide the two troubleshooting items, but I'd prefer to add `(click to expand)` at the end of the summary line so as to minimize the risk of people not knowing how to use them.

CommanderStorm:
Reasonable, have changed as suggested ^^

CONTRIBUTING.md
```diff
@@ -291,51 +291,56 @@ otherwise the test will fail in CI.
...
+  
+  So if you added code to a test crate, and it caused other lints to report new findings, consider:
+  - whether your code indeed contains the reported semver issue;
+  - whether the same semver issue is being reported only once, and not multiple times
+    by different lints,
+  - and whether the new reported lint result points to the correct item and span information.
+  
+  If the answer to all is yes, then everything is fine! Just edit those other lints'
 expected output files to include the new items, and you can get back on track.
+
```
obi1kenobi, line 332:
```suggestion
```

obi1kenobi:
Thank you!

PR #956, opened 2024-09-29
tests: make sure that unrelated snapshots are passing CI

CommanderStorm:
These changes seem nessesary from my point of view via running `cargo test`.
Lets see if CI agrees with me and something has changed between https://github.com/obi1kenobi/cargo-semver-checks/commit/b65ccab29af3039b5ac14dd09348385b4b6667af and now or if my environment is faulty in this regard

### DIFF
--- a/test_outputs/integration_snapshots__bugreport.snap
+++ b/test_outputs/integration_snapshots__bugreport.snap
@@ -7,6 +7,7 @@ info:
     - "--bugreport"
   env:
     CARGO_TERM_COLOR: never
+    RUST_BACKTRACE: "0"
 ---
 success: true
 exit_code: 0
--- a/test_outputs/integration_snapshots__bugreport.snap
+++ b/test_outputs/integration_snapshots__bugreport.snap
@@ -7,7 +7,6 @@ info:
     - "--bugreport"
   env:
     CARGO_TERM_COLOR: never
-    RUST_BACKTRACE: "0"
 ---
 success: true
 exit_code: 0
--- a/test_outputs/integration_snapshots__unstable_options_without_flag.snap
+++ b/test_outputs/integration_snapshots__unstable_options_without_flag.snap
@@ -5,6 +5,9 @@ info:
   args:
     - semver-checks
     - "--witness-hints"
+  env:
+    CARGO_TERM_COLOR: never
+    RUST_BACKTRACE: "0"
 ---
 success: false
 exit_code: 1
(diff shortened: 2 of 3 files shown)


CommanderStorm:
How is CI passing on this, but also https://github.com/obi1kenobi/cargo-semver-checks/pull/957
Something is off here, will investigate with a fresh mind tomorrow though

obi1kenobi:
Is it possible you have some kind of cargo config that isn't in environment variables? Perhaps a `.cargo/config.toml` file somewhere?

[These lines](https://github.com/obi1kenobi/cargo-semver-checks/blob/85372ed24aa154d95dc64e8ae5fa6c3c12d492c4/src/snapshot_tests.rs#L139-L144) currently attempt to remove any non-reproducible configs, but perhaps they need to be adjusted to *override* env var values instead of merely removing them. This is because removals will make cargo fall back to other config, whereas overriding the env vars with desirable values will force cargo to explicitly apply the behavior we want instead of using default-or-otherwise-configured settings.

Could you see if changing those lines to explicitly set the env vars you've identified as necessary will fix the problem?

CommanderStorm:
can't really find why setting the environment varibles does not change the output of the snapshots.
This does quite make sense to me => is something which needs a more knowlegable engineer 🤷🏻‍♂️

obi1kenobi:
My guess is this.

On CI, the `ci.yml` workflow sets some environment variables, and there are no further config settings in any `.cargo/config.toml` file or similar. Those env vars are removed [here](https://github.com/obi1kenobi/cargo-semver-checks/blob/85372ed24aa154d95dc64e8ae5fa6c3c12d492c4/src/snapshot_tests.rs#L139-L144). As a result, the test runs without backtraces and works fine.

On your local system, you have a `.cargo/config.toml` file _somewhere_ (home dir, inside the project, [any of the locations cargo will accept](https://doc.rust-lang.org/cargo/reference/config.html)) that sets some of these config keys. You don't set any env vars, so there's nothing to remove when the test tries to remove them as on the `main` branch. Your tests then run with backtraces, and fail with a snapshot failure.

Env vars take precedence over config file settings, so setting the env vars fixes the problem you're seeing. On CI, it has no effect because the set env vars are applying values that are already the defaults.

I would go through the paths mentioned here and look through any files they contain to find where the configs are set: https://doc.rust-lang.org/cargo/reference/config.html

CommanderStorm:
Sorry, that is not the case.

image

I have no clue..
(neither a config file, nor environement variables on my side.. nothing)

PR #960, opened 2024-10-01
Add lint group infra

suaviloquence:
Adds:

- the `Identifier` enum for consistent configuration with lint groups and lints
- `OverrideStack::try_push` which tries to compile a `Map<Identifier, Override>` to a `Map<String (= lint id), Override>`, and errors on a conflict
- the `must_use_added` lint group
 - currently the `lint_group` field on `SemverQuery` is optional while we figure out what lints go in what groups (and what groups to include).  `must_use_added` is an example from my design draft, but we can definitely go in a different direction.  Eventually the goal is to make `lint_group` not an Option, so every lint is associated with exactly one lint group
- some tests + documentation, and user documentation in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,6 +432,47 @@ trait_must_use_added = "allow"
 union_must_use_added = "allow"
 ```
 
+#### Configuration Priority
+
+When there are multiple conflicting entries for a given lint (for example, configuring the
+`must_use_added` group containing the `function_must_use_added` lint and `function_must_use_added` itself),
+`cargo-semver-checks` needs more information to determine which configuration entry takes
+precedence. Like the Cargo `[lints]` table, each configuration entry has an optional
+`priority` field that defines the order of configuration entries. If an entry has no
+`priority` key, its priority is 0 by default. Entries with a lower (more negative)
+`priority` override those with a higher `priority` when there is a conflict.
+
+For example:
+
+```toml
+must_use_added = { level = "warn", required-update = "major" }
+function_must_use_added = { level = "deny" }
+```
+
+`cargo-semver-checks` will raise an error if invoked with this configuration, because there is a
+conflict for the `level` field in `must_use_added`. Let's use the `priority` field to describe
+the configuration we want:
+
+```diff,toml
+- function_must_use_added = { level = "deny" }
++ function_must_use_added = { level = "deny", priority = -1 }
+```
+
+By adding a `priority` of -1, the `function_must_use_added` entry will override the `must_use_added`
+(which has the default priority of 0) when there is a conflict. This will result in a final configuration
+of:
+
+```toml
+enum_must_use_added = { level = "warn", required-update = "major" }
+struct_must_use_added = { level = "warn", required-update = "major" }
+# ...
+function_must_use_added = { level = "deny", required-update = "major" }
+```
+
+Note that `function_must_use_added` still has a `major` required update bump, because the entry with
+priority -1 left the `required-update` to its default value, so `cargo-semver-checks` will use the
+entry with less precedence to calculate the required version bump.
+
 #### Implementation details & limitations
 
 When checking a package, `cargo-semver-checks` uses the manifest of the current subject's `Cargo.toml` file (plus its workspace configuration, if opted in).
--- a/src/query.rs
+++ b/src/query.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeMap;
+use std::{collections::BTreeMap, sync::OnceLock};
 
 use ron::extensions::Extensions;
 use serde::{Deserialize, Serialize};
@@ -103,6 +103,10 @@ pub struct SemverQuery {
     /// The default lint level for when this lint occurs.
     pub lint_level: LintLevel,
 
+    /// The [`LintGroup`] that this query is a member of (currently optional).
+    #[serde(default)]
+    pub lint_group: Option<LintGroup>,
+
     #[serde(default)]
     pub reference: Option<String>,
 
@@ -159,6 +163,9 @@ impl SemverQuery {
     }
 }
 
+// TODO: replace with LazyLock once MSRV is >= 1.80
+pub(crate) static ALL_QUERIES: OnceLock<BTreeMap<String, SemverQuery>> = OnceLock::new();
+
 /// Configured values for a [`SemverQuery`] that differ from the lint's defaults.
 #[derive(Debug, Clone, Default, PartialEq, Eq, Deserialize)]
 #[serde(rename_all = "kebab-case")]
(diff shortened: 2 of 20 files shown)


src/query.rs
```diff
@@ -183,8 +190,42 @@ pub struct QueryOverride {
...
+    /// Identifies a group of lints, based on the [`SemverQuery::lint_group`] field.
+    Group(LintGroup),
+    /// Identifies a singular lint by id.
+    Lint(String),
+}
+
+impl Identifier {
+    /// Returns a list of lint ids that this identifier refers to, given the set of
+    /// all queries.
+    pub(crate) fn lints<'a>(
```
suaviloquence, line 215: If/when we add dynamic groups like `warnings`, `major`, and `minor`, this will need an extra `&OverrideStack` argument to determine this.  We call this method in `OverrideStack::try_push`, so this will be an easy change.

src/query.rs
```diff
@@ -183,8 +190,42 @@ pub struct QueryOverride {
...
+        &'a self,
+        all_queries: &'a BTreeMap<String, SemverQuery>,
+    ) -> Vec<&'a str> {
```
suaviloquence, line 218: We could also make this a `Vec<&'a SemverQuery>`, but we would need to handle the case where an unknown lint id was provided (but this may be a good thing, currently there is no warning for this)

src/query.rs
```diff
@@ -206,8 +247,45 @@ impl OverrideStack {
...
+    pub fn try_push(&mut self, item: &OverrideMap) -> anyhow::Result<()> {
+        let mut compiled_map = BTreeMap::<_, QueryOverride>::new();
+
+        for (id, overrides) in item {
+            let lints = id.lints(ALL_QUERIES.get_or_init(SemverQuery::all_queries));
+
+            for lint in lints {
+                if let Some(entry) = compiled_map.get_mut(lint) {
+                    if let Some(lint_level) = overrides.lint_level {
+                        if let Some(old) = entry.lint_level {
```
suaviloquence, line 267: Currently it is an error even if `lint_level == old`.  Should this be the case? Maybe just a warning?

README.md
```diff
@@ -432,6 +432,47 @@ trait_must_use_added = "allow"
...
+```toml
+enum_must_use_added = { level = "warn", required-update = "major" }
+struct_must_use_added = { level = "warn", required-update = "major" }
+# ...
+function_must_use_added = { level = "deny", required-update = "major" }
+```
+
+Note that `function_must_use_added` still has a `major` required update bump, because the entry with
+priority -1 left the `required-update` to its default value, so `cargo-semver-checks` will use the
+entry with less precedence to calculate the required version bump.
```
obi1kenobi, line 474: We need a succinct, memorable way to describe this algorithm.

How about something like:
```suggestion
The final values for the `level` and `required-update` fields are computed separately, each in priority order.

If the entry with the most-negative `priority` doesn't set both `level` and `required-update` values, the final configuration may have elements of multiple config entries. This happens
in the above example, where `function_must_use_added` still has a `major` required update bump:
- The entry with priority -1 did not override the `required-update` field. We keep looking.
- In priority order, the next entry is the one with priority 0. It set `required-update = "major"`.
- The lint uses `required-update = "major"`.
```

README.md
```diff
@@ -432,6 +432,47 @@ trait_must_use_added = "allow"
...
 
+#### Configuration Priority
+
+When there are multiple conflicting entries for a given lint (for example, configuring the
+`must_use_added` group containing the `function_must_use_added` lint and `function_must_use_added` itself),
+`cargo-semver-checks` needs more information to determine which configuration entry takes
+precedence. Like the Cargo `[lints]` table, each configuration entry has an optional
+`priority` field that defines the order of configuration entries. If an entry has no
+`priority` key, its priority is 0 by default. Entries with a lower (more negative)
+`priority` override those with a higher `priority` when there is a conflict.
```
obi1kenobi, line 443: Consider explaining first, then providing an example, rather that doing both at once. Weaving lint groups and a specific lint into the middle of the explanation leaves readers with a lot of information to track.

Consider adapting the explanation I used below if you decide you like it. Something along the lines of "for each setting in each lint, we inspect config items in priority order and pick the first one that sets a value for that setting."

src/query.rs
```diff
@@ -164,6 +168,9 @@ impl SemverQuery {
     }
 }
 
+// TODO: replace with LazyLock once MSRV is >= 1.80
+pub(crate) static ALL_QUERIES: OnceLock<BTreeMap<String, SemverQuery>> = OnceLock::new();
```
obi1kenobi, line 172: I really dislike globals like this. They make testing a mess.

Can you think of any reasonable way we can avoid this?

src/query.rs
obi1kenobi: `Identifier` doesn't quite strike me as the right place for this logic. We're giving the larger pool of data (`&'a BTreeMap<String, SemverQuery>`) to the smaller type (a two-variant enum) and asking it to figure it out.

Also, I feel like we want the "resolve the identifiers" logic to have `Result` types so we can complain about bad IDs more explicitly. And we really need to test those cases before someone gets bitten by a silent bug around bad IDs...

src/query.rs
obi1kenobi: When might this happen?

If I'm understanding it correctly, there's no conflict since both sides are setting the same value, no?

src/query.rs
```diff
@@ -206,8 +247,45 @@ impl OverrideStack {
...
     /// The inserted overrides will take precedence over any lower item in the stack,
     /// if both maps have a not-`None` entry for a given lint.
-    pub fn push(&mut self, item: &OverrideMap) {
-        self.0.push(item.clone());
+    ///
+    /// # Errors
+    ///
+    /// Returns an `Err` if there is a configuration conflict in the passed map: when
+    /// multiple entries in `item` configure the same entry for the same lint. To handle
+    /// configuration precedence in a defined way, call this function multiple times, and
```
obi1kenobi, line 255: `defined way` is ominously vague here, and it's in the middle of a very long sentence which makes me feel it's unlikely I'd manage to use this function properly =/

src/query.rs
```diff
@@ -206,8 +247,45 @@ impl OverrideStack {
...
+        let mut compiled_map = BTreeMap::<_, QueryOverride>::new();
+
+        for (id, overrides) in item {
+            let lints = id.lints(ALL_QUERIES.get_or_init(SemverQuery::all_queries));
+
+            for lint in lints {
+                if let Some(entry) = compiled_map.get_mut(lint) {
+                    if let Some(lint_level) = overrides.lint_level {
+                        if let Some(old) = entry.lint_level {
+                            anyhow::bail!("Configuration conflict detected for lint `{lint}`:\n\
```
obi1kenobi, line 268: I *think* our bail errors are generally lowercase since `bail!(<stuff>)` gets printed as `error: <stuff>` when the program exits. If I'm right about that, any capitalized error messages should be tested + changed.

src/query.rs
```diff
@@ -206,8 +247,45 @@ impl OverrideStack {
...
+                                lint level is set to `{old:?}` and {lint_level:?} at the same config priority.");
```
obi1kenobi, line 269: This is unfortunately a supremely non-useful error message: it tells the user "go figure it out" without actually saying which entries are the ones at that priority. The entries might not be easy to find, since they might be on the lint itself, or any of its groups.

This would be very frustrating to encounter.

test_outputs/snapshot_tests/cargo_semver_checks__snapshot_tests__configuration_conflict-output.snap
```diff
@@ -0,0 +1,11 @@
+---
+source: src/snapshot_tests.rs
+expression: result
+---
+--- error ---
+add a `priority` key to the entries in [package.metadata.cargo-semver-checks.lints].
```
obi1kenobi, line 6: Not a great error message.

Imagine the user's `Cargo.toml` has 100 config entries. Suddenly, one of them caused a conflict. (Maybe it's because *we* added a new lint that is the first one to be in a particular combination of two groups simultaneously? The user had specified settings for each group, and they didn't conflict in our _previous_ release. In the new release with the new lint, they do! Oops!)

Now we just told the user "hey please add `priority` to all your 100 config entries." They'll uninstall the tool immediately.

obi1kenobi (commented):
Unfortunately, I think this needs a lot more design on the UX side of things first, before we get into the weeds of how it's implemented under the hood.

- We need to explicitly enumerate which error cases happen when.
- We need super high-quality error messages in each case, pointing to the specific places that caused the problem and offering concrete next steps.
- We need to consider what happens when we add more lint groups, add new lints to various combinations of lint groups, or newly add existing lints to existing groups. Edge cases are everywhere!
- We need to consider the "upgrade experience" of `cargo-semver-checks. If we aren't extremely careful, adding a new lint will be a major breaking change and will anger our biggest users first — the people we want to keep the most happy.

I think taking a step back from the code and producing something more akin to a design doc would be a great next step.

PR #962, opened 2024-10-05
feat: add lint `struct_with_no_pub_fields_changed`

### LINKED ISSUE #954: New lint: public API struct became enum or union
Inspired by this post: https://internals.rust-lang.org/t/rest-patterns-foo-should-match-non-struct-types/21607

TL;DR: Given a type `Example`, the pattern in `matches!(value, Example { .. })` is only valid if `Example` is a struct. It fails on both enums and unions at the moment: [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b7fc5c8589620a247846e82e8788c304)

We already have a similar lint for changing a public API struct _with public fields_ into an enum or union: [`struct_with_pub_fields_changed_type`](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/src/lints/struct_with_pub_fields_changed_type.ron). The breakage there is more compelling than "empty pattern match" so we should keep that lint. We should make the new lints disjoint from it, so that users see one or the other but never both.

We also have a lint for constructible structs changing to another type: [`constructible_struct_changed_type`](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/src/lints/constructible_struct_changed_type.ron). The breakage there is also more compelling than "empty pattern match" so we should keep that lint too. The new lints should be disjoint from it as well.

The new lints here should therefore be:
- [x] public API struct with _no public fields_ is no longer a struct: #962 
- [ ] public API `#[non_exhaustive]` struct with no fields of any kind is no longer a struct: #963

The new lints will be quite similar to the related lints mentioned above. Make sure to check out our [contributing guide](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md), and take a look at what [prior lints' merged PRs looked like](https://github.com/obi1kenobi/cargo-semver-checks/pull/844) so you know what to expect. It's easier than you think!

Thanks to @jhpratt for bringing this to my attention!

CommanderStorm:
~~This draft is mostly to prevent others thinking that nobody is working on this issue.~~

The query is mostly from `struct_with_pub_fields_changed`

There is still some tinkering I need to do with
- [x] review the wording of the description => https://github.com/obi1kenobi/cargo-semver-checks/pull/962#discussion_r1789082413
- [x] review that the whitness actually generates correct code
- [x] I need to self review. Learning from the semi-desaster in https://github.com/obi1kenobi/cargo-semver-checks/pull/951, I am going to do that tomorrow though.
- [x] See if I can fix the false-negative for `PubStructWithNonPubDocFieldChangedToUnion` ~~or mark it as `TODO` => https://github.com/obi1kenobi/cargo-semver-checks/pull/962#discussion_r1788714447~~

Resolves one checkbox of #954


src/lints/struct_with_no_pub_fields_changed.ron
```diff
@@ -0,0 +1,77 @@
...
+    arguments: {
+        "public": "public",
+        "true": true,
+        "zero": 0,
+    },
+    error_message: "A struct without pub fields became an enum or union, breaking pattern matching.",
+    per_result_error_template: Some("struct {{join \"::\" path}} became {{lowercase current_typename}} in file {{span_filename}}:{{span_begin_line}}"),
+    witness: (
+        hint_template: r#"matches!(value, {{join "::" path}}{...});"#,
+        witness_template: r#"matches!(value, {{join "::" path}}::Variant);"#,
```
CommanderStorm, line 75: Bit unsure if I should fill out both fields here and if their usage (what should I fill into each of the both fields) is.
`hint_template` is clear enough, but I don't get what `witness_template` is.

Need to review what the original PR intendeded to put here.
Since no other lint uses `witness_query` or `witness_template` this is a bit difficult ^^

Likely including a sentence what `witness_template` does in relation to `hint_template` would be helpfull in the contribution guide.
Likely this is [this todo](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md#full-witness-templates).

src/lints/struct_with_no_pub_fields_changed.ron
```diff
@@ -0,0 +1,77 @@
...
+                item {
+                    ... on Struct {
+                        struct_typename: __typename @tag @output
+                        visibility_limit @filter(op: "=", value: ["$public"]) @output
+                        struct_type @output
+
+                        # Ensure the struct does not have pub fields to prevent overlap with struct_with_pub_fields_changed
+                        field @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
+                            visibility_limit @filter(op: "=", value: ["$public"])
+                            public_api_eligible @filter(op: "=", value: ["$true"])
```
CommanderStorm, line 29: `public_api_eligible` is giving me a bit of (understandable) trouble understanding.

Do you think that https://predr.ag/blog/checking-semver-for-doc-hidden-items/ should be linked in the contribution guide?

On that note:
Is this correctly handled here?
My feeling is that `public_api_eligible` would need to be
- handled for the struct,
- the pub fields (hiding => not pub)
- and the the non-pub fields (hiding => not pub)
  
The last one is giving me a bit of grief, as the filters seem to be `and`-ed. I need it to be 
- `visibility_limit @filter(op: "!=", value: ["$public"])`
- OR `public_api_eligible @filter(op: "!=", value: ["$true"])`

Is there some docs or other lints which are using filters in this pattern?

src/lints/struct_with_no_pub_fields_changed.ron
suaviloquence: `witness_template` is not implemented yet, so you can leave it out for now.  it will be a full buildable program that we can use to prove that downstream could break with this change. 

```suggestion
        hint_template: r#"matches!(value, {{join "::" path}}{...});"#,
```

src/lints/struct_with_no_pub_fields_changed.ron
```diff
@@ -0,0 +1,76 @@
+SemverQuery(
+    id: "struct_with_no_pub_fields_changed",
+    human_readable_name: "public API struct with no public fields is no longer a struct",
+    description: "A struct without pub fields was converted into an enum or union, breaking pattern matching.",
```
CommanderStorm, line 4: should the description rather be 
```suggestion
    description: "A struct without pub fields, but with non-pub fields was converted into an enum or union, breaking pattern matching.",
```

CommanderStorm:
(idk if you get notifications for this, assuming you don't)

@obi1kenobi I have reviewed this as thoroughly as possible. Except for the comments above I have not found anything. => I think this is ready for review

src/lints/struct_with_no_pub_fields_changed.ron
```diff
@@ -0,0 +1,76 @@
...
+    required_update: Major,
+    lint_level: Deny,
+    reference_link: Some("https://internals.rust-lang.org/t/rest-patterns-foo-should-match-non-struct-types/21607"),
+    reference: Some(
+        r#"\
+Even if a structs does not expose pub fields, pattern matching like `matches!(value, Example { .. })` is accessible outside their defining crate. \
+Changing such a struct to make it an enum or union will break this pattern match, as their syntax is different.
```
obi1kenobi, line 11: Fix `structs -> struct` typo and simplify language a bit ("defining crate" might confuse some users). I also suggest removing "their syntax is different" since it isn't clear what "their" refers to: pattern matches, breaking changes, structs, enums, unions, or some combination of them.
```suggestion
Even if a struct does not expose pub fields, pattern matching like `matches!(value, Example { .. })` is allowed outside the struct's own crate. \
Changing such a struct into an enum or union will break such pattern matching.
```

src/lints/struct_with_no_pub_fields_changed.ron
obi1kenobi: Probably not necessary, I think. So long as the lint always fires when it should, I don't think all the edge cases have to be enumerated in the user-facing text. That sounds probably more confusing than helpful.

src/lints/struct_with_no_pub_fields_changed.ron
```diff
@@ -0,0 +1,76 @@
+SemverQuery(
+    id: "struct_with_no_pub_fields_changed",
```
obi1kenobi, line 2: Let's follow the naming convention. All other "struct changed to enum or union" lints end in "changed_type" so let's do that here as well. Just "changed" is vague and doesn't communicate what happened.

Remember to update the file name, the ID here, and the test crate.

src/lints/struct_with_no_pub_fields_changed.ron
obi1kenobi: > `public_api_eligible` is giving me a bit of (understandable) trouble understanding.
> 
> Do you think that https://predr.ag/blog/checking-semver-for-doc-hidden-items/ should be linked in the contribution guide?

Possibly. It's _usually_ not that relevant to most lints (this one is an exception) so I wouldn't make it a "must read". But if you can find a good way to tuck in in as "extra reading" or something, that'd be good. Open a PR?

> On that note: Is this correctly handled here? My feeling is that `public_api_eligible` would need to be
> 
> * handled for the struct,
> * the pub fields (hiding => not pub)
> * and the the non-pub fields (hiding => not pub)
> 
> The last one is giving me a bit of grief, as the filters seem to be `and`-ed. I need it to be
> 
> * `visibility_limit @filter(op: "!=", value: ["$public"])`
> * **OR `public_api_eligible @filter(op: "!=", value: ["$true"])`**
> 
> using OR there or adding the counts of both would allow to remove the false negative for the testcase `PubStructWithNonPubDocFieldChangedToUnion`.

I think this is equivalent to "has some fields, but has no public API fields" right?

`constructible_struct_changed_type` requires that the struct have no fields at all, and also that it not be `#[non_exhaustive]`. So any fields are sufficient to escape that lint.

Which raises a good point: we probably want to catch the case of `#[non_exhaustive]` structs with no fields with a separate lint, since that will escape this lint's implementation. Would you like to take care of that lint next?

> Is there some docs or other lints which are using filters in this pattern? How wrong is my intuition on this topic?

Your intuition is pretty good! We don't have "OR" operators or addition operators yet, though I have a WIP branch for that in Trustfall so hopefully I can get it merged by the end of the year.

test_crates/struct_with_no_pub_fields_changed/new/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+pub enum PubStructWithNonPubDocFieldChangedToEnum {
+    Foo(usize),
+}
+
+
+pub union PubStructWithNonPubDocFieldChangedToUnion {
+    /// Despite this field being pub, hiding it makes this not be `public_api_eligible` anymore
+    /// This struct should trigger `struct_with_no_pub_fields_changed` instead of `struct_with_pub_fields_changed`
```
obi1kenobi, line 16: Lint name correction:
```suggestion
    /// This struct should trigger `struct_with_no_pub_fields_changed` instead of `struct_with_pub_fields_changed_type`
```

Please also add a ruler to your editor at 100 chars and try to not exceed it unless there's good reason. Hard enforcement of line lengths is annoying because e.g. some URLs are just long and there's no much we can do about it. But this line is 110+ chars wide and there's no reason why it has to be that long all on one line.

obi1kenobi (commented):
Nice work! Especially good job with the query logic, it was trickier than I thought it would be.

There are a handful of remaining items that require another pass before we can merge. Most of them are the kinds of things that would be good to practice catching in your self-reviews: the top difference between a junior engineer and a senior engineer is whether one independently finds the issues in one's own code, or whether someone else has to point them out. So let's work on it together!

src/lints/struct_with_no_pub_fields_changed.ron
CommanderStorm: Idk why, I had in my head that these changes were just `_changed` from the initial looking through other lints.

src/lints/struct_with_no_pub_fields_changed.ron
CommanderStorm: > I think this is equivalent to "has some fields, but has no public API fields" right?

Yes, that works, was too focused on my path => did not zoom out to see that.

obi1kenobi: No worries. Practice makes perfect, and I have a lot of practice doing this :)

src/lints/struct_with_no_pub_fields_changed_type.ron
```diff
@@ -0,0 +1,76 @@
...
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+        "zero": 0,
+    },
+    error_message: "A struct without pub fields became an enum or union, breaking pattern matching.",
+    per_result_error_template: Some("struct {{join \"::\" path}} became {{lowercase current_typename}} in file {{span_filename}}:{{span_begin_line}}"),
+    witness: (
+        hint_template: r#"matches!(value, {{join "::" path}}{...});"#,
```
obi1kenobi, line 74: The pattern that is broken is `StructName { .. }` specifically. The dots are not a textual ellipsis, they are a Rust syntactic element signifying "arbitrary fields."
```suggestion
        hint_template: r#"matches!(value, {{join "::" path}} {..});"#,
```

obi1kenobi (commented):
Looks good modulo a small bug in the witness hint. If you can regenerate the witness file with the fix, this should be good to merge!

obi1kenobi (approved):
Awesome, nicely done 🚀

Issue #946, opened 2024-09-26
New lints: Breaking changes in macros


obi1kenobi:
There are several ways to cause a breaking change by changing macro item definitions that we can lint for today:
- [x] removing a declarative macro that used to exist (#966)
- [x] removing a function-like proc macro that used to exist (#1021)
- [x] removing an attribute proc macro that used to exist (#1026)
- [x] removing a derive proc macro that used to exist (#1027)
- [x] removing a [helper attribute](https://doc.rust-lang.org/reference/procedural-macros.html#derive-macro-helper-attributes) from a derive proc macro (#1028)
- [x] declarative macro became `#[doc(hidden)]` (#1029)
- [x] declarative macro is no longer `#[macro_export]` but is not `#[doc(hidden)]` -- variant of #966 in that the macro can't be used, but it's still defined, just local to its defining crate (#1031)
- [x] proc macro became `#[doc(hidden)]` (#1050)

Relevant docs we can use as references in the lints:
- declarative macros: https://doc.rust-lang.org/reference/macros-by-example.html
- procedural (proc) macros: https://doc.rust-lang.org/reference/procedural-macros.html

The lints here will be similar to existing "item missing" lints (e.g. [enum missing](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/src/lints/enum_missing.ron), [enum variant missing](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/src/lints/enum_variant_missing.ron)), with one important difference: macros are always at the top level of the crate that defined them. They don't have importable paths, so our matching doesn't need the `importable_path` edge like for other items and can instead simply rely on the macro's own name.

Make sure to check out our [contributing guide](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md), and take a look at what [prior lints' merged PRs looked like](https://github.com/obi1kenobi/cargo-semver-checks/pull/844) so you know what to expect. It's easier than you think!

miikka:
I'll give a go at implementing "removing a declarative macro that used to exist".

PR #966, opened 2024-10-10
Add lint `declarative_macro_missing`

miikka:
Working to implement the first item in #946: "removing a declarative macro that used to exist".

There's this case as well, but I'm not sure if it can be currently queried for:

```rust
#[macro_export]
macro_rules! inner_pub_use_will_be_removed {
    () => {};
}

pub mod inner {
    // Is it a breaking change if you remove the next line?
    pub use crate::inner_pub_use_will_be_removed;
}
```

### DIFF
--- a/src/lints/declarative_macro_missing.ron
+++ b/src/lints/declarative_macro_missing.ron
@@ -26,7 +26,6 @@ SemverQuery(
                 item {
                     ... on Macro {
                         visibility_limit @filter(op: "=", value: ["$public"])
-                        public_api_eligible @filter(op: "=", value: ["$true"])
                         name @filter(op: "=", value: ["%name"])
                     }
                 }
--- a/src/lints/declarative_macro_missing.ron
+++ b/src/lints/declarative_macro_missing.ron
@@ -1,0 +1,46 @@
+SemverQuery(
+    id: "declarative_macro_missing",
+    human_readable_name: "macro_rules declaration removed or renamed",
+    description: "A declarative macro marked with #[macro_export] can no longer imported by its prior name.",
+    required_update: Major,
+    lint_level: Deny,
+    reference_link: Some("https://doc.rust-lang.org/reference/macros-by-example.html#path-based-scope"),
+    query: r#"
+    {
+        CrateDiff {
+            baseline {
+                item {
+                    ... on Macro {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        public_api_eligible @filter(op: "=", value: ["$true"])
+                        name @output @tag
+
+                        span_: span @optional {
+                            filename @output
+                            begin_line @output
+                        }
+                    }
+                }
+            }
+            current @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
+                item {
+                    ... on Macro {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        public_api_eligible @filter(op: "=", value: ["$true"])
+                        name @filter(op: "=", value: ["%name"])
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "zero": 0,
+        "true": true,
+    },
+    error_message: "A publicly-visible `macro_rules` declarative macro cannot be imported by its prior name. A `#[macro_export]` may have been removed, or the macro itself may have been renamed or removed entirely.",
+    per_result_error_template:  Some("macro_rules! {{name}}, previously in file {{span_filename}}:{{span_begin_line}}"),
+    witness: (
+        hint_template: r#"{{name}}!(...);"#
+    ),
+)
--- a/test_crates/declarative_macro_missing/new/src/lib.rs
+++ b/test_crates/declarative_macro_missing/new/src/lib.rs
@@ -1,0 +1,3 @@
+macro_rules! will_no_longer_be_exported {
+    () => {};
+}
--- a/test_crates/declarative_macro_missing/new/src/lib.rs
+++ b/test_crates/declarative_macro_missing/new/src/lib.rs
@@ -1,3 +1,15 @@
 macro_rules! will_no_longer_be_exported {
     () => {};
 }
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! became_doc_hidden {
+    () => {};
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! always_doc_hidden {
+    () => {};
+}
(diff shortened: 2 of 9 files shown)


miikka:
Not sure how to address the beta failure...

```
The referenced type does not exist in the schema: Macro
```

obi1kenobi:
It might need a `cargo update` to pick up the latest schema.

miikka:
I tried `cargo update` and it didn't help. I assume the point would be to update the versions of trustfall-rustdoc-adapter? It didn't touch the versions of any of those.

obi1kenobi:
Hmm interesting, I might have messed something up then. I'll take a look
when I get back home later this week, sorry about that.

obi1kenobi:
My bad, I got everything ready then just failed to cut the release 🙈 I'll have a new release up momentarily, and I'll take care of updating your PR to merge it to so I can atone for sending you down a debugging rabbit hole unnecessarily...

Excellent catch on re-exporting macros as well! To my surprise, re-exporting does seem to work: [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bbe306631898fee8ffad61b4028d333a)
```rust
#[macro_export]
macro_rules! inner_pub_use_will_be_removed {
    () => {};
}

pub mod inner {
    /// We can re-export macros at non-root paths, watch:
    /// ` ``rust
    /// use playground::inner::inner_pub_use_will_be_removed;
    ///
    /// inner_pub_use_will_be_removed! {}
    /// ` ``
    pub use crate::inner_pub_use_will_be_removed;
}
```

I've opened #968 to track this issue, which you are also welcome to grab if you're interested!

src/lints/declarative_macro_missing.ron
```diff
@@ -0,0 +1,46 @@
...
+                            begin_line @output
+                        }
+                    }
+                }
+            }
+            current @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
+                item {
+                    ... on Macro {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        public_api_eligible @filter(op: "=", value: ["$true"])
```
obi1kenobi, line 29: Our lint is for "macro removed" so we don't actually want to check `public_api_eligible` here. If we check that here, then we'd also trigger this lint if the macro became `#[doc(hidden)]`, which isn't the same lint anymore.
```suggestion
```

src/lints/declarative_macro_missing.ron
```diff
@@ -0,0 +1,46 @@
...
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "zero": 0,
+        "true": true,
+    },
+    error_message: "A publicly-visible `macro_rules` declarative macro cannot be imported by its prior name. A `#[macro_export]` may have been removed, or the macro itself may have been renamed or removed entirely.",
```
obi1kenobi, line 41: Excellent diagnostic, I forgot that removing `#[macro_export]` is a way to cause breakage here! 🎉

test_crates/declarative_macro_missing/old/src/lib.rs
```diff
@@ -0,0 +1,13 @@
...
+}
+
+#[macro_export]
+macro_rules! will_no_longer_be_exported {
+    () => {};
+}
+
+macro_rules! textual_scope_macro_removed {
+    () => {};
+}
```
obi1kenobi, line 13: A couple more test cases here.
```suggestion
}

#[macro_export]
macro_rules! became_doc_hidden {
    () => {};
}

#[doc(hidden)]
#[macro_export]
macro_rules! always_doc_hidden {
    () => {};
}

```

test_crates/declarative_macro_missing/new/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+macro_rules! will_no_longer_be_exported {
+    () => {};
+}
```
obi1kenobi, line 3: Adding some more test cases for `#[doc(hidden)]` on macro-rules macros.

Also, highly recommend configuring your editor to automatically add trailing newlines at the end of all files, they are necessary to keep shell tools like `wc` working correctly since they assume all lines end in `\n`.
```suggestion
}

#[doc(hidden)]
#[macro_export]
macro_rules! became_doc_hidden {
    () => {};
}

#[doc(hidden)]
#[macro_export]
macro_rules! always_doc_hidden {
    () => {};
}

```

test_outputs/query_execution/declarative_macro_missing.snap
```diff
@@ -0,0 +1,18 @@
...
+    {
+      "name": String("will_be_removed"),
+      "span_begin_line": Uint64(2),
+      "span_filename": String("src/lib.rs"),
+    },
+    {
+      "name": String("will_no_longer_be_exported"),
+      "span_begin_line": Uint64(7),
+      "span_filename": String("src/lib.rs"),
+    },
```
obi1kenobi, line 16:
```suggestion
    },
    {
      "name": String("became_doc_hidden"),
      "span_begin_line": Uint64(16),
      "span_filename": String("src/lib.rs"),
    },
```

test_outputs/witnesses/declarative_macro_missing.snap
```diff
@@ -0,0 +1,14 @@
...
+---
+[["./test_crates/declarative_macro_missing/"]]
+filename = 'src/lib.rs'
+begin_line = 2
+hint = 'will_be_removed!(...);'
+
+[["./test_crates/declarative_macro_missing/"]]
+filename = 'src/lib.rs'
+begin_line = 7
+hint = 'will_no_longer_be_exported!(...);'
```
obi1kenobi, line 14:
```suggestion
hint = 'will_no_longer_be_exported!(...);'

[["./test_crates/declarative_macro_missing/"]]
filename = 'src/lib.rs'
begin_line = 16
hint = 'became_doc_hidden!(...);'
```

obi1kenobi (commented):
Thanks for putting this together, and for the great ideas on how to test macros further!

Accepting with minor tweaks, which I'll apply since I feel bad for making you debug why the macro queries weren't working 😅

test_outputs/query_execution/declarative_macro_missing.snap
obi1kenobi: Oops, I messed up the lint definition myself — I shouldn't code while jetlagged 😅 This suggestion is backwards, and this line should be reported in the lint about `#[doc(hidden)]`, not here.

test_outputs/witnesses/declarative_macro_missing.snap
obi1kenobi: Same here.

Issue #937, opened 2024-09-19
Add more witness hints to currently-existing lints.


suaviloquence:
### Describe your use case

With the new unstable flag `cargo semver-checks -Z unstable-options --witness-hints`, `cargo-semver-checks` is now capable of printing 1-3 line **witness hints** that give a brief example of how downstream breakage could occur for each instance of a lint finding breakage of a SemVer guarantee.

For example, with the `function_missing` lint:

```
--- failure function_missing: pub fn removed or renamed ---

Description:
A publicly-visible function cannot be imported by its prior path. A `pub use` may have been removed, or the function itself may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.35.0/src/lints/function_missing.ron

Failed in:
  function function_missing::pub_use_removed_fn, previously in file /home/m/dev/cargo-semver-checks/test_crates/function_missing/old/src/lib.rs:4
note: downstream code similar to the following would break:
function_missing::pub_use_removed_fn(...);

  function function_missing::will_be_removed_fn, previously in file /home/m/dev/cargo-semver-checks/test_crates/function_missing/old/src/lib.rs:1
note: downstream code similar to the following would break:
function_missing::will_be_removed_fn(...);
```

### Describe the solution you'd like

For each of the [currently existing lints in `cargo-semver-checks`](https://github.com/obi1kenobi/cargo-semver-checks/tree/main/src/lints), we want to either add a witness hint or determine that we can't write one.

The process for adding a witness hint is in the [contributor docs](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md#adding-a-witness).  Briefly, once you've determined that you can write a witness hint, add a `witness` key to the `<lint_name>.ron` file:

```ron
SemverQuery(
     // --- rest of lint declaration --- 
    witness: (
         hint_template: "write the hint here as a {{handlebars}} template",
    ),
)
```

More details can be found in the CONTRIBUTING.md section linked above.

If it's not possible to write a witness hint, it would be great to explicitly set `witness: None` in the `<lint_name>.ron` file, accompanied by a comment explaining why it is not (currently) possible to write a hint.

### Additional Context

This is a new feature, and if any questions or bugs come up when adding a witness hint, please ask me @suaviloquence so we can help improve the docs and contributor experience for the future!

[...]

PR #977, opened 2024-10-18
feat: add witnesses for `function_const_removed`

CommanderStorm:
Related to https://github.com/obi1kenobi/cargo-semver-checks/issues/937

A better whitness would be the following, but I have no clue if this is possible and if yes, how.
I have tried a few things, but I am missing something. Maybe this would require a query.
```rs
add() // if len(arguments) == 0
add(...) // if len(arguments) >= 1
```

=> is the compromise okay of letting this be `add(...)` to save on adding a whitness-query just for this?

### DIFF
--- a/src/lints/function_const_removed.ron
+++ b/src/lints/function_const_removed.ron
@@ -48,4 +48,9 @@ SemverQuery(
     },
     error_message: "A publicly-visible function is no longer `const` and can no longer be used in a `const` context.",
     per_result_error_template: Some("function {{join \"::\" path}} in file {{span_filename}}:{{span_begin_line}}"),
+    witness: (
+        hint_template: r#"const fn foo() {
+    {{join "::" path}}(...);
+}"#,
+    ),
 )
--- a/src/lints/function_const_removed.ron
+++ b/src/lints/function_const_removed.ron
@@ -49,7 +49,7 @@ SemverQuery(
     error_message: "A publicly-visible function is no longer `const` and can no longer be used in a `const` context.",
     per_result_error_template: Some("function {{join \"::\" path}} in file {{span_filename}}:{{span_begin_line}}"),
     witness: (
-        hint_template: r#"const fn foo() {
+        hint_template: r#"const fn witness() {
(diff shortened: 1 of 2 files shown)


src/lints/function_const_removed.ron
```diff
@@ -48,4 +48,9 @@ SemverQuery(
     },
     error_message: "A publicly-visible function is no longer `const` and can no longer be used in a `const` context.",
     per_result_error_template: Some("function {{join \"::\" path}} in file {{span_filename}}:{{span_begin_line}}"),
+    witness: (
+        hint_template: r#"const fn foo() {
```
obi1kenobi, line 52: We're still setting up norms for how we do witnesses, but let's try to name witness functions and types things like `witness` rather than `foo` whenever possible? Names like `foo` can be mildly problematic in some circumstances, and in this case we have a better name available so let's use it.
```suggestion
        hint_template: r#"const fn witness() {
```

test_outputs/witnesses/function_const_removed.snap
```diff
@@ -0,0 +1,12 @@
+---
+source: src/query.rs
+description: "Lint `function_const_removed` did not have the expected witness output.\nSee https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md#testing-witnesses\nfor more information."
+expression: "&actual_witnesses"
+---
+[["./test_crates/function_const_removed/"]]
+filename = 'src/lib.rs'
+begin_line = 1
+hint = '''
+const fn foo() {
```
obi1kenobi, line 10:
```suggestion
const fn witness() {
```

obi1kenobi (approved):
Very nice, thank you!

For the witness _hint_, it's totally fine to omit the function arguments and just use `...`. We're just trying to give the user a broad-strokes idea of what might go wrong. Whether function args are present or not isn't that important in this situation so `...` is fine.

Down the line, we'll have stricter rules for actual compilable witnesses as opposed to hints. But that's something for us to tackle down the line.

PR #978, opened 2024-10-18
chore: add a witness for `enum_variant_missing`

CommanderStorm:
Similar to https://github.com/obi1kenobi/cargo-semver-checks/pull/977 (but a bit more wrong syntax gets produced)

Similar to , this is not quite correct syntax:
a query for `variant` would be nessesary to have the correct brackets to construct the enum (struct, vs plain vs tuple).

### DIFF
--- a/src/lints/enum_variant_missing.ron
+++ b/src/lints/enum_variant_missing.ron
@@ -19,6 +19,8 @@ SemverQuery(
                         }
 
                         variant {
+                            kind: __typename @output
+
                             variant_name: name @output @tag
                             public_api_eligible @filter(op: "=", value: ["$true"])
 
@@ -57,6 +59,6 @@ SemverQuery(
     error_message: "A publicly-visible enum has at least one variant that is no longer available under its prior name. It may have been renamed or removed entirely.",
     per_result_error_template: Some("variant {{enum_name}}::{{variant_name}}, previously in file {{span_filename}}:{{span_begin_line}}"),
     witness: (
-        hint_template: r#"let witness = {{join "::" path}}::{{variant_name}};"#,
+        hint_template: r#"let witness = {{join "::" path}}::{{variant_name}}{{#if (eq kind "StructVariant")}}{...}{{/if}}{{#if (eq kind "TupleVariant")}}(...){{/if}};"#,
     ),
 )
--- a/src/lints/enum_variant_missing.ron
+++ b/src/lints/enum_variant_missing.ron
@@ -59,6 +59,6 @@ SemverQuery(
     error_message: "A publicly-visible enum has at least one variant that is no longer available under its prior name. It may have been renamed or removed entirely.",
     per_result_error_template: Some("variant {{enum_name}}::{{variant_name}}, previously in file {{span_filename}}:{{span_begin_line}}"),
     witness: (
-        hint_template: r#"let witness = {{join "::" path}}::{{variant_name}}{{#if (eq kind "StructVariant")}}{...}{{/if}}{{#if (eq kind "TupleVariant")}}(...){{/if}};"#,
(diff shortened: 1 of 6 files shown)


obi1kenobi:
In this case, I think it's worth getting the variant kind right.

We can tweak the lint query by adding `kind: __typename @output` on the original variant which will say if it's plain / tuple / struct, and then use either no extra notation, or `(...)`, or `{ ... }` accordingly. The templating language should be powerful enough to support such an addition.

Also, it seems like the PR title got cut off maybe?

CommanderStorm:
Shure. Please review, if I picked the correct query (=you did not mean `witness.witness_query`)

(I have also added a testcase to prove that it works)

src/lints/enum_variant_missing.ron
```diff
@@ -56,4 +58,7 @@ SemverQuery(
     },
     error_message: "A publicly-visible enum has at least one variant that is no longer available under its prior name. It may have been renamed or removed entirely.",
     per_result_error_template: Some("variant {{enum_name}}::{{variant_name}}, previously in file {{span_filename}}:{{span_begin_line}}"),
+    witness: (
+        hint_template: r#"let witness = {{join "::" path}}::{{variant_name}}{{#if (eq kind "StructVariant")}}{...}{{/if}}{{#if (eq kind "TupleVariant")}}(...){{/if}};"#,
```
obi1kenobi, line 62: To the best extent possible, let's try to output code that is analogous to how `cargo fmt` would represent it, so that we aid readability. Tuple variants get constructed as `Enum::Tuple(a, b)` but struct variants get constructed as `Enum::Struct { a, b }` — note the space before the curly brace.
```suggestion
        hint_template: r#"let witness = {{join "::" path}}::{{variant_name}}{{#if (eq kind "StructVariant")}} {...}{{/if}}{{#if (eq kind "TupleVariant")}}(...){{/if}};"#,
```

test_crates/enum_variant_missing/old/src/lib.rs
```diff
@@ -1,10 +1,21 @@
...
+pub enum TupleVariantWillBeRemoved {
+    Foo(usize),
+    /// Testing: <https://doc.rust-lang.org/cargo/reference/semver.html#item-remove>
+    Bar(usize),
+}
+
+pub enum StructVariantWillBeRemoved {
+    Foo { bar: usize },
+    /// Testing: <https://doc.rust-lang.org/cargo/reference/semver.html#item-remove>
+    Bar{ bar: usize },
```
obi1kenobi, line 16: Please make sure to `cargo fmt` test crate code.
```suggestion
    Bar { bar: usize },
```

test_outputs/witnesses/enum_variant_missing.snap
```diff
@@ -0,0 +1,24 @@
...
+
+[["./test_crates/enum_variant_missing/"]]
+filename = 'src/lib.rs'
+begin_line = 10
+hint = 'let witness = enum_variant_missing::TupleVariantWillBeRemoved::Bar(...);'
+
+[["./test_crates/enum_variant_missing/"]]
+filename = 'src/lib.rs'
+begin_line = 16
+hint = 'let witness = enum_variant_missing::StructVariantWillBeRemoved::Bar{...};'
```
obi1kenobi, line 24:
```suggestion
hint = 'let witness = enum_variant_missing::StructVariantWillBeRemoved::Bar {...};'
```

obi1kenobi (commented):
Good stuff. Still a few nitpicks that I'd like you to try to catch on your own in future PRs, either up front or in self-review, though.

obi1kenobi (approved):

obi1kenobi:
I'm happy to merge this once CI is happy with it again.

obi1kenobi (approved):

PR #1014, opened 2024-12-03
Update to latest major versions of dependencies.

obi1kenobi:


obi1kenobi:
Seems like this and #1003 both have to wait for us to drop Rust 1.80 support before they can merge.

PR #1016, opened 2024-12-05
Preserve `RUSTDOCFLAGS` so that custom `--cfg` settings can be applied.

obi1kenobi:
Allow users to set `--cfg` options included when scanning a crate by invoking the tool as:

```
RUSTDOCFLAGS="--cfg your-cfg-value" cargo semver-checks
```


obi1kenobi:
Thanks to @wmmc88 for suggesting it in #1012!

PR #1025, opened 2024-12-07
Remove more kinds of `cargo` output related to blocking on file locks.

obi1kenobi:
Encountered a new `cargo` output line while streaming with @orhun:
>     Blocking waiting for file lock on build directory

Let's eliminate it from our snapshots, since it isn't relevant to what we're testing.


orhun:
LGTM!

PR #1035, opened 2024-12-10
Ensure metadata of self-referential crates can be read.

obi1kenobi:
Also ensure that if reading metadata fails, we don't crash and instead just avoid checking manifest-related lints.


obi1kenobi:
Thanks to @jonathanpallant for reporting this issue!

PR #1048, opened 2024-12-16
Add custom registry support

fujiapple852:
@obi1kenobi this PR is a rebase (and rework) of @drewkett's earlier PR https://github.com/obi1kenobi/cargo-semver-checks/pull/945, to add support for custom cargo registries, on top of the latest `main`.  I say rework as `main` has changed quite a bit since the original PR and so a simple rebase was not enough.

This PR also adds an integration test for the custom registry functionality.  The test use a static file `mango` registry which is served via an ephemeral `axus` instance. The test is self contained and does not require any external resources.

Finally, an unrelated bug I observed, with the regex used for the `bugreport` test, is resolved. This could be pulled out into a separate PR if needed.

The goal here is to get this into a suitable state such that it could be merged to `main` to thus avoid the need to maintain a downstream fork.  I have kept my changes separate from @drewkett's original changes to aid in reviewing the delta, but the intention is to eventually squash these.

Thanks!


src/data_generation/request.rs
```diff
@@ -40,12 +42,22 @@ impl RequestKind<'_> {
...
             }
         })
     }
+
+    pub(super) fn index_url(&self) -> anyhow::Result<&str> {
+        Ok(match self {
+            Self::Registry(RegistryRequest { index_url, .. }) => index_url.as_str(),
+            Self::LocalProject(ProjectRequest { manifest: _ }) => {
+                // Unclear what we should do here, panic for now...
+                todo!()
```
fujiapple852, line 57: @drewkett @obi1kenobi I wasn't sure what to do here, thoughts?

obi1kenobi:
Sorry for missing this. Fundamentally, my concern about maintainability and funding is blocking all third-party registry support progress. Right now, two things are true:
- AFAICT it's _companies_ that need third-party registry support.
- Only a couple of companies are funding `cargo-semver-checks` with even $100/month, and AFAICT none of them need third-party registry support.

I'm already at my limit of what I can maintain with the present circumstances and funding level, and third-party registry support risks massively increasing the surface area _without_ anywhere close to a corresponding increase in funding or corporate support.

I'd love to work with customers who need third-party registry support on a plan for sustainable project funding that would allow this functionality to merge and be supported. In the meantime, the status quo level of project funding does not allow it.

Issue #1068, opened 2025-01-05
error: output of --print=file-names missing when learning about target-specific information from rustc when `--target` is specified


zjp-CN:
### Steps to reproduce the bug with the above code

Thanks for this great tool!


I see a bunch of these repos whos semver-checks results on **non-** x86_64-unknown-linux-gnu targets report as the title shows.
One of the repos is https://github.com/arceos-org/arm_gicv2 . (I can list more if needed.)

Reproducible commands:

```sh
# latest hash: https://github.com/arceos-org/arm_gicv2/commit/cf756f7623b6b6af2f6fe423d1da19ebe96e7e1d
git clone https://github.com/arceos-org/arm_gicv2.git 
cd arm_gicv2
rustup target add aarch64-unknown-none-softfloat
cargo semver-checks --target aarch64-unknown-none-softfloat
```

### Actual Behaviour

I got compilation error from rustc.

```rust
error: output of --print=file-names missing when learning about target-specific information from rustc
command was: `/root/.cargo/bin/sccache /root/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/rustc - --crate-name ___ --print=file-names --cap-lints=allow --target aarch64-unknown-none-softfloat --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=split-debuginfo --print=crate-name --print=cfg`

--- stdout
___
lib___.rlib
lib___.a
/root/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu
off
___
debug_assertions
fmt_debug="full"
overflow_checks
panic="abort"
proc_macro
relocation_model="static"
target_abi="softfloat"
target_arch="aarch64"
target_endian="little"
target_env=""
target_has_atomic
target_has_atomic="128"
target_has_atomic="16"
target_has_atomic="32"
target_has_atomic="64"
target_has_atomic="8"
target_has_atomic="ptr"
target_has_atomic_equal_alignment="128"
target_has_atomic_equal_alignment="16"
target_has_atomic_equal_alignment="32"
target_has_atomic_equal_alignment="64"
target_has_atomic_equal_alignment="8"
target_has_atomic_equal_alignment="ptr"
target_has_atomic_load_store
target_has_atomic_load_store="128"
target_has_atomic_load_store="16"
target_has_atomic_load_store="32"
target_has_atomic_load_store="64"
[...]

PR #1071, opened 2025-01-08
fix(--target): RUSTDOCFLAGS=--cap-lints=warn

zjp-CN:
close https://github.com/obi1kenobi/cargo-semver-checks/issues/1068

Cargo relies on warning diagnostics to find information, --cap-lints=allow eliminates the warnings, so set the value to warn level to keep it.


obi1kenobi:
I'm pretty worried about the possibility of something along the lines of `-Dwarnings` (either in code as a `#[deny]`, or in the CLI, or in some config file) that causes those generated warnings to turn into errors. We added `--cap-lints` precisely because of `-Dwarnings` style issues.

I'd be open to merging such a change if you could include tests and test crates that check as many of those cases as possible, so that we know that `--cap-lints=warn` is sufficient and won't just cause something else to explode. I'd be happy to point you in the right direction for writing those test cases, I just want to first check if you're open to doing that :)

zjp-CN:
Thanks for reminding me of -Dwarnings...

The fix here is unnecessary if https://github.com/rust-lang/cargo/pull/15036 lands on nightly in a few days. I'll close this PR when it works. 

I can submit a testcast that needs --target argument in another PR. But there seems to [have one already](https://github.com/obi1kenobi/cargo-semver-checks/blob/7273ccdfede459e5d402309c19148cae0fb4ff32/tests/specified_target.rs)?

zjp-CN:
I'm closing this since this should be solved on cargo side.

PR #1073, opened 2025-01-10
Ajust and add target testcases

zjp-CN:
Context: https://github.com/obi1kenobi/cargo-semver-checks/pull/1071#issuecomment-2581912310

These conditinal compiled tests requring target can be tested if `rustup target add riscv64gc-unknown-linux-gnu` is set up. I've tested them locally, and they will pass with riscv64gc-unknown-linux-gnu target installed. 

But I'm not sure if adding [target field in ci.yml](https://github.com/obi1kenobi/cargo-semver-checks/blob/7273ccdfede459e5d402309c19148cae0fb4ff32/.github/workflows/ci.yml#L186) is enough. (Haven't added it yet) 

It's weird riscv64gc-unknown-linux-gnu target can be recognized by cargo, but aarch64_unknown_none not, which is the problem I ran into.  `with_{env_var,flag}_aarch64_unknown_none` testcases always fail on `["1.81", "1.82", "1.83", "stable", "beta"]` toolchains and [the nightly one before this weekend](https://github.com/obi1kenobi/cargo-semver-checks/issues/1068#issuecomment-2581704041). So it'd be a bad idea to incorporate them for now and near future.

Feel free to close this PR if you do not accept it.

### DIFF
--- a/tests/specified_target.rs
+++ b/tests/specified_target.rs
@@ -34,6 +34,14 @@ fn with_env_var_riscv64gc_unknown_linux_gnu() {
 }
 
 #[test]
+fn with_env_var_aarch64_unknown_none() {
+    base()
+        .env("CARGO_BUILD_TARGET", "aarch64-unknown-none")
+        .assert()
+        .success();
+}
+
+#[test]
 #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
 fn with_flag() {
     base()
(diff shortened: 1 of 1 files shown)


obi1kenobi:
I feel these might be better as separate test jobs in `ci.yml` while keeping the conditional compilation attributes, than as tests that run on every `cargo test` including locally.

We already have a relatively complex setup for getting the repo bootstrapped to the point that tests pass. I don't think it's worth the maintenance burden of requiring contributors to do `rustup target add riscv64gc-unknown-linux-gnu` before e.g. contributing a new lint. This is especially the case if the targets have any potential of having compatibility issues on Windows or macOS.

PR #1090, opened 2025-01-29
feat: static became unsafe lint

Frank-III:
Hi @obi1kenobi, I could not get the test crate to work while the query should be correct? The reason I suppose is the doc seems not getting generated correctly. I could get the doc correctly with `cargo rustdoc -Zunstable-options --output-format=json` in a test crate with the same code .

Could you guide me on how can I get it to work or there is something wrong with my setup


obi1kenobi:
Interesting, what's the error that you're seeing when generating the rustdoc locally with the script? It seems that the rustdoc build is succeeding on CI, so it might be some kind of local setup issue.

src/lints/static_unsafe_added.ron
```diff
@@ -0,0 +1,51 @@
+SemverQuery(
+    id: "static_unsafe_added",
```
obi1kenobi, line 2: I appreciate you naming this while keeping in mind consistency with existing lints. Our lint naming isn't particularly consistent in general, and in this case I think we may be better off breaking consistency to avoid confusion. To trigger this lint, it's not the case that someone added `unsafe` but that they replaced `safe` with `unsafe`.

So perhaps a name like `static_became_unsafe` would be best? If you agree, make sure to rename both the `id` and the file name, as well as the test crate, snapshot file, and `query.rs` entry.

In the future, I imagine we might rename other lints to be consistent with this one, instead of vice versa.

src/lints/static_unsafe_added.ron
```diff
@@ -0,0 +1,51 @@
...
+    human_readable_name: "pub static became unsafe",
+    description: "a extern static became unsafe to call.",
```
obi1kenobi, line 4: "Call" doesn't seem like the right verb for statics, since they aren't callable. Let's go with "use" instead:
```suggestion
    description: "An extern static became unsafe to use.",
```

src/lints/static_unsafe_added.ron
```diff
@@ -0,0 +1,51 @@
...
+    required_update: Major,  
+    lint_level: Deny,  
+    reference_link: Some("https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html#accessing-or-modifying-a-mutable-static-variable"),
```
obi1kenobi, line 7: I'm a bit confused by this reference link. It seems to be about _mutable_ statics, whereas this lint doesn't check mutability at all.

Is there a more appropriate reference document we can link to?

src/lints/static_unsafe_added.ron
```diff
@@ -0,0 +1,51 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+    },
+    error_message: "A publicly-visible static became `unsafe`, so calling it now requires an `unsafe` block.",
```
obi1kenobi, line 49: Same "call" -> "use" replacement as above:
```suggestion
    error_message: "A publicly-visible static became `unsafe`, so using it now requires an `unsafe` block.",
```

obi1kenobi (commented):
The query looks good! A couple of small suggestions for improvements around the metadata.

Were you able to resolve the rustdoc building issue you were facing?

obi1kenobi:
Also unfortunately, static safety information is only available in Rust 1.85-beta and newer, so the snapshot test will fail on older Rust versions =/

We'll have to either figure out a way to gate the snapshot test based on the Rust version, or delay merging this lint until 1.85 is the _oldest_ Rust supported by `cargo-semver-checks`. I haven't thought this through, but there are a few other tests elsewhere that are version-specific so perhaps we can reuse the same idea here. If you're open to looking into it, look for `rustc_version::version()` invocations to find them.

src/lints/static_unsafe_added.ron
Frank-III: do you think this is a good one: https://doc.rust-lang.org/reference/items/static-items.html

Frank-III:
> Also unfortunately, static safety information is only available in Rust 1.85-beta and newer, so the snapshot test will fail on older Rust versions =/
> 
> We'll have to either figure out a way to gate the snapshot test based on the Rust version, or delay merging this lint until 1.85 is the _oldest_ Rust supported by `cargo-semver-checks`. I haven't thought this through, but there are a few other tests elsewhere that are version-specific so perhaps we can reuse the same idea here. If you're open to looking into it, look for `rustc_version::version()` invocations to find them.

Yeah I thinks so. I fixed the snap but it just not work with old version of rust. I would take a look into that!

obi1kenobi:
The lint looks good! If we can get the snapshot unblocked on Rust 1.83 and stable, this is good to merge!

Frank-III:
> Also unfortunately, static safety information is only available in Rust 1.85-beta and newer, so the snapshot test will fail on older Rust versions =/
> 
> We'll have to either figure out a way to gate the snapshot test based on the Rust version, or delay merging this lint until 1.85 is the _oldest_ Rust supported by `cargo-semver-checks`. I haven't thought this through, but there are a few other tests elsewhere that are version-specific so perhaps we can reuse the same idea here. If you're open to looking into it, look for `rustc_version::version()` invocations to find them.

could you give more info about how we could get this lint to be unblocked for Rust 1.83 using `rustc_version::version()`, it seems that all snapshots inside the `query_execution` folder can not control the version of Rust to run the query, or should I have a separate test for it?

obi1kenobi:
For example, perhaps by updating `check_query_execution()` by adding a block around line 753 with something like:
```rust
if lint_name == "static_became_unsafe" && rustc_version::version().map_or(false, |version| version < Version::new(1, 85, 0)) {
    eprintln!("skipping query execution test for lint `static_became_unsafe` since data for it isn't available in Rust prior to 1.85");
    return;
}
```

Frank-III:
Thank you for your detailed guidance, I will keep up with the codebase more!

obi1kenobi:
Thank you for the perseverance through all the tricky issues you've run into! I'm always happy to mentor folks who care about writing thoughtful and correct code.

Issue #57, opened 2022-08-09
Added deprecated attribute API lint, minor incompatibility level


epage:
Semver says deprecations are minor changes.

Reference: https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

Will require separate checks for:
- [x] struct, enum
- [x] union
- [x] trait: #1092
- [x] fn: #1093 
- [x] const / static (`GlobalValue`): #1101 
- [x] trait methods: #1098
- [x] `ImplOwner` (struct/enum/union) methods: #1130
- [x] `ImplOwner` associated constants: #1131
- [x] trait associated types: #1099
- [x] trait associated constants: #1100
- [x] enum variants: #1179 
- [x] fields within enum struct variants: #1180 
- [x] fields within enum tuple variants: #1202
- [x] struct fields: #1125
- [x] declarative macros: #1102
- [x] proc macros: #1118

pksunkara:
@obi1kenobi It hasn't been implemented for `union`. You might want to separate it out from `struct & enum` in the issue description.

obi1kenobi:
Done, thanks!

Frank-III:
`struct fields` should be marked done: #1125. the only two missing parts have PR opened.

obi1kenobi:
Sweet! Thank you for all your work, that's a lot of boxes you've checked!

Might you be interested in doing a bit of schema extension next, to unlock even more lints elsewhere? If so, adding "field ordering" information to struct fields and enum tuple/struct variant fields can unlock ~10ish more new lints. This is because changing the order of struct fields (e.g. swapping which field is first and which is second) can be a major breaking change in some circumstances.

Frank-III:
> Sweet! Thank you for all your work, that's a lot of boxes you've checked!

Thank you for your guidance, the more lint I added the more I realized how important c-s-c would be in the rust ecosystem!

> Might you be interested in doing a bit of schema extension next, to unlock even more lints elsewhere? If so, adding "field ordering" information to struct fields and enum tuple/struct variant fields can unlock ~10ish more new lints. This is because changing the order of struct fields (e.g. swapping which field is first and which is second) can be a major breaking change in some circumstances.

Definitely yes. Could you explain more about how it would be implemented? like adding a field to the `StructField` / `Variant` type ? and for the lints, we need to check if say a struct's field's order is changed, and report something like:
- "field `a` in `struct` order has changed"
[...]

PR #1092, opened 2025-01-31
new lint: `trait_marked_deprecated`

Frank-III:
part of: #57


obi1kenobi:
This looks great! I'm away from my computer for the weekend so I can't
merge it yet, but assume it's merged 😅

Feel free to hammer out more of the lints on that list if you'd like. Also
for that snapshot test where you had to change a 7 to an 8, consider
updating the rules there to replace the number with a placeholder, just
like the `[PASS]` next to it. We must have forgotten to do that originally.

obi1kenobi (approved):

PR #1095, opened 2025-02-02
new lint:  added `enum_variant_deprecated`

Gmin2:
part of #57

PR #1096, opened 2025-02-02
feat: added new lint `enum_variant_field_deprecated`

Gmin2:
part of #57

### DIFF
--- a/src/lints/enum_variant_field_deprecated.ron
+++ b/src/lints/enum_variant_field_deprecated.ron
@@ -1,0 +1,66 @@
+SemverQuery(
+    id: "enum_variant_field_deprecated",
+    human_readable_name: "enum variant field marked with #[deprecated]",
+    description: "A field in an enum variant was marked with the #[deprecated] attribute.",
+    required_update: Minor,
+    lint_level: Deny,
+    reference_link: Some("https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute"),
+    query: r#"
+    {
+        CrateDiff {
+            current {
+                item {
+                    ... on Enum {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        enum_name: name @output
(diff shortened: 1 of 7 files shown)


src/lints/enum_variant_field_deprecated.ron
```diff
@@ -0,0 +1,66 @@
...
+                item {
+                    ... on Enum {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        enum_name: name @output
+                        importable_path {
+                            path @output @tag
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+                        variant {
+                            ... on StructVariant {
```
obi1kenobi, line 21: Tuple variants can also have deprecated fields:
```rust
pub enum Example {
    Variant(#[deprecated] i64),
}
```

We'll need to add separate lints for struct and tuple variants, so let's name this lint `enum_struct_variant_field_deprecated` and the tuple one `enum_tuple_variant_field_deprecated`.

src/lints/enum_variant_field_deprecated.ron
```diff
@@ -0,0 +1,66 @@
...
+                                variant_name: name @output
+                                field {
+                                    field_name: name @output @tag
```
obi1kenobi, line 24: Both variants and their fields can be `#[doc(hidden)]`, so we'll need `public_api_eligible` checks in both the variant and the field blocks, in both baseline and current. Test cases for all these situations would be good too.

obi1kenobi (commented):
This PR is also broadly in the right direction, though it seems to have some of the same issues as #1095. I didn't point out those duplicate issues since I figured you can probably spot them already, but feel free to ask questions if you aren't sure.

obi1kenobi:
Going to close this for inactivity for now, just to un-claim the issue it's attached to. If you get a chance to come back to it and complete it, feel free to reopen the PR. Hope all is well!

PR #1098, opened 2025-02-03
new lint: `trait_method_marked_deprecated`

Frank-III:
part of #57

### DIFF
--- a/src/lints/trait_method_marked_deprecated.ron
+++ b/src/lints/trait_method_marked_deprecated.ron
@@ -21,6 +21,7 @@ SemverQuery(
 
                         method {
                             method_name: name @output @tag
+                            public_api_eligible @filter(op: "=", value: ["$true"])
                             deprecated @filter(op: "=", value: ["$true"])
                         }
 
@@ -43,6 +44,7 @@ SemverQuery(
 
                         method {
                             name @filter(op: "=", value: ["%method_name"])
+                            public_api_eligible @filter(op: "=", value: ["$true"])
                             deprecated @filter(op: "!=", value: ["$true"])
                         }
                     }
--- a/src/lints/trait_method_marked_deprecated.ron
+++ b/src/lints/trait_method_marked_deprecated.ron
@@ -1,0 +1,59 @@
+SemverQuery(
+    id: "trait_method_marked_deprecated",
+    human_readable_name: "trait method #[deprecated] added",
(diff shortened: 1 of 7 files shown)


src/lints/trait_method_marked_deprecated.ron
```diff
@@ -0,0 +1,59 @@
...
+                        name @output
+
+                        importable_path {
+                            path @tag @output
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        method {
+                            method_name: name @output @tag
+                            deprecated @filter(op: "=", value: ["$true"])
```
obi1kenobi, line 24: This needs a `public_api_eligible` check both here and on the baseline side.

Otherwise, this lint will flag trait methods that were `#[doc(hidden)]` in the baseline, which is incorrect since they weren't public API.

Relevant docs here:
https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/rustdoc-v39/src/rustdoc_schema.graphql#L91-L110

obi1kenobi: Consider adding test cases for `#[doc(hidden)]` on the trait and on the trait method, to make sure that non-public-API items don't get reported by this lint.

Frank-III: Thank you! was looking into the sealed thing, and got stuck, why it is so complex 😭. I would continue and finish `trait associated types` and `trait associated constants` then

obi1kenobi: > why it is so complex

The best answer is that it's a system that grew organically, and was not designed intentionally as a whole. There's no concept of a "sealed" or "public API sealed" trait in rustc or the Rust language proper — those things are ones that someone merely noticed are possible within the current language boundaries, and they've proven useful to the community so they remain in use.

If they were designed intentionally, we'd have e.g. a `#[sealed]` attribute on traits instead. This may yet happen! The process for `#[non_exhaustive]` was similar: there was a common workaround for it that was broadly used in the community, and eventually enough momentum built for a language-level fix that became the `#[non_exhaustive]` attribute.

I'm reasonably confident that sooner or later, we'll have _something_ along these lines for trait sealing as well. But first, someone has to analyze the present situation, figure out all the edge cases, and write it up in a form that gets the community interested. In this case, that someone was me :)

obi1kenobi (approved):

PR #1101, opened 2025-02-04
new lint: `global_value_marked_deprecated`

Frank-III:
part of #57 

I am actually not very sure when I should use `public_api_eligible` or `visibility_limit` since we check that for trait method/type/const and enum, enum variant, but not for trait itself, impl_owner itself?


obi1kenobi:
> I am actually not very sure when I should use `public_api_eligible` or `visibility_limit` since we check that for trait method/type/const and enum, enum variant, but not for trait itself, impl_owner itself?

This is a bit of nuance, yes:
- `visibility_limit` is "how is the item declared". A `pub trait` is public, and without `pub` it isn't, obviously.
- `public_api_eligible` is "is there anything on this item directly that would prevent it from being public API?" It's a synthesis of `visibility_limit`, `doc_hidden`, and `deprecated`: non-`pub` items aren't public API eligible, and hidden items aren't either _unless_ they are also deprecated.
- `public_api` along the `importable_path` edge is "is this specific import path public API."

Using `visibility_limit` to determine public API isn't sufficient by itself. For example, this item has public visibility but isn't public API:
```rust
#[doc(hidden)]
pub struct Example;
```

Using `public_api_eligible` by itself isn't sufficient either. For example, here's an instance where an item is _in principle, by itself_ able to be public API, but still isn't public API because of surrounding reasons:
```rust
mod private {
    pub trait Sealed {}
}
```
You'll recognize this as a trick common in trait-sealing, where `Sealed` would be applied as a super-trait to traits we wish to seal. `Sealed` is `public_api_eligible` because it's a `pub` item that isn't hidden. But it isn't public API, because there's no publicly-visible path by which a downstream crate can import it — it _cannot be used_.

So, concrete advice for queries:
- For an item to be public API, `visibility_limit` must be public or "default" (like in enum variants, which don't have their own visibility). Including this check is merely an optimization, to quickly prune away results that won't match. Not including it won't make the lint incorrect, since `public_api` and `public_api_eligible` already take into account `visibility_limit` internally.
- If the item is importable (its schema shows an `importable_path` edge), whether it's public API is governed by the `public_api` property through the `importable_path` edge. Ignore the `public_api_eligible` property on such items.
- For _associated_ items (struct fields, enum variants, trait associated types, `ImplOwner` methods, etc.), that item is public API if it's `public_api_eligible` and _also_ the item it's associated with is public API, per the above.
- Enum variants are special, since they are the only associated item in Rust that can be imported directly. Quite frankly, I haven't worked out all the edge cases here, and `cargo-semver-checks` doesn't handle breakage related to directly importing variants. This is not a very common concern, but if it happens to be something you're interested in tackling, I'd be happy to point you in the right direction! Also happy to suggest other more impactful things as well!

test_crates/global_value_marked_deprecated/new/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+#[deprecated]
+pub const VALUE_TO_DEPRECATED: i32 = 42;
+
+#[deprecated]
+pub static STATIC_TO_DEPRECATED: &str = "hello";
+
+#[deprecated]
+pub const fn CONST_FN_TO_DEPRECATED(x: i32) -> i32 {
+    x + 1
+}
```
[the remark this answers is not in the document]
Frank-III: Yes I knew when I ran the test, I was worried about if it would trigger both! Should I keep it?

Frank-III:
Thank you for the detailed explanation, learnt a lot again!

> * For an item to be public API, `visibility_limit` must be public or "default" (like in enum variants, which don't have their own visibility). Including this check is merely an optimization, to quickly prune away results that won't match. Not including it won't make the lint incorrect, since `public_api` and `public_api_eligible` already take into account `visibility_limit` internally.
> * If the item is importable (its schema shows an `importable_path` edge), whether it's public API is governed by the `public_api` property through the `importable_path` edge. Ignore the `public_api_eligible` property on such items.
> * For _associated_ items (struct fields, enum variants, trait associated types, `ImplOwner` methods, etc.), that item is public API if it's `public_api_eligible` and _also_ the item it's associated with is public API, per the above.

I see, very helpful advice on when to use which! I would apply this rule when I add new lints!

> Enum variants are special, since they are the only associated item in Rust that can be imported directly. Quite frankly, I haven't worked out all the edge cases here, and cargo-semver-checks doesn't handle breakage related to directly importing variants. This is not a very common concern, but if it happens to be something you're interested in tackling, I'd be happy to point you in the right direction! Also happy to suggest other more impactful things as well!

Yes, I would definitely want to work on deeper things after getting familiar with the whole project more!

test_outputs/query_execution/pub_module_level_const_missing.snap
```diff
@@ -3,6 +3,18 @@ source: src/query.rs
...
+      "name": String("CONST_BECOMING_STATIC"),
+      "path": List([
+        String("global_value_marked_deprecated"),
+        String("CONST_BECOMING_STATIC"),
+      ]),
+      "span_begin_line": Uint64(9),
+      "span_filename": String("src/lib.rs"),
+      "visibility_limit": String("public"),
+    },
+  ],
```
obi1kenobi, line 17: This is actually a false-positive in the `pub_module_level_const_missing` lint, as of Rust 1.83 which relaxed the rules on when & how `static` items may be used in `const` items. Previously, assigning a `static` to a `const` item was always a hard error. As of 1.83, certain `static` items can be assigned to `const` items so we can't prove breakage _merely_ by saying "this item used to be `const` but now is `static`."

Care to fix that lint? All you need to do is flip the `baseline` type check from `... on Constant` to `... on GlobalValue`

If you're also interested in writing new lints that flag the new breakage behavior made possible by Rust 1.83, I'd be happy to point you in the right direction there as well!

Frank-III: > Care to fix that lint? All you need to do is flip the baseline type check from ... on Constant to ... on GlobalValue

Definitely

> If you're also interested in writing new lints that flag the new breakage behavior made possible by Rust 1.83, I'd be happy to point you in the right direction there as well!

Would like to know more about this, would you write it down in a new issue or?

Frank-III: >Care to fix that lint? All you need to do is flip the baseline type check from ... on Constant to ... on GlobalValue

I think we want to flip `current` to be `GlobalValue` since now a `const` became `static` should be allowed?

src/lints/pub_module_level_const_missing.ron
```diff
@@ -1,7 +1,7 @@
 SemverQuery(
     id: "pub_module_level_const_missing",
     human_readable_name: "pub module-level const is missing",
-    description: "A pub const is missing, renamed, or changed to static.",
+    description: "A pub const is missing or renamed.",
```
obi1kenobi, line 4: Nice catch!

obi1kenobi (approved):

test_outputs/query_execution/pub_module_level_const_missing.snap
obi1kenobi: Yes, of course. Thanks for the catch!

PR #1125, opened 2025-02-05
new lint: `struct_field_marked_deprecated` 

Frank-III:
part of: #57 

I think I got the query right, but it seems to get triggered when the struct is marked deprecated which should not be true?


src/lints/struct_field_marked_deprecated.ron
```diff
@@ -0,0 +1,66 @@
...
+                    ... on Struct {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        struct_name: name @output
+
+                        # It's possible that both the struct and its field became deprecated.
+                        # This is a rare case where we don't consider the lint on the field
+                        # to be a duplicate lint of the one on the struct.
+                        #
+                        # Deprecation on the struct means we shouldn't use it in new code,
+                        # but existing code can still access its fields until they are removed.
```
obi1kenobi, line 40: Consider adding some test cases that involve `#[doc(hidden)]` on both the struct and its fields, to pin down the exact behavior of this lint.

This is especially important when leaving a comment describing some non-obvious decision made in the lint, like this one here.

test_crates/struct_field_marked_deprecated/new/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+pub struct Plain {
+    #[deprecated]
+    pub field_becomes_deprecated: i64,
+    #[deprecated = "This field will be removed in the next major version"]
+    pub field_with_deprecation_message: String,
+}
+
+pub struct Tuple(
+    #[deprecated] pub i64,
+    #[deprecated = "Use new_tuple_field instead"] pub String,
```
obi1kenobi, line 11: Could you make sure to run `cargo fmt` on the test crates to standardize their formatting and make them easier to read? I think this would get reformatted to put the attributes on a separate line.

test_outputs/query_execution/type_marked_deprecated.snap
```diff
@@ -3,6 +3,30 @@ source: src/query.rs
...
+      "owner_type": String("Struct"),
+      "path": List([
+        String("struct_field_marked_deprecated"),
+        String("StructBecomesDeprecated"),
+      ]),
+      "span_begin_line": Uint64(47),
+      "span_filename": String("src/lib.rs"),
+      "visibility_limit": String("public"),
+    },
+  ],
```
obi1kenobi, line 29: These results make sense — the `struct_field_marked_deprecated` test crate added two structs that get deprecated, and the `type_marked_deprecated` lint caught them. We run every lint on every test crate during testing, so this is expected.

What I'm confused by is why there's no new snapshot file for the new lint? Is it not catching anything?

Frank-III: >These results make sense — the struct_field_marked_deprecated test crate added two structs that get deprecated, and the type_marked_deprecated lint caught them. We run every lint on every test crate during testing, so this is expected.

yep

>What I'm confused by is why there's no new snapshot file for the new lint? Is it not catching anything?


Yes, I rejected the new snap file since it is incorrect; I don't want it to pass the test. The result would be in the test failing logs

test_crates/struct_field_marked_deprecated/new/src/lib.rs
Frank-III: I think I ran the formatter, let me try it again

test_outputs/query_execution/type_marked_deprecated.snap
obi1kenobi: Makes sense! Longer reply (with edits) outside the thread 👇

Frank-III:
>Is it triggering on only the struct's deprecation right now? If so, the best thing to check would be to look directly inside the generated rustdoc JSON for that test crate in your test data directory, and look up the struct / its fields and check the deprecation status field. I suppose it's possible (though I'd find it surprising) that a deprecation of the struct may cause rustdoc to automatically emit deprecations on the fields too. If that's the case, we can open an issue and discuss it with the rustdoc team to figure out if that's intentional for some good reason, or just accidental or buggy.

Hmm, it seems to be the case...

The doc for the struct 
```json
"109": {
  "id": 109,
  "crate_id": 0,
  "name": "StructBecomesDeprecated",
  "visibility": "public",
  "docs": null,
  "links": {},
  "attrs": [
    "#[deprecated = \"Use NewStruct instead\"]"
  ],
  "deprecation": {
    "since": null,
    "note": "Use NewStruct instead"
  },
  "inner": {
    "struct": {
      "kind": {
        "plain": {
          "fields": [
            107,
            108
          ],
          "has_stripped_fields": false
        }
      },
      ...
    }
  }
},
```
The doc for its first field:
```json
"107": {
  "id": 107,
  "crate_id": 0,
  "name": "field_stays_public",
  "visibility": "public",
  "docs": null,
  "links": {},
  "attrs": [],
  "deprecation": {
    "since": null,
    "note": "Use NewStruct instead"
  },
  "inner": {
    "struct_field": {
      "primitive": "i64"
    }
  }
},
```

test_crates/struct_field_marked_deprecated/new/src/lib.rs
Frank-III: Still the same bahavior

Frank-III:
> Interesting! What happens if they are both deprecated, which deprecation data does the field end up carrying?

The field's deprecation data would end up carrying, then I think this is intended behavior?

test_crates/struct_field_marked_deprecated/new/src/lib.rs
obi1kenobi: Ah my bad, this is a tuple struct not a plain struct, and their formatting rules are different.

obi1kenobi:
It could be! Would you mind checking if it's consistent across other, arguably similar, cases?
- Does deprecating an enum mark its variants as deprecated?
- Does deprecating a tuple/struct variant in an enum mark all their fields as deprecated?
- Does deprecating a union mark its fields as deprecated?
- Does deprecating a trait mark its associated items (methods, types, consts) as deprecated?
- Does deprecating an `impl` block mark everything contained inside as deprecated?

Knowing this will give us a good sense of whether the behavior is consistent and intentional, and will help us figure out what to do next.

Frank-III:
> * Does deprecating an enum mark its variants as deprecated?

Yes, it propagates into struct/tuple variant fields.

> Does deprecating a tuple/struct variant in an enum mark all their fields as deprecated?

Yes

>Does deprecating a union mark its fields as deprecated?

How can find the fields in the doc? it seems to not exist?

> Does deprecating a trait mark its associated items (methods, types, consts) as deprecated?

Yes, I should have added a test for that, so I could identify this yesterday...😭

>Does deprecating an impl block mark everything contained inside as deprecated?

This seems not a valid syntax now?


The test repo: https://github.com/Frank-III/test_deprecated_reproduce

obi1kenobi:
> > * Does deprecating an enum mark its variants as deprecated?
> 
> Yes, it propagates into struct/tuple variant fields.

Does it propagate into the variants themselves too, or just their fields? If just the fields, that would be an interesting and unexpected divergence.

> > Does deprecating a union mark its fields as deprecated?
> 
> How can find the fields in the doc? it seems to not exist?

They should appear as regular struct-like fields, and their IDs are named inside the union's item so you can look them up by ID if you'd like.

> > Does deprecating a trait mark its associated items (methods, types, consts) as deprecated?
> 
> Yes, I should have added a test for that, so I could identify this yesterday...😭

The second-best time to add a test is right now :) Also, let's see if we can amend the lint to say "if the trait is not deprecated but the associated item is" or something to that effect, so we can narrow down the associated item lints to only associated-item-specific deprecations.

> > Does deprecating an impl block mark everything contained inside as deprecated?
> 
> This seems not a valid syntax now?

Oh okay, whew! One fewer case to worry about! This is one of the key (and non-obvious!) difficulties in building `cargo-semver-checks` — it's not obvious what the rules are, what they should be, or what are the constraints and edge cases that dictate those choices!

Frank-III:
>Does it propagate into the variants themselves too, or just their fields?

it propagates into variants.

> They should appear as regular struct-like fields, and their IDs are named inside the union's item so you can look them up by ID if you'd like.

```rs
#[deprecated]
pub union DeprecatedUnion {
    pub unique_field1: i64,
    pub unique_field2: f64,
}
```

get turn into: 

```json
"122": {
  "id": 122,
  "name": "DeprecatedUnion",
  "visibility": "public",
  "docs": "Union case, should trigger the lint for the union and its fields",
  "links": {},
  "attrs": [
    "#[deprecated]"
  ],
  "deprecation": {
    "since": null,
    "note": null
  },
  "inner": {
    "union": {
      "generics": {
        "params": [],
        "where_predicates": []
      },
      "has_stripped_fields": true,
      "fields": [],
      "impls": [
        123,
        124,
        125,
        126,
        127,
        128,
        129,
        130,
        131,
        132,
        133,
        134,
        135
      ]
    }
  }
}
```

>The second-best time to add a test is right now :) Also, let's see if we can amend the lint to say "if the trait is not deprecated but the associated item is" or something to that effect, so we can narrow down the associated item lints to only associated-item-specific deprecations.

would do it in a separate pr then.

obi1kenobi:
Re: the union, can you try regenerating the rustdoc JSON just to be extra sure that it corresponds to the code you're seeing?

Because `"fields": []` is where the field IDs should appear. If it's indeed empty in a way we can reproduce, that's a _serious_ bug we need to report right away.

obi1kenobi:
> Does this mean we need to also fix `Impl` in the rustdoc adapter?

What fix did you have in mind?

I think it's broadly speaking fine for it to have a deprecation field that is always false. But perhaps I've missed something.

Frank-III:
> > Does this mean we need to also fix `Impl` in the rustdoc adapter?
> 
> What fix did you have in mind?
> 
> I think it's broadly speaking fine for it to have a deprecation field that is always false. But perhaps I've missed something.

that makes sense, I don't have one.

> Re: the union, can you try regenerating the rustdoc JSON just to be extra sure that it corresponds to the code you're seeing?
> 
> Because `"fields": []` is where the field IDs should appear. If it's indeed empty in a way we can reproduce, that's a _serious_ bug we need to report right away.

Just tried again, still the same. (I already tried multiple times). 

I have never used union ever could it be due to its natural that: 
> _A union looks like a struct in terms of declaration, but all of its fields exist in the same memory, superimposed over one another_

obi1kenobi:
Ah I missed this bit: `"has_stripped_fields": true,`

That means rustdoc sees the fields and decides to strip them out, for some reason. This is usually if they are `#[doc(hidden)]` or non-pub, and if you don't pass the appropriate flags to disable the stripping: https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/cbd8e99619e036e15ecb280bea146a24050979e6/scripts/regenerate_test_rustdocs.sh#L59

But both with and without stripping, and using last night's nightly, I'm not able to reproduce the empty fields list with the union code you provided.

Can you try reproducing everything from zero to make it as reliable as possible in a bug report to the Rust team? Something copy-pasteable and unambiguous, like:
- update to latest nightly, and record what version it is
- make a new repro crate
- copy in the repro Rust code
- generate rustdoc JSON for it
- show what gets generated for the union item

Frank-III:
> Ah I missed this bit: `"has_stripped_fields": true,`
> 
> That means rustdoc sees the fields and decides to strip them out, for some reason. This is usually if they are `#[doc(hidden)]` or non-pub, and if you don't pass the appropriate flags to disable the stripping: https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/cbd8e99619e036e15ecb280bea146a24050979e6/scripts/regenerate_test_rustdocs.sh#L59
> 
> But both with and without stripping, and using last night's nightly, I'm not able to reproduce the empty fields list with the union code you provided.
> 
> Can you try reproducing everything from zero to make it as reliable as possible in a bug report to the Rust team? Something copy-pasteable and unambiguous, like:
> 
> * update to latest nightly, and record what version it is
> * make a new repro crate
> * copy in the repro Rust code
> * generate rustdoc JSON for it
> * show what gets generated for the union item

Ah sorry my bad, I was trying to test it in a separate repo and did not pass the appropriate flags. 🤦‍♂️

> Does deprecating a union mark its fields as deprecated?

Yes, it does.

obi1kenobi:
Even without the flags for not omitting private and hidden fields, `pub` fields of a deprecated union should still be returned. If you can reliably make rustdoc behave otherwise, I'd still consider that a bug.

Frank-III:
> Even without the flags for not omitting private and hidden fields, `pub` fields of a deprecated union should still be returned. If you can reliably make rustdoc behave otherwise, I'd still consider that a bug.

I think it could be that I am not using latest nightly since I did not pass the flag and it still get to run(and when explicitly pass the flag it would error out). I could not reproduce it either.

Back to this lint, if we add a check that the struct has to not be deprecated for this lint for both baseline and current. when they are both marked deprecated this lint would not get reported, is this the behavior we want? and I think it would be the same for trait method/associate type/associate const.

obi1kenobi:
> when they are both marked deprecated this lint would not get reported, is this the behavior we want?

I think this is the best we're able to get, so it'll have to do :) I think it's fine, it's not worth changing the rustdoc JSON format over a tiny incremental improvement.

src/lints/struct_field_marked_deprecated.ron
```diff
@@ -30,16 +31,12 @@ SemverQuery(
                 item {
                     ... on Struct {
                         visibility_limit @filter(op: "=", value: ["$public"])
+                        deprecated @filter(op: "!=", value: ["$true"])
```
obi1kenobi, line 34: Would you mind adding some comments around these new `deprecated` lines that explain why they are necessary and a good idea?

All the necessary info is in the PR, but we're not going to remember the PR in 3 months and it'll be harder to track that info the next time someone is wondering about it.

obi1kenobi (commented):
One small documentation nitpick, otherwise this is good to merge. Thanks for being super careful and thoughtful here!

Frank-III:
> One small documentation nitpick, otherwise this is good to merge. Thanks for being super careful and thoughtful here!

Thank you! I could do the same for trait method/associated type/associated const in the next one.

obi1kenobi:
That'd be wonderful!

obi1kenobi (approved):

PR #1127, opened 2025-02-06
fix: check if trait is deprecated or not

Frank-III:
following: #1125

### DIFF
--- a/test_outputs/query_execution/trait_method_marked_deprecated.snap
+++ b/test_outputs/query_execution/trait_method_marked_deprecated.snap
@@ -3,30 +3,6 @@ source: src/query.rs
 expression: "&query_execution_results"
 ---
 {
-  "./test_crates/trait_marked_deprecated/": [
-    {
-      "method_name": String("method"),
-      "name": String("TraitToDeprecatedTrait"),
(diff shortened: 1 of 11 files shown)


test_outputs/query_execution/trait_method_marked_deprecated.snap
```diff
@@ -3,30 +3,6 @@ source: src/query.rs
...
-      "name": String("TraitToDeprecatedMessageTrait"),
-      "path": List([
-        String("trait_marked_deprecated"),
-        String("TraitToDeprecatedMessageTrait"),
-      ]),
-      "span_begin_line": Uint64(10),
-      "span_filename": String("src/lib.rs"),
-      "visibility_limit": String("public"),
-    },
-  ],
```
Frank-III, line 29: Hmm I should've catch it up, would be more careful next time!

obi1kenobi (approved):

test_outputs/query_execution/trait_method_marked_deprecated.snap
obi1kenobi: Learning is the most important thing!

PR #1128, opened 2025-02-06
Add a test that fails when generated rustdocs are older than test crate data

suaviloquence:
Caught me in #1126 so I wanted to make a test to prevent it again. Might be overkill.


src/query.rs
```diff
@@ -1082,6 +1083,64 @@ mod tests {
...
+                }
+            }
+
+            Ok(())
+        }
+
+        for id in SemverQuery::all_queries().into_keys() {
+            for version in ["new", "old"] {
+                let test_crate_path = format!("./test_crates/{id}/{version}");
+                // skip lints without test crates
```
obi1kenobi, line 1113: I think it might be better to enumerate test crates explicitly instead of relying on lint names. I suspect we might have at least a few test crates that don't match any lint name, and it would be *very* confusing if this test worked 99% of the time except for a few rare crates.

This change might even allow us to have a stronger invariant here: "if you don't have a `localdata/test_data` entry for each test crate, you need to regenerate because the remaining tests will _definitely_ fail."

src/query.rs
```diff
@@ -1082,6 +1083,64 @@ mod tests {
...
+
+                let localdata_path = format!("./localdata/test_data/{id}/{version}");
+                let mut localdata_times = BTreeSet::new();
+
+                recursive_file_times(localdata_path.clone(), &mut localdata_times).context(
+                    "If this directory doesn't exist, run `scripts/regenerate_test_rustdocs.sh`",
+                )?;
+
+                // if the most recently modified test crate file comes after the earliest localdata
+                // file, it is potentially stale
```
obi1kenobi, line 1129: Should this check versus the _latest_ localdata file? In other words, if the most recent time the test data was generated is prior to the most recent edit to the test files?

Or even better, should it check each test crate's latest timestamp compared to the rustdoc _specific to_ that crate?

I'd expect the local min can be quite old in practice. If I write a test crate, generate its data, and don't touch it ever again, I'll never regenerate its data because the script will skip it each time.

I think the script already compares the modified times between each test crate and its corresponding rustdoc JSON, and that check seems more accurate and more granular than this test.

suaviloquence: Just to clarify, this is currently comparing the oldest file in e.g. `localdata/test_data/function_missing/new` with the newest file in `test_crates/function_missing/new`.

src/query.rs
suaviloquence: IIRC `regenerate_test_rustdocs` makes the rustdocs for any directory in `test_crates` that has a toplevel `new` and `old` directory. I can copy that over.

suaviloquence:
> I'm particularly concerned about the "can't run on CI" part, because tests that don't run on CI have a bad tendency to be a maintenance burden and quietly break all the time. I'd at minimum want to CI-test the edge cases in this test, such as "there's no local data at all" and "there's some stale local data which is missing a specific test crate's data."

I think the reason the caching is failing on CI is that the checkout sets all the modified times to the current time, but the caching action restores the `localdata` files with their old modification times if there's a cache hit (and so regen-test-rustdocs doesnt run).  If there's a way to add new modification times in the action, that would work. 

> I'm not entirely convinced this will be worth the effort to polish up, but I'm not opposed to it and would happily help you move it forward and merge it if you feel it's worth it.

I think it would be helpful for developer experience, especially for new contributors when adding a new crate (and it's gotten me a few times).  I'm definitely willing to work on it, but I don't know if it's worth adding if we have to test the test itself in CI.  let me know what you think, though.

suaviloquence:
@obi1kenobi were you able to get a chance to look at this?

obi1kenobi:
Sorry! I've been traveling this week and I haven't had any time to sit down
and look at things in any depth.

Top of my queue for this weekend.

src/query.rs
```diff
@@ -1082,6 +1084,91 @@ mod tests {
...
+        fn recursive_file_times<P: Into<PathBuf>>(
+            dir: P,
+            set: &mut BTreeSet<SystemTime>,
+        ) -> std::io::Result<()> {
+            for item in fs_err::read_dir(dir)? {
+                let item = item?;
+                let metadata = item.metadata()?;
+                if item.file_name() == "target" || item.file_name() == "Cargo.lock" {
+                    continue;
+                }
```
obi1kenobi, line 1100: I agree with dropping `target` from the recursion and skipping the lockfile. But perhaps we should actively only select `.rs` or `.toml` files in addition?

What I'm worried about is something silly like "macOS made a new `.DS_Store` file in the directory" causing tests to fail. I don't think we can hope to exclude all such possible ephemeral files on all operating systems. If we fail to exclude such a fail, we get a false-positive test result here: the test suite would + should have passed, but this test prevented that.

I think an explicit "allow list" of files we check is better, and a rule like "don't recurse into `target`" plus "check times for `*.rs` and `*.toml` files" should be sufficient.

obi1kenobi (commented):
Sorry for the delay! Overall this looks good — I'm just curious if you have opinions on whether the condition for inclusion in the file times list should be positive ("these files go in") versus negative ("skip these files").

If you agree with my concern but don't have the cycles to make the change right now, don't worry about it. I wasn't able to review this PR in a timely fashion so that's on me — I'll find the time to make the change too.

obi1kenobi (approved):
Excellent work! Thanks again, and apologies for the delay!

PR #1130, opened 2025-02-06
new lint: `type_method_marked_deprecated`

Frank-III:
part of: #57


src/lints/type_method_marked_deprecated.ron
```diff
@@ -0,0 +1,72 @@
+SemverQuery(
+    id: "type_method_marked_deprecated",
+    human_readable_name: "type method #[deprecated] added",
+    description: "A method in a type implementation has been newly marked with #[deprecated].",
```
obi1kenobi, line 4: There's a common technical term for this that we should use:
```suggestion
    description: "An inherent method has been newly marked with #[deprecated].",
```

src/lints/type_method_marked_deprecated.ron
```diff
@@ -0,0 +1,72 @@
...
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+    },
+    error_message: "A type method is now #[deprecated]. Downstream crates will get a compiler warning when using this method.",
+    per_result_error_template: Some("method {{method_name}} in {{owner_type}} {{join \"::\" path}} in {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 71: Usually it's best to keep the per-line templates as short and consistent as possible. Concretely:
- Whenever there's readily-available Rust syntax that can make things shorter, let's use it unless it'll be confusing.
- The emphasis here is on the method, and whether it's on a struct/enum/union isn't _super_ important right now. We can omit that.
- Consistency is quite important. Here, `{{owner_type}}` would have come out capitalized (which isn't legal Rust!) so it would have been confusing.
```suggestion
    per_result_error_template: Some("method {{join \"::\" path}}::{{method_name}} in {{span_filename}}:{{span_begin_line}}"),
```

obi1kenobi (commented):
A couple of minor nitpicks.

Also, could you double-check with a test that if we _simultaneously_ deprecate the type and _also separately_ deprecate its method, that gets flagged on both the type and the method? That _probably_ works already, so it's more of a "just in case something changes in future rustdoc versions" kind of thing, because it would be a major pain to debug later.

Frank-III:
>Also, could you double-check with a test that if we simultaneously deprecate the type and also separately deprecate its method, that gets flagged on both the type and the method? That probably works already, so it's more of a "just in case something changes in future rustdoc versions" kind of thing, because it would be a major pain to debug later.

yeah it should probably work, would add it!

obi1kenobi (approved):
Sweet! Thanks for putting this together 🚀

PR #1135, opened 2025-02-07
Add "certs-source" flag to CLI

alilleybrinker:
Adds a new "certs-source" flag to the CLI, exposed as a new field on `GlobalConfig` with a getter a setter, and with an enum reflecting a choice of either the WebPKI Mozilla certificate store or the native/system one.

This also updates the `tame-index` feature-set to include the native-cert support, and updates the client-building logic for the use of `tame-index` to turn the certificate store options on or off depending on the user's choice.

Note that this commit also removes the call to
`ClientBuilder::http2_prior_knowledge`. That call causes the `tame-index` client to assume the endpoint its accessing supports HTTP2 and just attempt to connect with that protocol rather than trying HTTP1 and upgrading based on server response.

Unfortunately, while Crates.io *does* support HTTP2 (and thus this setting would be correct for it), not all corporate network security software *does*, and in my case the use of HTTP2 prior knowledge assumptions mean requests from `cargo-semver-checks` fail even with the correct certificates.

The result of the removal is that HTTP requests will initiate with an HTTP1 request and then upgrade to HTTP2 when possible (when not going through something like the network security software I'm dealing with).

If that's not desirable to turn off unconditionally, it could _also_ be added as a CLI flag, or inferred based on the choice of certificate store.

---

See this Discussion question for more background: https://github.com/obi1kenobi/cargo-semver-checks/discussions/1134

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -514,6 +514,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "core-foundation"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b55271e5c8c478ad3f38ad24ef34923091e0548492a266d19b3c0b4d82574c63"
+dependencies = [
+ "core-foundation-sys",
+ "libc",
+]
+
+[[package]]
 name = "core-foundation-sys"
 version = "0.8.7"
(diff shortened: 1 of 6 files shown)


alilleybrinker:
Closing, after discussion to determine this wouldn't be supported.

PR #1137, opened 2025-02-10
add actionlint workflow

Frank-III:
close: #1133

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -1,0 +1,22 @@
+name: Lint GitHub Actions workflows
+
+on:
+  push:
+    paths:
+      - ".github/workflows/*.ya?ml"
+  pull_request:
+    paths:
+      - ".github/workflows/*.ya?ml"
+
+jobs:
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Download actionlint
+        id: get_actionlint
+        run: bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
+        shell: bash
+      - name: Check workflow files
+        run: ${{ steps.get_actionlint.outputs.executable }} -color
+        shell: bash
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -15,8 +15,8 @@ jobs:
       - uses: actions/checkout@v4
       - name: Download actionlint
         id: get_actionlint
-        run: bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
+        run: bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/7ddc72c1a2c9ee3a2b16ba1e3da5b46ff456b33c/scripts/download-actionlint.bash)
         shell: bash
       - name: Check workflow files
-        run: ${{ steps.get_actionlint.outputs.executable }} -color
+        run: ${{ steps.get_actionlint.outputs.executable }} -color -ignore 'SC2086|SC2016'
         shell: bash
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v4
       - name: Download actionlint
         id: get_actionlint
-        run: bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/7ddc72c1a2c9ee3a2b16ba1e3da5b46ff456b33c/scripts/download-actionlint.bash)
+        run: bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/2ab3a12c7848f6c15faca9a92612ef4261d0e370/scripts/download-actionlint.bash)
         shell: bash
       - name: Check workflow files
         run: ${{ steps.get_actionlint.outputs.executable }} -color -ignore 'SC2086|SC2016'
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -18,5 +18,6 @@ jobs:
         run: bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/2ab3a12c7848f6c15faca9a92612ef4261d0e370/scripts/download-actionlint.bash)
         shell: bash
       - name: Check workflow files
-        run: ${{ steps.get_actionlint.outputs.executable }} -color -ignore 'SC2086|SC2016'
+        # - https://www.shellcheck.net/wiki/SC2016 - "Expressions don't expand in single quotes, use double quotes for that."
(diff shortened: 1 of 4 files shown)


.github/workflows/actionlint.yml
```diff
@@ -0,0 +1,22 @@
...
+      - ".github/workflows/*.ya?ml"
+
+jobs:
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Download actionlint
+        id: get_actionlint
+        run: bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
```
obi1kenobi, line 18: Let's pin a specific hash instead of using `main`, _just in case_.

We're just `curl | bash`-ing here, and I'd feel safer if it's a specific, reproducible, probably-not-pwned version of it.

.github/workflows/add_new_rust_to_matrix.yml
```diff
@@ -48,16 +48,16 @@ jobs:
...
-                  [$numerical_versions[] | sub("\d+\.(\d+)(?:\.\d+)?", "${1}") | to_number]
+                  [\$numerical_versions[] | sub(\"\\d+\\.(\\d+)(?:\\.\\d+)?\", \"\${1}\") | to_number]
                   | sort
                   | .[-1]
-              ) as $max_named_minor |
-              "1.\($max_named_minor + 2).0"
-            ' .github/workflows/ci.yml)"
+              ) as \$max_named_minor |
+              \"1.\\(\$max_named_minor + 2).0\"
+            " .github/workflows/ci.yml)"
```
obi1kenobi, line 60: I don't think these edits are accurate..? I mean that the new code is correct but so was the old code -- the only change is that the code is less readable now because it has all the escape sequences and the possibility of footguns from un-escaped `$` values.

.github/workflows/ci.yml
```diff
@@ -1884,22 +1886,22 @@ jobs:
...
-          ./scripts/is_version_already_uploaded.sh cargo-semver-checks
+          "./scripts/is_version_already_uploaded.sh" "cargo-semver-checks"
           export EXIT_CODE="$?"
           set -e
           if [[ "$EXIT_CODE" == "7" ]]; then
-            echo 'is_new_version=no' >> $GITHUB_OUTPUT
+            echo "is_new_version=no" >> "$GITHUB_OUTPUT"
           elif [[ "$EXIT_CODE" == "0" ]]; then
-            echo 'is_new_version=yes' >> $GITHUB_OUTPUT
+            echo "is_new_version=yes" >> "$GITHUB_OUTPUT"
```
obi1kenobi, line 1895: These changes are strange too. I agree with `"$GITHUB_OUTPUT"` but everything else seems _at best_ a preference and at worst making the code less good.

obi1kenobi (commented):
Did `actionlint` offer information on why it made some of these changes? A few of them seem quite strange and I'm not sure what to make of them.

.github/workflows/add_new_rust_to_matrix.yml
Frank-III: yeah I think so... should we like ignore some type of `shellcheck` warnings

obi1kenobi: If possible, that'd be great. Otherwise we're making the code worse (less readable + less maintainable) just to satisfy a tool.

This is perhaps even worth a bug report to `shellcheck`, if you're up for it. Might even be a known issue, or something they'd be open to taking a PR for if you're up for that too.

Frank-III: The related issue: https://github.com/rhysd/actionlint/issues/26, I think shellcheck cannot do more than that? but we should be able to specify what kind of error we want to suppress in actionlint.

.github/workflows/actionlint.yml
```diff
@@ -0,0 +1,22 @@
...
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Download actionlint
+        id: get_actionlint
+        run: bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/2ab3a12c7848f6c15faca9a92612ef4261d0e370/scripts/download-actionlint.bash)
+        shell: bash
+      - name: Check workflow files
+        run: ${{ steps.get_actionlint.outputs.executable }} -color -ignore 'SC2086|SC2016'
```
obi1kenobi, line 21: Let's document what the suppressed rules are, to make it easier for future maintenance.
```suggestion
        # Ignored shellcheck rules:
        # - https://www.shellcheck.net/wiki/SC2086 - "Double quote to prevent globbing and word splitting."
        # - https://www.shellcheck.net/wiki/SC2016 - "Expressions don't expand in single quotes, use double quotes for that."
        run: ${{ steps.get_actionlint.outputs.executable }} -color -ignore 'SC2086|SC2016'
```

Btw, what happens if we don't suppress `SC2086`? That rule looks useful _in general_ so I'm wondering what it's flagging here.

Frank-III: > I agree with "$GITHUB_OUTPUT" 

The same as this. Sorry my mistake here, was fighting with my yaml formatter and ignored these two it is😔

Frank-III:
can we retrigger it, it seems to work on my fork

.github/workflows/actionlint.yml
obi1kenobi: No worries, mistakes happen and learning from them is what's important :) Thanks for all the work you've put into cargo-semver-checks already!

obi1kenobi:
The re-triggered run failed in the same manner =/

I think GitHub might implicitly merge `main` into your branch before testing it, which may explain the discrepancy. Try pulling the upstream `main` and merging it into your fork's branch to see if it passes?

Frank-III:
> The re-triggered run failed in the same manner =/
> 
> I think GitHub might implicitly merge `main` into your branch before testing it, which may explain the discrepancy. Try pulling the upstream `main` and merging it into your fork's branch to see if it passes?

Ah, I see, I did not realize new commits include workflow changes, would do!

Frank-III:
> The re-triggered run failed in the same manner =/
> 
> I think GitHub might implicitly merge `main` into your branch before testing it, which may explain the discrepancy. Try pulling the upstream `main` and merging it into your fork's branch to see if it passes?

really sorry about causing another inconvenience for you, so unprofessional...

obi1kenobi:
No worries at all — this GitHub behavior is not at all obvious and you had no way of knowing about it. It's really not a problem, and you've contributed very positively to the project both overall and in this PR that I'm happy to do my part :)

obi1kenobi (approved):

obi1kenobi:
Thanks for putting this PR together! If you'd be open to making similar PRs (with the same ignored shellcheck rule) in the other related projects here, I'd happily merge those too:
- https://github.com/obi1kenobi/trustfall-rustdoc-adapter
- https://github.com/obi1kenobi/trustfall-rustdoc
- https://github.com/obi1kenobi/trustfall

Issue #1132, opened 2025-02-06
New lints: Reordering fields/variants in structs, enums, and variants


obi1kenobi:
We currently don't have the ability to run type-checking lints, but there's a set of lints that don't require type-checking to catch changes in behavior caused by field or variant reordering.

- [ ] public `repr(C)` plain struct has its fields reordered - major, warning

The breakage here is a divergence between the API and ABI: setting field `x` which used to be at index `i` likely no longer touches the same bits as before.

There are caveats: zero-sized types, field size/offset/padding, etc. We don't have that information in rustdoc right now, which is why this is "warning" level for now: we're saying "please have a closer look at this". When we can get that info and conclusively prove breakage, we can upgrade it to "error".

- [ ] struct variant of a public `repr(C)` enum has its fields reordered - major, warning

Same as above, except the "struct" is actually an enum variant.

- [ ] public `#[derive(PartialOrd)]` struct has its fields reordered - major, warning

The breakage here is that the sort order of the struct has changed.

This depends on the ordering (both previously and currently) being a result of `#[derive(PartialOrd)]` _specifically_ and not a manual `impl PartialOrd for MyStruct`. We can check this by looking for `#[automatically_derived]` on the impl, which is added by `#[derive(PartialOrd)]` automatically.

- [ ] public `#[derive(PartialOrd)]` enum has its variants reordered - major, warning

Same reasoning as above.

- [ ] public `#[derive(PartialOrd)]` enum has a variant whose fields were reordered - major warning

Same reasoning as above.

obi1kenobi:
cc @Frank-III the lints we can add here 👆

Frank-III:
I am not very sure how to implement `fields reordered` query, what I could think of is like: field `x` position changed, but we would end up reporting multiple warning of this, do you have some queries I could refer to

obi1kenobi:
I think the "field x position" approach is right. We can't implement
anything smarter right now, as far as I can tell. Just go for it :)

Frank-III:
> public `#[derive(PartialOrd)]` enum has its variants reordered - major, warning

It seems that this would also trigger `enum_no_repr_variant_discriminant_changed` 
what should we do? 
like not just exclude `repr` but also automatically derive `PartialOrd`?

obi1kenobi:
> > public `#[derive(PartialOrd)]` enum has its variants reordered - major, warning
> 
[...]

PR #1138, opened 2025-02-10
new lint: `repr_c_plain_struct_fields_reordered`

Frank-III:
part of: #1132


src/lints/repr_c_plain_struct_fields_reordered.ron
```diff
@@ -0,0 +1,83 @@
...
+    }"#,
+    arguments: {
+        "public": "public",
+        "plain": "plain",
+        "repr": "repr",
+        "c": "C",
+        "true": true,
+    },
+    error_message: "A public repr(C) struct had its fields reordered. This changes the struct's memory layout, breaking FFI use cases that depend on field ordering.",
+    per_result_error_template: Some("field {{field_name}} in struct {{name}} changed position in {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 82: Do you think it's worth adding the position indexes here? I'm a bit torn.

Something vaguely like:
```suggestion
    per_result_error_template: Some("{{name}}.{{field_name}} changed to position {{old_position}}->{{new_position}} in {{span_filename}}:{{span_begin_line}}"),
```

src/lints/repr_c_plain_struct_fields_reordered.ron
```diff
@@ -0,0 +1,83 @@
...
+                                }
+                            }
+                        }
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        field {
```
obi1kenobi, line 59: Just wondering: should this lint check if the number of fields has remained the same (i.e. the change is _purely_ a reordering)? Or should the lint trigger even if e.g. a field was added / removed from the middle of the struct?

obi1kenobi (commented):
A couple of open questions for you to consider. I'm not 100% sure what the answers are — this lint is a bit less clear-cut than the previous ones. I'm curious what you think the best option is here! You've shown good judgment so far, and I'd love to think this through with you before we decide on the right course of action for this lint and similar ones going forward.

src/lints/repr_c_plain_struct_fields_reordered.ron
Frank-III: I actually thought about that too, but I feel that the old and new position is not important information since user could still identify it very quickly

src/lints/repr_c_plain_struct_fields_reordered.ron
Frank-III: hmm, that makes sense. will a field added/removed become a breakage?

obi1kenobi: > they will trigger other lint

Interesting, which other lint do they trigger currently?

Adding / removing fields is _generally_ a breaking change in `repr(C)` public structs, because it changes size and layout -- same as reordering fields. I say _generally_ because the same exception applies as in this lint: if the affected field's type is a "1-ZST" (zero-sized type with alignment `1` AKA no alignment restriction) then that field doesn't matter because it has no representation as "bytes in memory" and no effect on other fields' representation.

We unfortunately cannot yet detect whether a field is 1-ZST or not. This has implications on several of our lints, including this one and the `repr(transparent)` one. (Perhaps it would be a good to have an issue tracking this, I'll make one.)

Ultimately, we want to offer a good experience for our users. Bad experiences would be both:
- a real problem was not spotted nor reported by cargo-semver-checks
- a real problem triggered a large number of lints at the same time, confusing the user and complicating their triage process

The situation here is really tricky, and it feels plausible that we might miss edge cases unless we're quite thorough. Would you be open to writing down some example cases of breakage + a [decision tree](https://en.wikipedia.org/wiki/Decision_tree#Overview) of what lints (present + hypothetical future ones) we want to trigger in which cases? That decision tree will then be a roadmap for how we implement this entire class of "`repr(C)` type is no longer compatible" lints, and will help us identify and discuss the best compromise solution given our current constraints.

obi1kenobi (commented):
I'm now realizing that this lint is in a much more complex landscape than I originally thought. Sorry about that!

On one hand, this is an opportunity to see the other half of the lint-writing process — _deciding_ which lints should exist and what they should cover. This is normally what I do on my own, and usually contributors then take the outputs of my thinking (summarized as implementation issues) and write lints accordingly. If you're interested, I know you're familiar enough with cargo-semver-checks and you've shown excellent attention to detail so I trust you to be able to work through that process! Of course, I'm happy to guide you and be a sounding board throughout.

On the other hand, I also understand this suddenly has much larger scope than originally intended, and it might not be a level of scope and responsibility you want to take on right now — e.g. due to school or other personal considerations. If so, no worries and no hard feelings! I'd be happy to suggest other tasks and take this back onto my plate.

Entirely your call to make. No pressure from me, you have my support either way.

src/lints/repr_c_plain_struct_fields_reordered.ron
Frank-III: would trigger `constructible_struct_adds_field` and `struct_pub_field_missing`(when the field is `pub`), which is expected. 

Yeah I have the same feeling too, say when the 1st field is removed and we do not report this lint but the orders are actually changed which makes this lint more restricted.

> Ultimately, we want to offer a good experience for our users. Bad experiences would be both:
> - a real problem was not spotted nor reported by cargo-semver-checks
> - a real problem triggered a large number of lints at the same time, confusing the user and complicating their triage process

agreed

> Would you be open to writing down some example cases of breakage + a [decision tree](https://en.wikipedia.org/wiki/Decision_tree#Overview) of what lints (present + hypothetical future ones) we want to trigger in which cases?

I am willing to dig into it, but what would the scope of this "`repr(C)` type is no longer compatible" lints would be should we consider the existing lints like the two get triggered

obi1kenobi: Yes, I think we should consider existing lints as much as possible. Though `constructible_struct_adds_field` might not trigger if e.g. the struct is `#[non_exhaustive]`.

Which then raises other interesting questions: is `#[non_exhaustive]` plus `#[repr(C)]` a code smell? Sounds like it, right? One says "we can add fields" and the other says "you can use this in a binary manner." Additions to the _end_ of the struct won't change the layout of prior fields, but would change the size which is _arguably_ not SemVer-guaranteed by default unless explicitly promised. But any additions to the start or middle of the struct (i.e. ones causing reordering of subsequent fields) would still break `#[repr(C)]` _even though_ `#[non_exhaustive]` technically allows them.

Ideally, we want to write down as many such possible combinations of edge cases, reason through what's in the best interest of our users in each case, and then implement lints accordingly. The goal is that in every case, the user is presented with feedback that leaves them feeling well-informed enough to make a sound decision, ideally without overwhelming them with lint reports.

For example, given the case above, I'm actually inclined to believe that checking the field count wasn't the right call since it wouldn't catch the case `#[non_exhaustive] #[repr(C)]` case above. In this case, it's better to raise multiple lints than to let users ship breakage.

Frank-III: I put together some cases we may want to consider in a table: 
| **Case**                                            | **Break / Warning & Should report?**                         |
| --------------------------------------------------- | ------------------------------------------------------------ |
| **1. `#[repr(C)]`, field reordered**                | Should have a lint: **Break** (ABI change)                   |
| **2. `#[repr(C)]`, field added**                    | Should have a lint: **Break** (size and layout change)       |
| **3. `#[repr(C)]`, field removed**                  | Should have a lint: **Break** (size and layout change)       |
| **6. `#[repr(C)]` + `#[non_exhaustive]`**           | Not very sure about this one, I think we should maybe have a lint that detect this and be a Warn? |
| **8. non-`#[repr(C)]` non `#[non_exhaustive]`, add a public/private field** | covered by `constructible_struct_adds_field` & `constructible_struct_adds_private_field` |
| **9. non-`#[repr(C)]`, remove a public field**      | covered by `struct_pub_field_missing`                        |
| **11. private non-`#[repr(C)]` struct changes**                      | Should not report                                                |

These are some cases that take into account existing lints, and in addition I think we should also report if `#[repr(C)]` struct field type changes since this would change the memory layout too, or non-`#[repr(C)]` struct pub field changes(will the `raw type` field in `StructField` be sufficient for this change?)

For `#[repr(C)]` private struct should we consider their fields changes become Breaking or not, maybe yes? since it is still possibly being used by FFI?

Also, I did not take fully into consideration fields being pub/private, I think for `#[repr(C)]` it does not matter, and for pub struct, I think the lints we have already taken into consideration of it.

Overall, maybe for the existing lints mentioned we rule out `#[repr(C)]` struct and have separate check for them individually?

I might not have a good understanding of which should be breaking or in which way we could offer users a better experience, let me know your opinions!

obi1kenobi: Good summary! Based on this framework, what do you think is best to do if we discover the code hits several of these cases at once, such as e.g. 1. and 2.?
- For example, say it added a new non-pub field as the first field in the struct?
- In this case, does the answer change at all if the new field is `pub`? (I think you said "no", just making sure.)

The biggest trickiness is where we choose to trigger multiple lints, vs try to only trigger one lint tightly tailored to the specific case. Similarly, an important question is where we look to split lints into different cases vs merge them together. For example, note how `constructible_struct_adds_field` and `constructible_struct_adds_private_field` are different lints, because even though the _conceptual_ problem is the same, the actual breakage (and explanation) were different enough to justify it.

A concrete example of the above principle that's present here: should we have one lint that checks "size and layout changed" or should we have separate lints for fields being added and removed? In each case, is reordering somehow covered by the "layout changed" portion of that lint (if so, how?) or is that a different lint? If it's a different lint, when do we want both lints to be triggered, and when do we only want one of them?

I'm not sure if we want to rule out `#[repr(C)]` types from the existing lints. We basically have two flavors of breakage: "my `use upstream::SomeType` API use is broken" and "my 'write some bytes to memory, which happen to be `upstream::SomeType` in `repr(C)`' ABI use is broken." It feels like we should maybe warn about these separately, since the affected users and the remediation of the breakage should arguably be different. What do you think?

Frank-III: >Good summary! Based on this framework, what do you think is best to do if we discover the code hits several of  these cases at once, such as e.g. 1. and 2.?
> - For example, say it added a new non-pub field as the first field in the struct?
> - In this case, does the answer change at all if the new field is pub? (I think you said "no", just making sure.)

Yeah my answer is no, but your standpoint is quite strong and reasonable and I am rethinking if this is the good way.


> The biggest trickiness is where we choose to trigger multiple lints, vs try to only trigger one lint tightly tailored to the specific case. Similarly, an important question is where we look to split lints into different cases vs merge them together. For example, note how constructible_struct_adds_field and constructible_struct_adds_private_field are different lints, because even though the conceptual problem is the same, the actual breakage (and explanation) were different enough to justify it.

Thank you again for the detailed explanation! I didn't see the bigger picture but I saw it now and would agree with you since they are technically the same lint as you thought, and it also add some level of maintenance burden.

> I'm not sure if we want to rule out #[repr(C)] types from the existing lints. We basically have two flavors of breakage: "my use upstream::SomeType API use is broken" and "my 'write some bytes to memory, which happen to be upstream::SomeType in repr(C)' ABI use is broken." It feels like we should maybe warn about these separately, since the affected users and the remediation of the breakage should arguably be different. What do you think?

Yeah I think so, but how can we achieve that 🤔

obi1kenobi: > > Good summary! Based on this framework, what do you think is best to do if we discover the code hits several of  these cases at once, such as e.g. 1. and 2.?
> > 
> > * For example, say it added a new non-pub field as the first field in the struct?
> > * In this case, does the answer change at all if the new field is pub? (I think you said "no", just making sure.)
> 
> Yeah my answer is no, but your standpoint is quite strong and reasonable and I am rethinking if this is the good way.

I think it's reasonable to have a "no" answer to the second question.

For the first question, what do you think we should do? The options I see are:
- Raise multiple lints.
- Raise just one of the general-case lints that will exist anyway (if so — which one, reordering or "field added"?)
- Raise a special-cased lint written just for this case, and exclude this case from the general-case lints.

> > I'm not sure if we want to rule out #[repr(C)] types from the existing lints. We basically have two flavors of breakage: "my use upstream::SomeType API use is broken" and "my 'write some bytes to memory, which happen to be upstream::SomeType in repr(C)' ABI use is broken." It feels like we should maybe warn about these separately, since the affected users and the remediation of the breakage should arguably be different. What do you think?
> 
> Yeah I think so, but how can we achieve that 🤔

I think we just keep separate `repr(C)` and non-`repr(C)` lints, and have each set warn about its portion of breakage. The `repr(C)` lints talk about "bytes in memory" and the non-`repr(C)` lints talk about Rust API-level use.

Frank-III: > For the first question, what do you think we should do? The options I see are:
>- Raise multiple lints.
>- Raise just one of the general-case lints that will exist anyway (if so — which one, reordering or "field added"?)
> -Raise a special-cased lint written just for this case, and exclude this case from the general-case lints.

My thought would be either the first or second one, I would go for the first for now, if it turns out that it is not good UX I think we could go for the second one and be "field added"?


Glad that we are kinda sort this out! Could you update #1132 (or a new one covering just `#[repr(C)]` and normal struct and leave it only for `PartialOrd`) with these decisions and any further details you’d like to see in the implementation? Once that’s done, I’ll be happy to start coding it. I think we would need to do:
- modify existed related lint to exclude `#[repr(C)]`
- implement new lint to account for field changes and if both `#[repr(C)]` and `#[non_exhaustive]` are encountered.

obi1kenobi: I'm confused about why #1132 needs editing 🤔

Assuming we're going with the "raise multiple lints" strategy, then it seems to me that this PR's lint is almost exactly right except we should remove the `field @fold @transform(op: "count") @tag(name: "field_count")` clause and the associated filter on that tag. That would seem to check off the first box in #1132 as-is.

I think the `#[non_exhaustive]` case is a separate lint for separate (potential) breakage: "even though the _API_ allows adding new fields, the _ABI_ does not." This might even be a good lint to suggest for clippy, since it it's more of a possible Rust footgun than a SemVer-specific case or something that only applies across new releases. (We also could make it apply to new releases by writing it as "newly adding `#[repr(C)]` to an already `#[non_exhaustive]` struct" or vice versa, etc. This is a separate thing to consider.)

Frank-III: I see, make sense. I was thinking sperate out `#[repr(C)]` with `PartialOrd` since there are some more lints would possibly get added for it, but yeah we could add new issue for just that.

> Assuming we're going with the "raise multiple lints" strategy, then it seems to me that this PR's lint is almost exactly right except we should remove the field @fold @transform(op: "count") @tag(name: "field_count") clause and the associated filter on that tag. That would seem to check off the first box in https://github.com/obi1kenobi/cargo-semver-checks/issues/1132 as-is.

would do !

test_crates/repr_c_plain_struct_fields_reordered/new/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+pub struct DocHiddenStruct {
+    pub b: u16,
+    pub c: u32,
+    pub a: u8,
+}
+
+// Public repr(C) struct with mixed visibility fields
+#[repr(C)]
+pub struct MixedVisibilityStruct {
+    b: u16,     // private field moved
```
obi1kenobi, line 36: We probably don't want to trigger the lint on the private field itself, just on the `pub` ones. If a field isn't `pub`, it arguably shouldn't be accessed in a binary fashion just as much as it shouldn't be accessed via the API.

So for example, the following shouldn't trigger the lint at all:
```rust
#[repr(C)]
pub struct PrivateReordered {
    pub a: i64,
    b: i64,  // this field
    c: i64,  // and this field swap places
    pub d: i64,
}
```

test_crates/repr_c_plain_struct_fields_reordered/new/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+
+// Public non-repr(C) struct with reordered fields - should not trigger
+pub struct RegularStruct {
+    pub b: u16,
+    pub c: u32,
+    pub a: u8,
+}
+
+// Public repr(C) struct with doc(hidden) - should not trigger
+#[doc(hidden)]
```
obi1kenobi, line 25: Oh this is a good idea to test, and gave me an idea for another thing to check: what if the `pub` field itself is `doc(hidden)`?

I think our answer should be similar to the private fields below: we don't trigger on that field, but if it affects the order of anything else then we trigger the lint. If only `doc(hidden)` fields get reordered, then the lint doesn't trigger at all.

src/lints/repr_c_plain_struct_fields_reordered.ron
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "plain": "plain",
+        "repr": "repr",
+        "c": "C",
+        "true": true,
+    },
+    error_message: "A public repr(C) struct had its fields reordered. This changes the struct's memory layout, breaking FFI use cases that depend on field ordering.",
```
obi1kenobi, line 81: Tweak since we don't know for certain whether the layout has changed or not:
```suggestion
    error_message: "A public repr(C) struct had its fields reordered. This can change the struct's memory layout, possibly breaking FFI use cases that depend on field position and order.",
```

src/lints/repr_c_plain_struct_fields_reordered.ron
```diff
@@ -0,0 +1,83 @@
...
+    per_result_error_template: Some("{{name}}.{{field_name}} changed position in {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 82: I've been thinking about this, and I think we should include the position change here. It isn't immediately actionable, but it does improve believability — otherwise people might not realize that "adding a field to the top of the struct" reorders the other fields down layout-wise. If we say `changed position from 1 to 2`, then it's more obvious and will get people to count if they want to disagree.

```suggestion
    per_result_error_template: Some("{{name}}.{{field_name}} moved from position {{position}} to {{new_position}}, in {{span_filename}}:{{span_begin_line}}"),
```

src/lints/repr_c_plain_struct_fields_reordered.ron
```diff
@@ -0,0 +1,83 @@
...
+                            }
+                        }
+
+                        importable_path {
+                            path @tag @output
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+                        
+                        field {
+                            field_name: name @output @tag
```
obi1kenobi, line 32: To implement the `pub` and not-doc-hidden check:
```suggestion
                            public_api_eligible @filter(op: "=", value: ["$true"])
                            field_name: name @output @tag
```

src/lints/repr_c_plain_struct_fields_reordered.ron
```diff
@@ -0,0 +1,83 @@
...
+                        }
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        field {
+                            name @filter(op: "=", value: ["%field_name"])
+                            position @filter(op: "!=", value: ["%position"])
```
obi1kenobi, line 61: To implement the position output:
```suggestion
                            
                            position @filter(op: "!=", value: ["%position"]) @output(name: "new_position")
```

src/lints/repr_c_plain_struct_fields_reordered.ron
```diff
@@ -0,0 +1,83 @@
...
+                            }
+                        }
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        field {
+                            name @filter(op: "=", value: ["%field_name"])
```
obi1kenobi, line 60: To implement the `pub` and not-doc-hidden check:
```suggestion
                            name @filter(op: "=", value: ["%field_name"])
                            public_api_eligible @filter(op: "=", value: ["$true"])
```

obi1kenobi (commented):
Looks very good! Your thorough test cases gave me ideas for more tests, and for refining some of the edge case behaviors, so I left a few suggestions. I'll be very happy to merge this before making the next release.

Frank-III:
> Looks very good! Your thorough test cases gave me ideas for more tests, and for refining some of the edge case behaviors, so I left a few suggestions. I'll be very happy to merge this before making the next release.

Wow, thank you for the detailed review again! would fix it and then implement for others soon!

test_crates/repr_c_plain_struct_fields_reordered/new/src/lib.rs
Frank-III: so if field `b` and `a` swap places, we won't trigger the lint for `b` right just `a` ?

Frank-III: yeah I think I covered both, new added tests are not triggering the lint at all, and one original test has private field and pub field change position

obi1kenobi:
No rush — I appreciate all the hard work you're putting in here!

PR #1150, opened 2025-02-24
Added public api sealed in trait_associated_const_default_removed

### LINKED ISSUE #1109: trait_associated_const_default_removed

geetanshjuneja:
closes #1109

### DIFF
--- a/test_crates/trait_items_hidden_from_public_api/new/src/lib.rs
+++ b/test_crates/trait_items_hidden_from_public_api/new/src/lib.rs
@@ -1,16 +1,16 @@
 pub trait PublicAPISealed {
-    /// Changing the bounds on a hidden type of a public api sealed trait is non breaking    #[doc(hidden)]
+    /// Changing the bounds on a hidden type of a public API sealed trait is non breaking    #[doc(hidden)]
     type Bounded: Send + Sync;
 
-    /// Removing the default value for a hidden associated const of a public api sealed trait is not breaking
+    /// Removing the default value for a hidden associated const of a public API sealed trait is not breaking
     #[doc(hidden)]
     const DEFAULT_REMOVED: i64;
 
-    /// Changing a function signature without a default impl in a public api sealed trait is not breaking
+    /// Changing a function signature without a default impl in a public API sealed trait is not breaking
     #[doc(hidden)]
     fn changed_signature(x: i64, y: i64, z: i64) -> i64;
 
-    /// Removing a default impl for a function in a public api sealed trait is not breaking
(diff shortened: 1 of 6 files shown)


obi1kenobi:
Is there a comment near the test case whose output changed that might need updating? We usually use comments to describe the expected outcome of each test, and if we did so in this case, that comment will now be out of date.

geetanshjuneja:
> Is there a comment near the test case whose output changed that might need updating? We usually use comments to describe the expected outcome of each test, and if we did so in this case, that comment will now be out of date.

Sorry for forgetting to update the comments. I'll update and push it.

obi1kenobi:
No worries at all, and thank you for all your hard work here!

test_crates/trait_items_hidden_from_public_api/new/src/lib.rs
```diff
@@ -1,21 +1,16 @@
-pub trait NonSealed {
-    /// Changing the bounds on an associated type is breaking,
-    /// even if the type was hidden.
-    #[doc(hidden)]
+pub trait PublicAPISealed {
+    /// Changing the bounds on a hidden type of a publically sealed trait is non breaking    #[doc(hidden)]
```
obi1kenobi, line 2: Probably intended as:
```suggestion
    /// Changing the bounds on a hidden type of a public API sealed trait is non breaking
    #[doc(hidden)]
```

test_crates/trait_items_hidden_from_public_api/old/src/lib.rs
```diff
@@ -1,49 +1,33 @@
...
 
-    /// Removing the default value for a hidden associated const of a non-sealed trait is breaking:
-    /// any implementations of this trait are now required to specify a value for the const.
+    /// Removing the default value for a hidden associated const of a public api sealed trait is not breaking
     #[doc(hidden)]
     const DEFAULT_REMOVED: i64 = 5;
 
-    /// Removing an associated function without a default impl in a non-sealed trait is breaking,
-    /// under the same reasoning as for associated constants without a default.
+    /// Removing an associated function without a default impl in a public api sealed trait is not breaking    #[doc(hidden)]
```
obi1kenobi, line 22:
```suggestion
    /// Removing an associated function without a default impl in a public api sealed trait is not breaking
```

obi1kenobi (commented):
As a small nitpick, when editing a file, it's a really good idea to aim for style consistency with the existing code. In the test files, that means small things like keeping periods at the end of sentences and capitalizing `API` (just like in `PublicAPISealed`).

PR #1161, opened 2025-02-25
new lint `partial_ord_struct_field_reordered`

Frank-III:
part of: #1132

### DIFF
--- a/test_crates/partial_ord_struct_fields_reordered/new/src/lib.rs
+++ b/test_crates/partial_ord_struct_fields_reordered/new/src/lib.rs
@@ -1,0 +1,77 @@
+// Public struct with #[derive(PartialOrd)] and reordered fields - should trigger warning
+#[derive(PartialOrd, PartialEq)]
+pub struct PublicStruct {
+    pub b: u16, // moved
+    pub c: u32, // moved
+    pub a: u8,  // moved
+}
+
+// Private struct with #[derive(PartialOrd)] and reordered fields - should not trigger
+#[derive(PartialOrd, PartialEq)]
+struct PrivateStruct {
+    pub b: u16,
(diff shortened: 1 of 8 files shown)


test_crates/partial_ord_struct_fields_reordered/new/src/lib.rs
obi1kenobi, line 1: If I may suggest a few more test cases that are worth including:
- A `pub` struct that used to be `PartialOrd` but is no longer `PartialOrd` shouldn't trigger this lint, even if its fields are reordered. (It should probably trigger a lint for no longer being `PartialOrd` though; not sure if we have that implemented yet.)
- Similarly, reordering fields on a struct that _also_ newly starts deriving `PartialOrd` shouldn't trigger the lint.
- Switching from a derived `PartialOrd` to a hand-impl (or vice versa) while also reordering the fields shouldn't trigger the lint. For example, the hand-impl could account for the ordering change and prevent any behavior change from it.

I believe the current lint implementation is accurate. But these 4 test cases test ways in which our lint query might have been wrong (or might become wrong in the future) and which wouldn't have been caught with today's test suite. So they'll be great to have!

obi1kenobi (commented):
This looks really good! A couple of minor polishing suggestions and this is good to merge 🚀

test_crates/partial_ord_struct_fields_reordered/new/src/lib.rs
Frank-III: thank you, so trapped in the `doc(hidden)` thing and forget to add those!

obi1kenobi:
Btw the nightly failure is unrelated — there's a new rustdoc JSON format version that we don't support yet.

Issue #949, opened 2024-09-26
New lints: opt-in minor lints for added functionality


obi1kenobi:
By default, Rust isn't particularly prescriptive about the difference between a minor and a patch version bump. But many libraries and maintainers may choose to offer stronger guarantees about which kinds of changes should go patches vs minor versions. For example, they may prefer to publish new functionality (new public APIs) in minor versions while keeping patches reserved for bugfixes or other small changes that aren't noticeable at the API level.

Just like detecting removals of public APIs, `cargo-semver-checks` lints can also be used to detect additions to public APIs. This issue is about writing such lints.

These lints should be *opt-in* rather than *opt-out*. They should have defaults of `required_update: Minor` and `lint_level: Allow` in their [definition files](https://github.com/obi1kenobi/cargo-semver-checks/blob/df0ef27cf4d15dca6801ddb82e76dfa4c568b993/src/lints/enum_missing.ron#L5-L6), and [users can opt into them by increasing the lint level](https://github.com/obi1kenobi/cargo-semver-checks#configuration) (e.g. to "warn") as per their preference.

Keep in mind that some API additions are *major breaking changes* and we already have lints for them, such as [adding a variant to an exhaustive enum](https://github.com/obi1kenobi/cargo-semver-checks/blob/df0ef27cf4d15dca6801ddb82e76dfa4c568b993/src/lints/enum_variant_added.ron). New lints added here should take care to not overlap with such major lints — we never want to simultaneously flag the same change as both major and minor, since that would be very confusing to our users!

Lints in this category:
- [ ] adding a new public API struct that is not `#[non_exhaustive]` and has no non-public fields: #495
  - Here we may want to suggest that users add `#[non_exhaustive]` to the struct, since adding new fields or `#[non_exhaustive]` itself later on is a *major breaking change*.
- [ ] adding a new `#[non_exhaustive]` public API struct (one of two complements of the above lint)
- [ ] adding a new public API struct that has a non-public field and no `#[non_exhaustive]` (the remaining complement of the top lint)
- [ ] adding a new public API enum that is not `#[non_exhaustive]`
  - Here we may want to suggest that users add `#[non_exhaustive]` to the enum, since adding new variants or `#[non_exhaustive]` itself later on is a *major breaking change*.
- [ ] adding a new `#[non_exhaustive]` public API enum (the complement of the above lint)
- [ ] adding a new public API union
- [ ] public API type has gained a new auto-trait implementation
[...]

PR #1184, opened 2025-03-13
Add `function_now_const` lint.

shreyans413:
public API function has newly become const


obi1kenobi (commented):
Unfortunately, you've run into the truly hard part of cargo-semver-checks: it's not writing the lints, it's using them to communicate with and educate the user who sees them.

While the lint query is accurate and the lint will fire at the right times, if we merged it now it would simply be frustrating and not helpful to users.

I've added some concrete questions to help guide you toward better error and diagnostic messages. This is a hard skill to earn and will take determination to master! If you decide to commit to earning that skill, I'm happy to review another iteration of this PR :+1:

shreyans413:
> Unfortunately, you've run into the truly hard part of cargo-semver-checks: it's not writing the lints, it's using them to communicate with and educate the user who sees them.
> 
> While the lint query is accurate and the lint will fire at the right times, if we merged it now it would simply be frustrating and not helpful to users.
> 
> I've added some concrete questions to help guide you toward better error and diagnostic messages. This is a hard skill to earn and will take determination to master! If you decide to commit to earning that skill, I'm happy to review another iteration of this PR 👍

Thanks for the detailed feedback! I’ve improved the error message to clearly explain what happened and why it matters, keeping the user's perspective in mind. I’d love to hear your thoughts on the updated version and whether it better communicates the intent of the lint. Looking forward to your review!

obi1kenobi (approved):

obi1kenobi:
Nice, much better!

PR #1185, opened 2025-03-13
Add `function_no_longer_unsafe` lint.

shreyans413:
public API function has stopped being unsafe


src/lints/function_no_longer_unsafe.ron
```diff
@@ -0,0 +1,52 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+    },
+    error_message: "A pub function is no longer `unsafe`, which may break safety assumptions in downstream code.",
```
obi1kenobi, line 50:
```suggestion
    error_message: "A pub function is no longer `unsafe`. Calls to it that use an `unsafe` block may encounter `unused_unsafe` lints. Reverting this change would also be a major breaking change.",
```

obi1kenobi (approved):

shreyans413:
This is also good to go now!

obi1kenobi (approved):

PR #1186, opened 2025-03-13
Add `enum_non_exhaustive_tuple_variant_field_added` lint.

shreyans413:
adding a new field to an existing #[non_exhaustive] tuple variant of a public API enum


obi1kenobi:
Unfortunately the new nightly Rust slightly changed the representation of `repr()` attributes. Now, no matter how they are applied in the source code, they will be combined and normalized in the representation we get: `repr(C, i8)` rather than `repr(i8, C)`, etc. This is causing tests to fail on nightly.

The easiest way to work around this is to ensure our test data is already using `repr()` in normalized form (the way nightly shows the attributes); then older Rust versions will also display the same form, and the tests will be consistent across versions. I know this will cost us a bit of test diversity, but that's an acceptable cost and a more in-depth fix would be too much work.

obi1kenobi:
Apologies, I didn't see you had made the representation changes here so I ended up merging the fix separately.

To simplify the process on your end, I've taken the liberty of removing those commits from this PR and then merged the `main` branch to pick up the CI fixes. I'll now review the PR and, if it looks good, hopefully merge it!

obi1kenobi (approved):

PR #1187, opened 2025-03-13
Add `enum_non_exhaustive_struct_variant_field_added` lint.

shreyans413:
adding a new field to an existing #[non_exhaustive] struct variant of a public API enum

### DIFF
--- a/src/lints/enum_non_exhaustive_struct_variant_field_added.ron
+++ b/src/lints/enum_non_exhaustive_struct_variant_field_added.ron
@@ -1,7 +1,7 @@
 SemverQuery(
     id: "enum_non_exhaustive_struct_variant_field_added",
     human_readable_name: "pub enum #[non_exhaustive] struct variant field added",
-    description: "An enum's #[non_exhaustive] struct variant has a new field.",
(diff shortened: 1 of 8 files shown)


src/lints/enum_non_exhaustive_struct_variant_field_added.ron
```diff
@@ -0,0 +1,75 @@
...
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+        "zero": 0,
+        "non_exhaustive": "#[non_exhaustive]",
+    },
+    error_message: "A pub enum's #[non_exhaustive] struct variant has a new field. Pattern matches that explicitly list fields will now fail to compile.",
```
obi1kenobi, line 73: > Pattern matches that explicitly list fields will now fail to compile.

Are you sure this is true? Did you test this out? Recall that the struct variant is `#[non_exhaustive]`.

It is extremely important that all these descriptions and messages are 100% correct. I cannot stress this enough! It's 100x better to not have a lint at all, than to have a lint that says the wrong thing to the user.

obi1kenobi (approved):

PR #1193, opened 2025-03-15
Add `enum_repr_int_added` lint.

shreyans413:
integer repr (e.g. repr(i8)) newly added to a public API enum that previously did not have any repr


src/lints/enum_repr_int_added.ron
```diff
@@ -0,0 +1,65 @@
+SemverQuery(
+    id: "enum_repr_int_added",
+    human_readable_name: "enum repr(u*)/repr(i*) added",
+    description: "An enum's repr attribute was added.",
```
obi1kenobi, line 4:
```suggestion
    description: "An enum gained an integer repr attribute.",
```

src/lints/enum_repr_int_added.ron
```diff
@@ -0,0 +1,65 @@
...
+                            filename @output
+                            begin_line @output
+                        }
+                    }
+                }
+            }
+            baseline {
+                item {
+                    ... on Enum {
+
```
obi1kenobi, line 41:
```suggestion
```

src/lints/enum_repr_int_added.ron
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "repr": "repr",
+        "repr_regex": "[ui](\\d+|size)",
+        "true": true,
+        "zero": 0,
+    },
+    error_message: "repr(u*) or repr(i*) was added to an enum. This changes its memory representation, and may break FFI use cases. Removing it in the future will cause a major breaking change.",
```
obi1kenobi, line 63:
```suggestion
    error_message: "repr(u*) or repr(i*) was added to an enum. Downstream code may now begin to depend on this enum's binary representation. Undoing this change in the future would be a major breaking change.",
```

test_crates/enum_repr_int_added/new/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+    Baz,
+}
+
+// This enum already has a repr attribute, so adding an integer repr to it should not trigger this lint.
+#[repr(u8)]
+#[repr(C)]
+pub enum CEnumToSeparateU8CEnum {
+    Bar,
+    Baz,
+}
```
obi1kenobi, line 40: This is incorrect. Existing or concurrently added `repr(C)`, `repr(align(2))` etc. are all irrelevant for `repr(i*/u*)` purposes. The lint should flag all cases like this.

This would have been a great opportunity to either do more research (such as reading the reference link in the lint, which describes that different `repr` arguments are orthogonal to each other), or as a last resort ask a question about it. We seem to catch a lot of serious problems in code review of your PRs, which can result in bugs slipping through and is not an efficient use of either your time or mine.

We cannot afford for this trend to continue. Issues caught in code review are a signal that something went badly wrong earlier in the process and needs to be corrected so it never happens again. If you were to apply for a GSoC project here, you'd be expected to do independent research, understand the domain thoroughly, and produce high-quality work. Otherwise, the PRs aren't a contribution — they are a prompt for *me* to fix all the bugs, which would take me longer to do than to just write the PR myself from scratch.

obi1kenobi (changes requested):

test_crates/enum_repr_int_added/new/src/lib.rs
shreyans413: I am really sorry for any inconvenience caused, I did go through the reference link that was mentioned in the lint, I got a bit confused as it was mentioned in the checkbox to write a lint for integer repr (e.g. repr(i8)) newly added to a public API enum that previously did not have any repr. I am still in my learning phase and i have learnt a lot of new things in the past few days. I'll be more careful and I will make sure that it does not happen again.

obi1kenobi (approved):

PR #1194, opened 2025-03-16
feat: add a witness hint for `function_parameter_count_changed` lint

sandptel:
I’ve tried to keep the hint simple and consistent with others(#978,#977): `parameter_count_changed::function_with_parameters_removed(#parameter != <old_count>);`

However, the syntax `(#parameter != <old_count>)` or as alternative `(|parameter| != <old_count>)` might be confusing when compared to standard Rust syntax. Would it be better to replace this with `(...)` or `(...,...,...)` where number of `...` be equal to old_parameter_count for clarity?

>I'm still getting used to reading and writing these, so please forgive any mistakes.

### DIFF
--- a/src/lints/function_parameter_count_changed.ron
+++ b/src/lints/function_parameter_count_changed.ron
@@ -51,4 +51,7 @@ SemverQuery(
     },
     error_message: "A publicly-visible function now takes a different number of parameters.",
     per_result_error_template: Some("{{join \"::\" path}} now takes {{current_parameter_count}} parameters instead of {{old_parameter_count}}, in {{span_filename}}:{{span_begin_line}}"),
+    witness: (
+    hint_template: r#"{{join "::" path}}( #parameters != {{old_parameter_count}} );"#,
+    ),
 )
--- a/src/lints/function_parameter_count_changed.ron
+++ b/src/lints/function_parameter_count_changed.ron
@@ -19,6 +19,10 @@ SemverQuery(
                         }
 
                         old_parameter_: parameter @fold @transform(op: "count") @output @tag(name: "parameters")
+
+                        parameter @fold {
+                        old_parameter_names: name @output
+                        }
                     }
                 }
             }
@@ -52,6 +56,6 @@ SemverQuery(
     error_message: "A publicly-visible function now takes a different number of parameters.",
     per_result_error_template: Some("{{join \"::\" path}} now takes {{current_parameter_count}} parameters instead of {{old_parameter_count}}, in {{span_filename}}:{{span_begin_line}}"),
     witness: (
-    hint_template: r#"{{join "::" path}}( #parameters != {{old_parameter_count}} );"#,
+    hint_template: r#"let witness = |{{join " , " old_parameter_names}}| {{join "::" path}}({{join " , " old_parameter_names}});"#,
     ),
 )
--- a/src/lints/function_parameter_count_changed.ron
+++ b/src/lints/function_parameter_count_changed.ron
@@ -54,6 +54,6 @@ SemverQuery(
     error_message: "A publicly-visible function now takes a different number of parameters.",
     per_result_error_template: Some("{{join \"::\" path}} now takes {{current_parameter_count}} parameters instead of {{old_parameter_count}}, in {{span_filename}}:{{span_begin_line}}"),
     witness: (
-    hint_template: r#"let witness = |{{join ", " old_parameter_names}}| {{join "::" path}}({{join ", " old_parameter_names}});"#,
+        hint_template: r#"let witness = |{{join ", " old_parameter_names}}| {{join "::" path}}({{join ", " old_parameter_names}});"#,
     ),
 )
--- a/src/lints/function_parameter_count_changed.ron
+++ b/src/lints/function_parameter_count_changed.ron
@@ -18,10 +18,8 @@ SemverQuery(
                             public_api @filter(op: "=", value: ["$true"])
                         }
 
-                        old_parameter_: parameter @fold @transform(op: "count") @output @tag(name: "parameters")
-
-                        parameter @fold {
-                        old_parameter_names: name @output
+                        old_parameter_: parameter @fold @transform(op: "count") @output @tag(name: "parameters") {
+                            names: name @output
                         }
                     }
                 }
@@ -56,6 +54,6 @@ SemverQuery(
     error_message: "A publicly-visible function now takes a different number of parameters.",
     per_result_error_template: Some("{{join \"::\" path}} now takes {{current_parameter_count}} parameters instead of {{old_parameter_count}}, in {{span_filename}}:{{span_begin_line}}"),
     witness: (
-    hint_template: r#"let witness = |{{join " , " old_parameter_names}}| {{join "::" path}}({{join " , " old_parameter_names}});"#,
+    hint_template: r#"let witness = |{{join ", " old_parameter_names}}| {{join "::" path}}({{join ", " old_parameter_names}});"#,
     ),
 )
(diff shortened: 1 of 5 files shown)


obi1kenobi:
I just got home and I'm still catching up. I noticed you haven't yet had a chance to iterate based on my feedback, so in the interest of catching up faster, I'll let you do another pass at the code based on the existing feedback before circling back.

sandptel:
> I just got home and I'm still catching up. I noticed you haven't yet had a chance to iterate based on my feedback, so in the interest of catching up faster, I'll let you do another pass at the code based on the existing feedback before circling back.

I am trying to get familiar with the semantics of how querying works (also watching some of your recent youtube streams lately 🙂). Will update on this soon.

sandptel:
I was to trying to figure out how to extract parameter names and found a similar example for struct field names at the [Trustfall Query Playground](https://play.predr.ag/rustdoc#?f=2&q=*3-Structs-and-a-list-of-their-field-names.*lquery---0Crate---2item---4*E-Struct---6name-*o--_0*3-by-default*L-outputs-use-the-name-of-their-field*l*l--_6field-*F---8fields*B-name-*o--_0*3-renamed-output-to-avoid-name-collision*l--_6--*4--*2--*0*J*l*J&v=*C*l*l*J&crate=serde-1.0.163) and I tried to convert the same to output parameter names by adding this `parameter @fold { old_parameter_names: name @output}` ([my exact trustfall query](https://play.predr.ag/rustdoc#?f=2&q=*3-Structs-and-a-list-of-their-field-names.*lquery---0Crate---2item---4*E-Function---6name-*o--_0*3-by-default*L-outputs-use-the-name-of-their-field*l*l--_6parameter-*F---8parameters*B-name-*o--_0*3-renamed-output-to-avoid-name-collision*l--_6--*4--*2--*0*J*l*J&v=*C*l*l*J&crate=serde-1.0.163)). That worked out :) 

Added `parameter @fold { old_parameter_names: name @output}` to the previous lint and added witness like this `witness: (
    hint_template: r#"let witness = |{{join " , " old_parameter_names}}| {{join "::" path}}({{join " , " old_parameter_names}});"#,
    ),`

That generates the following witness output
```
---
source: src/query.rs
description: "Lint `function_parameter_count_changed` did not have the expected witness output.\nSee https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md#testing-witnesses\nfor more information."
expression: "&actual_witnesses"
---
[["./test_crates/parameter_count_changed/"]]
filename = 'src/lib.rs'
begin_line = 3
hint = 'let witness = |_| parameter_count_changed::function_with_a_parameter_added(_);'

[["./test_crates/parameter_count_changed/"]]
filename = 'src/lib.rs'
begin_line = 5
hint = 'let witness = |_ , _| parameter_count_changed::function_with_parameters_removed(_ , _);'

```

I assume there has to be a better method to get function parameter names rather than using this that probably would be an addition to this line `old_parameter_: parameter @fold @transform(op: "count") @output @tag(name: "parameters")` to get `old_parameter_names` directly. Is that so?

src/lints/function_parameter_count_changed.ron
```diff
@@ -19,6 +19,10 @@ SemverQuery(
                         }
 
                         old_parameter_: parameter @fold @transform(op: "count") @output @tag(name: "parameters")
+
+                        parameter @fold {
+                        old_parameter_names: name @output
+                        }
```
obi1kenobi, line 25: Yes, your intuition was correct — here's how to merge them.
```suggestion
                        old_parameter_: parameter @fold @transform(op: "count") @output @tag(name: "parameters") {
                            names: name @output
                        }
```
This block is equivalent to the one you wrote, just shorter. The `names:` alias is appended to the `old_parameter_:` block alias, resulting in a final output name of `old_parameter_names` just like your original query.

Please do pay close attention to the formatting in the future though, the indentation was off and we unfortunately don't have an auto-formatter for these files at the moment.

src/lints/function_parameter_count_changed.ron
```diff
@@ -51,4 +55,7 @@ SemverQuery(
     },
     error_message: "A publicly-visible function now takes a different number of parameters.",
     per_result_error_template: Some("{{join \"::\" path}} now takes {{current_parameter_count}} parameters instead of {{old_parameter_count}}, in {{span_filename}}:{{span_begin_line}}"),
+    witness: (
+    hint_template: r#"let witness = |{{join " , " old_parameter_names}}| {{join "::" path}}({{join " , " old_parameter_names}});"#,
```
obi1kenobi, line 59: Let's leave space only after the comma, not before. This is also what `cargo fmt` does so we should adhere to it as well.

test_outputs/query_execution/function_parameter_count_changed.snap
```diff
@@ -22,6 +24,10 @@ snapshot_kind: text
       "current_parameter_count": Uint64(0),
       "name": String("function_with_parameters_removed"),
       "old_parameter_count": Uint64(2),
+      "old_parameter_names": List([
+        String("_"),
+        String("_"),
+      ]),
```
obi1kenobi, line 30: It's unfortunate that the tests don't name these parameters better, so we can determine e.g. whether the order is correct.

Would you mind editing the test crate to give the parameters names like `a, b` etc. so we can ensure we use appropriate order here?

obi1kenobi (commented):
Looks very nice! Unfortunately it seems like the test crate we previously had didn't cover this new use case particularly well, so it'll need some small edits to ensure our new code continues to be correct in the future.

Hopefully should only be the work of a couple of minutes, and it'll be ready to merge! 🚀

sandptel:
> Looks very nice! Unfortunately it seems like the test crate we previously had didn't cover this new use case particularly well, so it'll need some small edits to ensure our new code continues to be correct in the future.
> 
> Hopefully should only be the work of a couple of minutes, and it'll be ready to merge! 🚀

I understand these changes are small yet I'm currently occupied with my university term project for few days which might cause some delay but will update the changes and push them as soon as possible. Thanks for your patience!

src/lints/function_parameter_count_changed.ron
sandptel: Sure, What do you use for syntax highlighting on these though?

obi1kenobi: AFAIK no editor I've ever used has great nested syntax highlighting for GraphQL in RON files =/ I usually copy out the query into a separate file, syntax-highlight it with GraphQL (it isn't GraphQL semantically, but the syntax is shared so the highlighting works) and then copy it back in.

obi1kenobi:
No rush, thanks for the heads up! Good luck with your term project!

sandptel:
Does this work?

src/lints/function_parameter_count_changed.ron
```diff
@@ -51,4 +53,7 @@ SemverQuery(
     },
     error_message: "A publicly-visible function now takes a different number of parameters.",
     per_result_error_template: Some("{{join \"::\" path}} now takes {{current_parameter_count}} parameters instead of {{old_parameter_count}}, in {{span_filename}}:{{span_begin_line}}"),
+    witness: (
+    hint_template: r#"let witness = |{{join ", " old_parameter_names}}| {{join "::" path}}({{join ", " old_parameter_names}});"#,
```
obi1kenobi, line 57: We don't have an autoformatter for RON files, so please try to be mindful of indentation and formatting by hand. Each nested field should be one more step indented compared to its parent field.
```suggestion
        hint_template: r#"let witness = |{{join ", " old_parameter_names}}| {{join "::" path}}({{join ", " old_parameter_names}});"#,
```

obi1kenobi (approved):
Nice! Merging with a small fix.

PR #1197, opened 2025-03-18
chore: Add witness hint for `function_abi_no_longer_unwind` lint

qstommyshu:
Related to #937 

Hi, I tried to add a witness hint for `function_abi_no_longer_unwind` lint. I've checked #977 , #978 and tried to keep the syntax consistent. 

There are two potential options for the witness hint:
1. Option 1 (Concise): 
```let witness = function_abi_no_longer_unwind::unwind_function_becomes_non_unwind(...);```

	This option is cleaner and shorter. It directly indicates which function violates the semver rules, with the violation type already detailed in the err_message section.
	
2. Option 2 (Verbose):
```let witness = pub extern "C" fn function_abi_no_longer_unwind::unwind_function_becomes_non_unwind(...);```

	This version provides richer information in the witness, allowing the user to easily identify where and what is violating the semver rules by just looking at the witness hint. However, it is a bit longer and includes some redundant details that are also present in the err_message. Additionally, I'm not completely certain if the syntax `pub extern "C" fn {{long_function_path}}` is legit.
	
Please let me know which one suites better for this lint :)

### DIFF
--- a/src/lints/function_abi_no_longer_unwind.ron
+++ b/src/lints/function_abi_no_longer_unwind.ron
@@ -60,4 +60,7 @@ SemverQuery(
     },
     error_message: "A pub fn changed from an unwind-capable ABI to the same-named ABI without unwind. If that function causes an unwind (e.g. by panicking), its behavior is now undefined.",
     per_result_error_template: Some("pub fn {{name}} changed ABI from {{abi_raw_name}} to {{new_abi_raw_name}} in {{span_filename}}:{{span_begin_line}}"),
+    witness: (
+        hint_template: r#"let witness = {{join "::" path}}(...);"#,
+    ),
 )
--- a/test_outputs/witnesses/function_abi_no_longer_unwind.snap.new
+++ b/test_outputs/witnesses/function_abi_no_longer_unwind.snap.new
@@ -1,0 +1,10 @@
+---
+source: src/query.rs
+assertion_line: 847
+description: "Lint `function_abi_no_longer_unwind` did not have the expected witness output.\nSee https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md#testing-witnesses\nfor more information."
+expression: "&actual_witnesses"
+---
+[["./test_crates/function_abi_no_longer_unwind/"]]
+filename = 'src/lib.rs'
+begin_line = 3
+hint = 'let witness = function_abi_no_longer_unwind::unwind_function_becomes_non_unwind(...);'

### CODE REVIEW

**Comment by obi1kenobi** [comment]
Hi! Thanks for opening this PR.

Unfortunately I'm not sure how either witness demonstrates the breakage. I think option 2 isn't valid syntax, and option 1 seems like it'll run just fine with or without an unwinding ABI. Compare this against a witness for a deleted pub function, where the witness code produces an immediate compilation error if used when the function is deleted.

This particular lint is very hard to write a witness for — possibly one of the hardest in the entire suite of lints we have. I'd recommend setting this PR aside for a bit and writing witnesses for a handful of easier cases first, so you can slowly work up the difficulty ladder instead of ramping the difficulty straight to 11.

**Comment by qstommyshu** [comment]
> Hi! Thanks for opening this PR.
> 
> Unfortunately I'm not sure how either witness demonstrates the breakage. I think option 2 isn't valid syntax, and option 1 seems like it'll run just fine with or without an unwinding ABI. Compare this against a witness for a deleted pub function, where the witness code produces an immediate compilation error if used when the function is deleted.
> 
> This particular lint is very hard to write a witness for — possibly one of the hardest in the entire suite of lints we have. I'd recommend setting this PR aside for a bit and writing witnesses for a handful of easier cases first, so you can slowly work up the difficulty ladder instead of ramping the difficulty straight to 11.

Got it, thanks for your reply. I think I might not fully understand the purpose of a witness in a lint, hence I cannot identify the difficulty of adding *witness hint* to a specific lint (I've looked at #977 , #978 and some other lints with `witness: None`, i.e. `derive_helper_attr_removed.ron`, but I don't fully understand the hint tells user the lint violation or why it is impossible to write a *witness hint*).  

My previous understanding of *witness hint* is a hint to tell the user which function violates the semver rules, and the user can quickly go to check the function (I thought that's what the path name is for).

Would you might giving me more materials to read about so that I can get a clearer picture of this? Thanks!

**Comment by obi1kenobi** [comment]
The [`per_result_error_template` in the lint](https://github.com/obi1kenobi/cargo-semver-checks/pull/1197/files#diff-52ec86af228617546c7bf0acbb0f0a83520d28dd67a89dcb6d120355caa874e7R62) already tells the user which function is the problem.

Our "witness" term refers to a program that demonstrates the breakage beyond any doubt: https://predr.ag/blog/semver-violations-are-common-better-tooling-is-the-answer/#automated-validation-via-witnesses

A full witness can be hard to write for a variety of reasons. For example, it needs to be a full Rust program that passes `cargo check` prior to the breakage, not just a bare code snippet like a single Rust statement. Occasionally, a witness even requires specific `Cargo.toml` file contents.

To be a valid witness, it needs to be valid with the baseline version of the code while also being broken in an unambiguously-provable way with the new code: a `cargo check` compilation failure, a a `miri` detection of UB or other unsoundness, etc.

A witness _hint_ is intended to be a lightweight version of the witness, summarizing the _gist_ of the idea behind a full witness.

For example, one possible witness for "this function no longer exists" might look like:
```rust
fn witness(a: RequiredType, b: OtherRequiredType) {
    scanned_crate::deleted_function(a, b);
}
```
where the `a` and `b` types are whatever the function in question required as arguments. This is a valid Rust program! A witness _hint_ doesn't have to be a valid Rust program, and in this case could merely be: `scanned_crate::deleted_function(...);` showing that any downstream call to the function is going to be a compile error.

qstommyshu:
> The [`per_result_error_template` in the lint](https://github.com/obi1kenobi/cargo-semver-checks/pull/1197/files#diff-52ec86af228617546c7bf0acbb0f0a83520d28dd67a89dcb6d120355caa874e7R62) already tells the user which function is the problem.
> 
> Our "witness" term refers to a program that demonstrates the breakage beyond any doubt: [predr.ag/blog/semver-violations-are-common-better-tooling-is-the-answer#automated-validation-via-witnesses](https://predr.ag/blog/semver-violations-are-common-better-tooling-is-the-answer/#automated-validation-via-witnesses)
> 
> A full witness can be hard to write for a variety of reasons. For example, it needs to be a full Rust program that passes `cargo check` prior to the breakage, not just a bare code snippet like a single Rust statement. Occasionally, a witness even requires specific `Cargo.toml` file contents.
> 
> To be a valid witness, it needs to be valid with the baseline version of the code while also being broken in an unambiguously-provable way with the new code: a `cargo check` compilation failure, a a `miri` detection of UB or other unsoundness, etc.
> 
> A witness _hint_ is intended to be a lightweight version of the witness, summarizing the _gist_ of the idea behind a full witness.
> 
> For example, one possible witness for "this function no longer exists" might look like:
> 
> ```rust
> fn witness(a: RequiredType, b: OtherRequiredType) {
>     scanned_crate::deleted_function(a, b);
> }
> ```
> 
> where the `a` and `b` types are whatever the function in question required as arguments. This is a valid Rust program! A witness _hint_ doesn't have to be a valid Rust program, and in this case could merely be: `scanned_crate::deleted_function(...);` showing that any downstream call to the function is going to be a compile error.

Oh, I think I understand what a `witness` mean now, thanks for the explanation. 

Let me explain it in my word just to confirm I really understand it. A "witness" is a piece of valid Rust example code generated by tool, that compiles on the older library code, but will not compile on the new library code ((any valid Rust code can be a "witness" as long as it breaks due to the code changes). 

In your given example, both the program below and `scanned_crate::deleted_function(...);` are valid witnesses because they will run into compile error (as the `deleted_function(a,b);` is deleted, hence the definition of this `fn` does not exist anymore).

```Rust
fn witness(a: RequiredType, b: OtherRequiredType) {
    scanned_crate::deleted_function(a, b);
}
```
 
Or something like below is also valid (although it is unnecessarily complicated in this case)
```Rust
[(a,b), (c,d)].map(|(a, b)| scanned_crate::deleted_function(a, b);)
```

With the same idea, it would be very hard or impolissble to write a witness for `function_abi_no_longer_unwind` lint becuase  changing `extern "C-unwind"` to `extern "C"` does not directly change the functionality of this code to any Rust witnesses (a piece of valid Rust example code), it only changes how the C program using this library calls this function.

Am I understanding it correctly?

obi1kenobi:
You got the general idea. I want to make a few small corrections.

> Let me explain it in my word just to confirm I really understand it. A "witness" is a piece of valid Rust example code generated by tool, that compiles on the older library code, but will not compile on the new library code ((any valid Rust code can be a "witness" as long as it breaks due to the code changes).

This is almost entirely correct, with a small tweak: "not compile" is just one way to prove the witness is valid (i.e. demonstrates real breakage). It can also fail under miri, or misbehave in any other obvious and unquestionable way.

> In your given example, both the program below and `scanned_crate::deleted_function(...);` are valid witnesses because they will run into compile error (as the `deleted_function(a,b);` is deleted, hence the definition of this `fn` does not exist anymore).

The full program is a witness because it fails under `cargo check`. 

`scanned_crate::deleted_function(...);` is not a valid witness because it isn't valid Rust. You can't add that string to a `lib.rs` file and have the compiler accept it with the previous version of the crate.

But `scanned_crate::deleted_function(...);` is a valid witness _hint_, because it demonstrates the crux of the problem. The full witness program is based on the hint, and just expands it into a complete Rust program which necessarily involves some additional boilerplate.

> Or something like below is also valid (although it is unnecessarily complicated in this case)
> 
> ```rust
> [(a,b), (c,d)].map(|(a, b)| scanned_crate::deleted_function(a, b);)
> ```

This would also be a valid witness _hint_, though yes — an unnecessarily complicated one. It isn't a valid _witness_ because it isn't a complete Rust program.

You might be tempted to write something like this as a witness program:
```rust
fn witness() {
    let _ = |a, b| { scanned_crate::deleted_function(a, b); };
}
```
Unfortunately, if `deleted_function()` is generic over either argument, I'm reasonably confident that would leave the closure's type signature poorly specified and rustc might demand type annotations for it. If so, this wouldn't be a valid witness since it only works _some of the time_.

> With the same idea, it would be very hard or impolissble to write a witness for `function_abi_no_longer_unwind` lint becuase changing `extern "C-unwind"` to `extern "C"` does not directly change the functionality of this code to any Rust witnesses (a piece of valid Rust example code), it only changes how the C program using this library calls this function.

It would indeed be very hard.

In principle, I _think_ it would be possible to use Rust to call another Rust library via the C FFI and demonstrate breakage that way. But this would require more research, both on how to confirm the breakage and how precisely to construct the broken program. Hence, not a good first issue :)

qstommyshu:
Thanks for the quick reply! 

I see, "witness" has to be a valid Rust program enclosed by `fn func_name() { /*some valid Rust code inside*/ }`, "witness hint" can be a code snippet that demonstrates the code changes breakage. 

I think I get it now, I'll find some other lint that is easier to start with then! 

But I still have some questions: what exactly is the reason we want a "full witness"? As a "witness hint" already can demo the crux of the breaking change. Will the "full witness" be checked by other checking tools just to make sure it is valid?

obi1kenobi:
> Will the "full witness" be checked by other checking tools just to make sure it is valid?

Yes, precisely. For example, our SemVer study actually checked all witnesses to ensure we aren't reporting invalid results: https://predr.ag/blog/semver-violations-are-common-better-tooling-is-the-answer/

> I see, "witness" has to be a valid Rust program enclosed by `fn func_name() { /*some valid Rust code inside*/ }`

Valid program yes, necessarily wrapped by `fn func_name() {}` no. For example, `use scanned_crate::deleted_function;` is a complete Rust program and a valid witness for public API function deletions. I suspect it's the one we actually use for free functions (not methods, since they can't be imported).

suaviloquence:
<details>

<summary>derailing</summary>

> For example, `use scanned_crate::deleted_function;`

But what if there is (an evil):

```diff
- pub fn deleted_function();
+ pub mod deleted_function {}
```

</details>

obi1kenobi:
Heh, good question. This is why writing witnesses is hard :)

Whenever possible, we should write the "more correct" witness. But if such an ~adversarial edge case is standing between us and having a witness at all, I think we should ship the almost-always-correct witness and not worry about the edge case for the v1 implementation.

I'd expect we'll keep finding edge cases no matter what, so after collecting a bunch of them, we can iterate on the design and try to fix them, then collect new edge cases and iterate some more.

qstommyshu:
Hi, just one more question. Where can I find the syntax references for semver query? I need to manipulate the query a little in order to add a witness hint for a lint.

obi1kenobi:
We unfortunately don't have a syntax reference page up yet. The playgrounds demo all the syntax features and are the best place to look right now:
- https://play.predr.ag/hackernews
- https://play.predr.ag/rustdoc

An older iteration of the query engine has some more docs, but since Trustfall is a ground-up rewrite, we took the opportunity to streamline the syntax with a focus on ergonomics, so that reference is a bit outdated. Including it here just in case: https://graphql-compiler.readthedocs.io/en/latest/language_specification/query_directives.html

PR #1198, opened 2025-03-19
Chore: Add witness hint for `function_unsafe_added` lint

qstommyshu:
Added a witness hint for `function_unsafe_added` lint for #937 .

This witness hint should be valid as the function is not enclosed in a `unsafe{ }` block

### DIFF
--- a/src/lints/function_unsafe_added.ron
+++ b/src/lints/function_unsafe_added.ron
@@ -49,4 +49,7 @@ SemverQuery(
     },
     error_message: "A publicly-visible function became `unsafe`, so calling it now requires an `unsafe` block.",
     per_result_error_template: Some("function {{join \"::\" path}} in file {{span_filename}}:{{span_begin_line}}"),
+    witness: (
+        hint_template: r#"let witness = {{join "::" path}}(...);"#,
+    ),
 )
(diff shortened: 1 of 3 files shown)


obi1kenobi:
I think you may have forgotten to run `cargo insta review`, and seem to have accidentally checked in the not-yet-accepted snapshot file instead. That's why tests are failing.

If you don't already have `cargo-insta` installed, you can get it via `cargo install --locked cargo-insta`. The tool is generally very intuitive, but just in case, here are the docs on how reviewing snapshots works: https://insta.rs/docs/quickstart/#reviewing-snapshots

qstommyshu:
> I think you may have forgotten to run cargo insta review, and seem to have accidentally checked in the not-yet-accepted snapshot file instead. That's why tests are failing.

Updated, thanks for the reminder :) 

It's also cool to learn about `cargo insta`, I've seen this command several times in open source community, but didn't know what exactly it is. I like the idea of creating an "accepted" snap shot and comparing future modification with it to test if code changes breaks the original program behaviour.

obi1kenobi (approved):

obi1kenobi:
> I like the idea of creating an "accepted" snap shot and comparing future modification with it to test if code changes breaks the original program behaviour.

Yes, it's a really neat idea — and it's a very useful building block for more powerful testing primitives too! I gave a talk on this last year: https://www.youtube.com/watch?v=3EFue8PDyic

PR #1199, opened 2025-03-19
Add `enum_struct_variant_changed_kind` lint.

shreyans413:
Addresses the following checkbox in #884 
public API struct variant with at least one public API field changes to another kind


src/lints/enum_struct_variant_changed_kind.ron
```diff
@@ -0,0 +1,77 @@
+SemverQuery(
+    id: "enum_struct_variant_changed_kind",
+    human_readable_name: "An enum struct variant changed kind",
+    description: "A pub enum struct variant with atleast one pub field changed kind",
```
obi1kenobi, line 4: nit: typo
```suggestion
    description: "A pub enum struct variant with at least one pub field changed kind",
```

src/lints/enum_struct_variant_changed_kind.ron
```diff
@@ -0,0 +1,77 @@
...
+                        }
+
+                        variant {
+                            ... on StructVariant {
+                                kind: __typename @output @tag
+                                public_api_eligible @filter(op: "=", value: ["$true"])
+                                attrs @filter(op: "not_contains", value: ["$non_exhaustive"])
+                                variant_name: name @output @tag
+
+                                field @fold @transform(op: "count") @filter(op: "!=", value: ["$zero"]) {
```
obi1kenobi, line 31: I don't 100% remember whether the query optimizer is smart enough to figure out that the count can never be negative, but it does understand monotonicity for sure — I remember adding that :)

So to play it extra safe, let's use the `>` operator here which will definitely be used for optimization.
```suggestion
                                field @fold @transform(op: "count") @filter(op: ">", value: ["$zero"]) {
```

This is just me using insider knowledge, there's no way you could have known this :)

test_crates/enum_struct_variant_changed_kind/new/src/lib.rs
obi1kenobi, line 1: Can you `cargo fmt` all test crate files? I'm pretty sure `cargo fmt` will add trailing commas to all enum variants.

Ideally, configure your editor to do it automatically for you on file save.

obi1kenobi (commented):
Nice job on the query! It isn't a simple one, and I think you nailed it.

A few small nitpicks and this should be good to merge.

src/lints/enum_struct_variant_changed_kind.ron
```diff
@@ -0,0 +1,77 @@
...
+                        importable_path {
+                            path @output @tag
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        variant {
+                            ... on StructVariant {
+                                kind: __typename @output @tag
+                                public_api_eligible @filter(op: "=", value: ["$true"])
+                                attrs @filter(op: "not_contains", value: ["$non_exhaustive"])
```
obi1kenobi, line 28: Reading this more closely, I don't think we need this clause. Accessing the named field is breaking already, exhaustive or not.
```suggestion
```

Can you ensure the tests cover this case appropriately?

src/lints/enum_struct_variant_changed_kind.ron
```diff
@@ -0,0 +1,77 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+        "zero": 0,
+        "non_exhaustive": "#[non_exhaustive]",
```
obi1kenobi, line 73: Once we delete the exhaustiveness check, we won't need this argument either.
```suggestion
```

src/lints/enum_struct_variant_changed_kind.ron
```diff
@@ -0,0 +1,77 @@
...
+    },
+    error_message: "A pub enum's struct variant with at least one pub field has changed to a different kind of enum variant, breaking possible instantiations and field accesses.",
```
obi1kenobi, line 75: We don't know if user code can create that variant (since we don't know if it's exhaustive or not), so let's not mention that. Accessing the pub field is a sufficient example of breakage.
```suggestion
    error_message: "A pub enum's struct variant with at least one pub field has changed to a different kind of enum variant, breaking access to its pub fields.",
```

obi1kenobi:
I think you forgot to update some snapshots.

shreyans413:
> I think you forgot to update some snapshots.

Oh, I forgot to update the snapshots for the newly added lints. I've rebased my branch and am updating them now.

shreyans413:
I think it is all good to go now.

obi1kenobi (approved):

PR #1202, opened 2025-03-21
new lint: `enum_tuple_variant_field_marked_deprecated`

Frank-III:
final part of: #57

### DIFF
--- a/src/lints/enum_tuple_variant_field_marked_deprecated.ron
+++ b/src/lints/enum_tuple_variant_field_marked_deprecated.ron
@@ -27,7 +27,7 @@ SemverQuery(
                                 deprecated @filter(op: "!=", value: ["$true"])
 
                                 field {
-                                    field_position: position @output @tag
+                                    field_name: name @output @tag
                                     public_api_eligible @filter(op: "=", value: ["$true"])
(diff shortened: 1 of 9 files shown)


src/lints/enum_tuple_variant_field_marked_deprecated.ron
```diff
@@ -0,0 +1,82 @@
...
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+    },
+    error_message: "A field in an enum's tuple variant is now #[deprecated]. Downstream crates will get a compiler warning when accessing this field.",
+    per_result_error_template: Some("field at position {{field_position}} in enum variant {{enum_name}}::{{variant_name}} in {{span_filename}}:{{span_begin_line}}"),
```
obi1kenobi, line 81: Any reason not to use the field name and Rust syntax for it? Like `MyEnum::Variant.0` for example?

Frank-III: good call! I think I thought about this but yeah you are right this is more consistent with what we did for all other lints.

obi1kenobi (approved):

PR #1203, opened 2025-03-21
Chore: Add witness hint for `module_missing` lint

qstommyshu:
Add a witness hint for `module_missing`, addressing #937 

This witness hint is valid as `use removed_mod::func;` does not compile in Rust. 

However, there is an edge cases where `use removed_mod::inner_mod` exists in the old version code, but it was replaced by `use removed_mod::inner_function (but the function name is same as the previously removed inner_mod)`, then this witness hint may be a True Negative witness hint.

### DIFF
--- a/src/lints/module_missing.ron
+++ b/src/lints/module_missing.ron
@@ -48,6 +48,6 @@ SemverQuery(
     error_message: "A publicly-visible module cannot be imported by its prior path. A `pub use` may have been removed, or the module may have been renamed, removed, or made non-public.",
     per_result_error_template: Some("mod {{join \"::\" path}}, previously in file {{span_filename}}:{{span_begin_line}}"),
     witness: (
-        hint_template: r#"use {{join "::" path}};"#,
+        hint_template: r#"use {{join "::" path}}::*;"#,
     ),
 )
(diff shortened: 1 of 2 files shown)


qstommyshu:
Sure, here is an edge case example:

Let's say this is our old version code:
```Rust
pub mod bb {
	pub mod will_change_to_fn {}
}
```

and it is changed to:

```Rust
pub mod bb {
	pub fn will_change_to_fn() {}
}
```

In the above case, the witness hint `use bb::will_change_to_fn;` will be able to compile on both version of code (which violates the definition of a witness hint), as there is no syntax differences between `use` a module and `use` a function. 

One way I can think of to resolve this edge case (how to distinguish `use` a module or function) is to exhaustively go down to the lowest level of a module and `use` an element under it. 

However, there is no limitation of levels for Rust nested modules, implementing such an idea probably requires recursive calls.  Also, there are edge cases for this approach as well ... assuming all modules have public elements under it (i.e. a function), then `use bb::will_change_to_fn::some_element;` (as bb::will_change_to_fn` is a module and there is a function called `some_element` under it)  can compile but `use bb::will_change_to_fn::some_element;` can not (as `bb::will_change_to_fn` is a function already, it cannot nest another function inside).

Hope that makes my statement more clear, please let me know you thoughts about it.

obi1kenobi:
Ah, I see. How about a different option then: `use bb::deleted_module::*;`

There's a similar edge case here: enums allow glob imports of all their variants. But clippy by default will complain about both uppercase-named modules and lowercase-named enums, so in practice this edge case will be much much more rare. I think that's probably an acceptable tradeoff for a witness hint.

What do you think?

qstommyshu:
Sounds good to me, I'll include the edge cases description as comments as well.

obi1kenobi (approved):
I didn't see the comments about the edge case you mentioned you might add, but this looks good already so we can add & merge those later if you'd like.

PR #1204, opened 2025-03-21
Chore: Add witness hint for `trait_unsafe_added` lint

qstommyshu:
Add witness hint for `trait_unsafe_added` lint, addressing #937 

This should be a valid witness hint as only `unsafe impl` can implement unsafe trait

### DIFF
--- a/src/lints/trait_unsafe_added.ron
+++ b/src/lints/trait_unsafe_added.ron
@@ -52,6 +52,6 @@ SemverQuery(
     per_result_error_template: Some("trait {{join \"::\" path}} in file {{span_filename}}:{{span_begin_line}}"),
     witness: (
         hint_template: r#"struct MyStruct;
-impl {{join "::" path}} for MyStruct {...};"#,
+impl {{join "::" path}} for MyStruct {...}"#,
     ),
 )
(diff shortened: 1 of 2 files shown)


src/lints/trait_unsafe_added.ron
```diff
@@ -50,4 +50,8 @@ SemverQuery(
     },
     error_message: "A publicly-visible trait became `unsafe`, so implementing it now requires an `unsafe impl` block.",
     per_result_error_template: Some("trait {{join \"::\" path}} in file {{span_filename}}:{{span_begin_line}}"),
+    witness: (
+        hint_template: r#"struct MyStruct;
+impl {{join "::" path}} for MyStruct {...};"#,
```
obi1kenobi, line 55: `impl` blocks don't end in `;`.

Please be very careful with things like this. The purpose of displaying witness code is to boost users' confidence in our tool. Suggesting invalid witness code will *hurt* our credibility, not help it.
```suggestion
impl {{join "::" path}} for MyStruct {...}"#,
```

obi1kenobi, line 55: Canonically, we want the type or value closest to the breakage to be called `Witness` or `witness` as appropriate. We want users to get used to seeing the name and looking for the breakage next to it.
```suggestion
        hint_template: r#"struct Witness;
impl {{join "::" path}} for Witness {...}"#,
```

qstommyshu: Ah, thanks for pointing that out. I will be more careful on my future PRs.

obi1kenobi (approved):

PR #1206, opened 2025-03-22
feat: add a witness hint for struct_missing lint

sandptel:
(PR for Idea Verification) Though given the current test_crates for struct_missing this should work. Yet I feel that it should also cover the cases when the struct contains fields? And if it does it should have witness output that  would be something like `let witness = |a: <type>, b: <type>| struct_missing::WillBeRemovedStruct{foo: a, bar: b};`  for a struct `pub struct WillBeRemovedStruct{foo: <type>, bar:<type>}` that is being completely removed. Shouldn't it?

Also, extracting field names seems/is straightforward but field types is something I can't figure out how to extract. Any help?

### DIFF
--- a/src/lints/struct_missing.ron
+++ b/src/lints/struct_missing.ron
@@ -51,4 +51,7 @@ SemverQuery(
     },
     error_message: "A publicly-visible struct cannot be imported by its prior path. A `pub use` may have been removed, or the struct itself may have been renamed or removed entirely.",
     per_result_error_template: Some("struct {{join \"::\" path}}, previously in file {{span_filename}}:{{span_begin_line}}"),
+    witness: (
+    hint_template: r#"let witness = {{join "::" path}}{};"#,
+    ),
 )
--- a/test_outputs/witnesses/struct_missing.snap
+++ b/test_outputs/witnesses/struct_missing.snap
@@ -1,0 +1,69 @@
+---
+source: src/query.rs
+description: "Lint `struct_missing` did not have the expected witness output.\nSee https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md#testing-witnesses\nfor more information."
+expression: "&actual_witnesses"
+---
+[["./test_crates/move_item_and_reexport/"]]
+filename = 'src/lib.rs'
+begin_line = 40
+hint = 'let witness = move_item_and_reexport::NonEquivalentReorderedGenerics{};'
+
+[["./test_crates/move_item_and_reexport/"]]
+filename = 'src/lib.rs'
+begin_line = 44
+hint = 'let witness = move_item_and_reexport::NonEquivalentRemovedLifetime{};'
+
+[["./test_crates/move_item_and_reexport/"]]
+filename = 'src/lib.rs'
+begin_line = 48
+hint = 'let witness = move_item_and_reexport::NonEquivalentRemovedConst{};'
+
+[["./test_crates/move_item_and_reexport/"]]
+filename = 'src/lib.rs'
+begin_line = 52
+hint = 'let witness = move_item_and_reexport::NonEquivalentRemovedType{};'
+
+[["./test_crates/repr_packed_added_removed/"]]
+filename = 'src/lib.rs'
+begin_line = 11
+hint = 'let witness = repr_packed_added_removed::StructBecomesPackedAndPrivate{};'
+
+[["./test_crates/repr_packed_added_removed/"]]
+filename = 'src/lib.rs'
+begin_line = 15
+hint = 'let witness = repr_packed_added_removed::StructBecomesUnpackedAndPrivate{};'
+
+[["./test_crates/semver_trick_self_referential/"]]
+filename = 'src/lib.rs'
+begin_line = 3
+hint = 'let witness = semver_trick_self_referential::Example{};'
+
+[["./test_crates/struct_missing/"]]
+filename = 'src/lib.rs'
+begin_line = 3
+hint = 'let witness = struct_missing::WillBeRemovedStruct{};'
+
+[["./test_crates/struct_missing/"]]
+filename = 'src/lib.rs'
+begin_line = 6
+hint = 'let witness = struct_missing::PubUseRemovedStruct{};'
+
+[["./test_crates/struct_now_doc_hidden/"]]
+filename = 'src/lib.rs'
+begin_line = 36
+hint = 'let witness = struct_now_doc_hidden::PublicStructThatGoesPrivate{};'
+
+[["./test_crates/struct_pub_field_missing/"]]
+filename = 'src/lib.rs'
+begin_line = 13
+hint = 'let witness = struct_pub_field_missing::StructRemoved{};'
+
+[["./test_crates/struct_with_no_pub_fields_changed_type/"]]
+filename = 'src/lib.rs'
+begin_line = 91
+hint = 'let witness = struct_with_no_pub_fields_changed_type::PubStructChangedToType{};'
+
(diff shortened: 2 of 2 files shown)


obi1kenobi:
> (PR for Idea Verification) Though given the current test_crates for struct_missing this should work. Yet I feel that it should also cover the cases when the struct contains fields?

Yes, the witness also needs to work for structs that have fields :+1:

> And if it does it should have witness output that would be something like `let witness = |a: <type>, b: <type>| struct_missing::WillBeRemovedStruct{foo: a, bar: b};` for a struct `pub struct WillBeRemovedStruct{foo: <type>, bar:<type>}` that is being completely removed. Shouldn't it?

Unfortunately, not all structs can be constructed with a literal. This includes structs that have private fields and `#[non_exhaustive]` structs.

This is the crux of why writing witnesses is hard! We need to come up with a way to demonstrate breakage that works for:
- structs with no fields
- structs that aren't constructible with a literal
- structs with generic parameters
- unit, tuple, and plain (`{ }`) structs

Solving problems like this requires thinking hard about all the possible ways one can use the given element of Rust — in this case, a struct.

> Also, extracting field names seems/is straightforward but field types is something I can't figure out how to extract. Any help?

Unfortunately, I don't think field types are in our schema yet. We need to design what that schema would look like and then implement it.

That particular case isn't beginner-friendly, but if you'd like to work up to it over the span of a few weeks, I'd be happy to suggest an easier schema evolution task that's more suitable for starting out. No pressure, purely if that's something you're interested in!

suaviloquence:
I think ```rs
fn downstream(value: path::to::Struct) {
    let path::to::Struct {..} = value;
}
```

works in all of the cases that you mentioned, but it's not the most user-friendly, so this would be more helpful as a full witness than a witness hint, I think. I agree that this one is maybe not the best first witness hint to add.

obi1kenobi:
Maybe we just avoid overcomplicating it for the witness hint and just do `use path::to::Struct` instead?

In principle someone may have added an module by that name (enum/union/trait are impossible because of the `ImplOwner` check in the query). But:
- modules with uppercase names cause a lint by default, and 
- we have fundamental tension between "all the edge cases" and "easy-to-understand hint."

If we're losing understandability in the _hint_ then it feels like we're not serving our users' best interests. Obviously the full (non-hint) witness needs to work perfectly in all cases, but there we can also afford to e.g. pull public fields / check for other types by the same name / check if the struct is a unit, tuple, or plain etc. and then if need be generate a customized witness for each case.

sandptel:
> Unfortunately, I don't think field types are in our schema yet. We need to design what that schema would look like and then implement it.

There is another lint namely `constructible_struct_adds_field` which I was trying to work on before. Which would require feild types as well (`let witness = |a: <type>, b: <type>| constructible_struct_adds_field::ExhaustivePlainStruct{foo: a, bar: b};` for the `struct pub struct ExhaustivePlainStruct{foo: <type>, bar:<type>}`). This lint seemed as a superset for that.

> That particular case isn't beginner-friendly, but if you'd like to work up to it over the span of a few weeks

I feel that working on this would be worth the time :)

>  I'd be happy to suggest an easier schema evolution task that's more suitable for starting out. No pressure, purely if that's something you're interested in!

If the following can provide a taste of how the above problem could be approached, I would be happy to look into both parallelly. No problems there.

obi1kenobi:
> There is another lint namely `constructible_struct_adds_field` which I was trying to work on before. Which would require feild types as well (`let witness = |a: <type>, b: <type>| constructible_struct_adds_field::ExhaustivePlainStruct{foo: a, bar: b};` for the `struct pub struct ExhaustivePlainStruct{foo: <type>, bar:<type>}`). This lint seemed as a superset for that.

I'm a bit confused, are the types required there? I think the closure could infer its arguments in _most_ cases (unless they are generic), and the struct literal instantiation doesn't seem like it would require types.

> > That particular case isn't beginner-friendly, but if you'd like to work up to it over the span of a few weeks
> 
> I feel that working on this would be worth the time :)
> 
> > I'd be happy to suggest an easier schema evolution task that's more suitable for starting out. No pressure, purely if that's something you're interested in!
> 
> If the following can provide a taste of how the above problem (adding field type to schema) could be approached, I would be happy to look into both parallelly. No problems there.

Sounds good! As a starting point, let's have you try to add information on `T: ?Sized` bounds to generic type arguments in the schema.

First, I'd recommend familiarizing yourself with the [`?Sized` feature](https://doc.rust-lang.org/std/marker/trait.Sized.html) ([nomicon link](https://doc.rust-lang.org/nomicon/exotic-sizes.html?highlight=%3FSized#dynamically-sized-types-dsts)). The docs on it are a bit sparse, so you may have to do some independent searching.

In parallel, I'd recommend writing a lint or two, so you get familiarized with the query syntax and its use cases. Follow our contributing guide to get started, and pick out an `A-lint E-mentor` issue.

[The schema](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/1f0af5c9b23a3da1558d54b0a94ff2a61c48079e/src/rustdoc_schema.graphql) that queries follow describes all the types of data we have access to. While reading about `?Sized`, consider how we may represent that information on [our `GenericTypeParameter` type](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/1f0af5c9b23a3da1558d54b0a94ff2a61c48079e/src/rustdoc_schema.graphql#L2521-L2591). You'll find we've already modeled some related things there, such as synthetic generic types (as in `fn example(x: impl Debug) {}`). Then let me know what you've decided and how you'd design `?Sized`!

sandptel:
> > There is another lint namely `constructible_struct_adds_field` which I was trying to work on before. Which would require feild types as well (`let witness = |a: <type>, b: <type>| constructible_struct_adds_field::ExhaustivePlainStruct{foo: a, bar: b};` for the `struct pub struct ExhaustivePlainStruct{foo: <type>, bar:<type>}`). This lint seemed as a superset for that.
> 
> I'm a bit confused, are the types required there? I think the closure could infer its arguments in _most_ cases (unless they are generic), and the struct literal instantiation doesn't seem like it would require types.

Did this just in an attempt to make witness more informative. Well, if it's addition doesn't affect much, does the rest of witness look fine for `constructible_struct_adds_field` lint ?

obi1kenobi:
> Did this just in an attempt to make witness more informative. Well, if it's addition doesn't affect much, does the rest of witness look fine for `constructible_struct_adds_field` lint ?

I'm confused by the current witness code, and by the question 👆 

Asking because you mention `constructible_struct_adds_field` but this PR touches `struct_missing`. For `constructible_struct_adds_field`, the current witness is broadly on the right track (but not correct yet), while for `struct_missing` the current witness is not correct.

sandptel:
> I'm confused by the current witness code, and by the question 👆
> 
> Asking because you mention `constructible_struct_adds_field` but this PR touches `struct_missing`. For `constructible_struct_adds_field`, the current witness is broadly on the right track (but not correct yet), while for `struct_missing` the current witness is not correct.

I am asking about `constructible_struct_adds_field` and the witness looks like this `let witness = |a, b| constructible_struct_adds_field::ExhaustivePlainStruct{foo: a, bar: b};`, whats wrong with that?

obi1kenobi:
What are `foo` and `bar` in that example? If they are field names, can we reuse those field names as arguments in the closure and make the witness shorter?

sandptel:
> What are `foo` and `bar` in that example? If they are field names, can we reuse those field names as arguments in the closure and make the witness shorter?

Yes `foo` and `bar` are field names and can be reused.

obi1kenobi:
Yeah, looks great!

PR #1208, opened 2025-03-22
Chore: Add witness hint for `enum_struct_variant_field_added` lint

GlitchlessCode:
Adds a witness hint for the `enum_struct_variant_field_added` lint.

## Example
`enum_struct_variant_field_added` outputs the following witness hint for the `./test_crates/enum_struct_variant_field_added/` test.
```rust
match value {
    enum_struct_variant_field_added::PubEnum::Foo { x } => (),
    _ => (),
}
```
This compiles on the old version, but not on the new due to the lack of exhaustiveness in the destructuring of the enum variant.

## Commit Notes
+ Added witness hint for `enum_struct_variant_field_added`
+ Added new parts to query for `enum_struct_variant_field_added` to fetch relevant data for the witness

## Comments
I had to add more to the `trustfall` query for `enum_struct_variant_field_added`, which took a little bit to figure out, but I'm starting to get the hang of it. Is there anywhere that would be good to read up on the syntax and whatnot for `trustfall`? I've found a few myself, but more resources are always welcome. 

Aditionally, feel free to give me any feedback on this one, especially surrounding the new parts of the query. This is what I found that worked, but if there's a better, or more idiomatic way to do this, I would appreciate knowing!

### DIFF
--- a/src/lints/enum_struct_variant_field_added.ron
+++ b/src/lints/enum_struct_variant_field_added.ron
@@ -68,14 +68,6 @@ SemverQuery(
                                 field @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
                                     name @filter(op: "=", value: ["%field_name"])
                                 }
-                            }
-                        }
-
-                        variant {
-                            ... on StructVariant {
-                                name @filter(op: "=", value: ["%variant_name"])
-                                attrs @filter(op: "not_contains", value: ["$non_exhaustive"])
-                                public_api_eligible @filter(op: "=", value: ["$true"])
 
                                 field @fold {
                                     baseline_field_names: name @output
--- a/src/lints/enum_struct_variant_field_added.ron
+++ b/src/lints/enum_struct_variant_field_added.ron
@@ -70,6 +70,18 @@ SemverQuery(
                                 }
                             }
                         }
+
+                        variant {
+                            ... on StructVariant {
+                                name @filter(op: "=", value: ["%variant_name"])
+                                attrs @filter(op: "not_contains", value: ["$non_exhaustive"])
+                                public_api_eligible @filter(op: "=", value: ["$true"])
+
+                                field @fold {
+                                    old_field_names: name @output
+                                }
+                            }
+                        }
                     }
                 }
             }
@@ -83,4 +95,10 @@ SemverQuery(
     },
     error_message: "An enum's exhaustive struct variant has a new field, which has to be included when constructing or matching on this variant.",
     per_result_error_template: Some("field {{field_name}} of variant {{enum_name}}::{{variant_name}} in {{span_filename}}:{{span_begin_line}}"),
+    witness: (
+        hint_template: r#"match value {
+    {{ join "::" path }}::{{ variant_name }} { {{ join ", " old_field_names }} } => (),
+    _ => (),
+}"#,
+    )
 )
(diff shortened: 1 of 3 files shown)


GlitchlessCode:
Just was looking through the code and found `witness_query` on the `Witness` type, may need to switch over to that, let me know if that would be preferred over adding it to the main query.

GlitchlessCode:
Actually now looking over it, it looks like `witness_query` is just meant for `witness_template` not `hint_template`

src/lints/enum_struct_variant_field_added.ron
```diff
@@ -70,6 +70,18 @@ SemverQuery(
...
+
+                        variant {
+                            ... on StructVariant {
+                                name @filter(op: "=", value: ["%variant_name"])
+                                attrs @filter(op: "not_contains", value: ["$non_exhaustive"])
+                                public_api_eligible @filter(op: "=", value: ["$true"])
+
+                                field @fold {
+                                    baseline_field_names: name @output
+                                }
```
obi1kenobi, line 82: You can tuck this portion inside the block above, right next to the `field @fold @transform...` one. That's because everything else in this block is completely repeated compared to the other block, and there are no aggregations or any other modifiers around the other block that would get in the way.

The query will produce the same outputs but will simpler and faster.

obi1kenobi (commented):
One small query tweak needed and this is ready to merge!

src/lints/enum_struct_variant_field_added.ron
GlitchlessCode: That makes sense. I tried doing that off the bat and ran into some issues which is why I broke it out like this, but I'll give it a shot!

obi1kenobi:
> Is there anywhere that would be good to read up on the syntax and whatnot for trustfall?

We don't have a great up-to-date syntax reference at the moment unfortunately. I'm sorry, I know that makes it harder to pick up =/ The playgrounds are the best live demos of Trustfall functionality:
- https://play.predr.ag/hackernews
- https://play.predr.ag/rustdoc

But you're doing great, and I'm happy to help if you get stuck.

obi1kenobi:
> Just was looking through the code and found `witness_query` on the `Witness` type, may need to switch over to that, let me know if that would be preferred over adding it to the main query.

You did the right thing :+1: If the query additions for the hint are too expensive, we should probably reconsider either the lint file structure or the hint structure. But just reading the fields of a variant that's already going to trigger the lint should be really cheap.

src/lints/enum_struct_variant_field_added.ron
obi1kenobi: If you run into the same issue, lmk what it is and we can fix it together.

GlitchlessCode:
Ok, sure enough, moving it into the existing StructVariant query worked just fine! I have no idea what my initial issue was the first time around, hopefully I don't run into it again...

obi1kenobi (approved):

obi1kenobi:
Excellent work!

In general, if you run into any weirdness with queries, please feel free to either give me a ping or post in our Discord (here's an invite if you aren't already there: https://discord.gg/XW6Wgqj6 ). Trustfall is a deterministic query engine, so it should be pretty easy to reproduce any issue and, if needed, turn it into a regression test case and ship a fix. Overall it's quite well-tested but bugs are always possible!

PR #1209, opened 2025-03-22
Chore: Add witness hint for `enum_struct_variant_field_marked_deprecated` lint

GlitchlessCode:
Adds a witness hint for the `enum_struct_variant_field_marked_deprecated` lint.

## Example
`enum_struct_variant_field_marked_deprecated` outputs the following witness hint for the `./test_crates/enum_struct_variant_field_marked_deprecated/` test.
```rust
#[deny(deprecated)]
match value {
    enum_struct_variant_field_marked_deprecated::NormalEnum::StructVariant { field_to_be_deprecated, .. } => (),
    _ => (),
}
```
This compiles on the old version, but not on the new due to the presence of `#[deny(deprecated)]`, as when`field_to_be_deprecated` becomes marked as `deprecated`, compilation fails.

## Comments
I hope using something like `#[deny(deprecated)]` is ok for witnesses, as that was the only practical way I could think of to display the issue in this case.

### DIFF
--- a/src/lints/enum_struct_variant_field_marked_deprecated.ron
+++ b/src/lints/enum_struct_variant_field_marked_deprecated.ron
@@ -80,8 +80,7 @@ SemverQuery(
     error_message: "A field in an enum's struct variant is now #[deprecated]. Downstream crates will get a compiler warning when accessing this field.",
     per_result_error_template: Some("field {{enum_name}}::{{variant_name}}.{{field_name}} in {{span_filename}}:{{span_begin_line}}"),
     witness: (
-        hint_template: r#"#[deny(deprecated)]
-match value {
+        hint_template: r#"match value {
     {{ join "::" path }}::{{ variant_name }} { {{ field_name }}, .. } => (),
     _ => (),
 }"#,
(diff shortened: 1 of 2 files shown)


obi1kenobi:
It looks good, and `#[deny(deprecated)]` is a solid idea. I think we should save it for the eventual "full" witness (a complete Rust program that actually compiles and demonstrates the issue).

For the witness _hint_, I think it's fine for lints that complain about new warnings to just cause that warning. `#[deny(deprecated)]` is probably going to look unfamiliar to users. Instead of the hint helping clarify the issue the lint is flagging, the user may instead end up more confused than before.

So I'd remove the `#[deny(deprecated)]` from the hint for now. Happy to merge it with that change applied.

GlitchlessCode:
I think that's understandable! So just this is fine?

GlitchlessCode:
Wait nevermind, I just realised I forgot to modify the snapshots

obi1kenobi (approved):

obi1kenobi:
Yeah, looks great!

PR #1212, opened 2025-03-23
Chore: Add witness hint for `enum_tuple_variant_field_added` lint

GlitchlessCode:
Adds a witness hint for the `enum_tuple_variant_field_added` lint.

## Example
`enum_tuple_variant_field_added` outputs the following witness hint for the `./test_crates/enum_tuple_variant_field_added/` test.
```rust
match value {
    enum_tuple_variant_field_added::PublicEnum::TupleVariantWithFieldAdded (_, _) => (),
    _ => (),
}
```
This compiles on the old version, but not on the new due to the lack of exhaustiveness in the destructuring of the enum variant.

## Commit Notes
+ Added witness hint for enum_tuple_variant_field_added
+ Added new query for baseline_field_names for extra witness data

## Comments
It took a little bit to figure out what would be the best way to execute this in the handlebars template. Settled on an `#each` iterating over the baseline fields, though the indices are discarded and just replaced with underscores. There's also an `#unless @last` to only join with the commas, rather than always placing it after.


GlitchlessCode:
Same as in #1216, want to redo this to just use `@transform` to count the fields. Will do that ASAP.

obi1kenobi:
Excellent work with the template! I particularly appreciate the attention to detail with `#unless @last` and the trailing comma 💯 Exactly the kind of energy this project needs.

obi1kenobi (approved):
Wonderful!

PR #1213, opened 2025-03-23
feat: add witness hint for `function_marked_deprecated` lint

sandptel:
This adds a witness hint for `function_marked_deprecated` hint. 

## Witness Example
```rust
let witness = |a, b| function_marked_deprecated::function_to_deprecated(a, b);
```
calling the function should invoke the warning 
```
warning: use of deprecated function `function_to_deprecated`
 --> <file-path:line:column>
  |
  |     let witness = |a, b| function_to_deprecated(a, b);
  |             ^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(deprecated)]` on by default
```

## Comments
Had an idea to mark this as `#[deny(deprecated)]` so it would represent breakage by throwing an error. Since its already been discussed in #1209, skipping that. It becomes pretty much similar to #1194.

### DIFF
--- a/test_outputs/witnesses/function_marked_deprecated.snap
+++ b/test_outputs/witnesses/function_marked_deprecated.snap
@@ -1,0 +1,14 @@
+---
+source: src/query.rs
+description: "Lint `function_marked_deprecated` did not have the expected witness output.\nSee https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md#testing-witnesses\nfor more information."
+expression: "&actual_witnesses"
+---
+[["./test_crates/function_marked_deprecated/"]]
+filename = 'src/lib.rs'
+begin_line = 7
+hint = 'let witness = || function_marked_deprecated::function_to_deprecated();'
+
+[["./test_crates/function_marked_deprecated/"]]
+filename = 'src/lib.rs'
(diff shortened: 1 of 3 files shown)


test_outputs/witnesses/function_marked_deprecated.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: src/query.rs
+description: "Lint `function_marked_deprecated` did not have the expected witness output.\nSee https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md#testing-witnesses\nfor more information."
+expression: "&actual_witnesses"
+---
+[["./test_crates/function_marked_deprecated/"]]
+filename = 'src/lib.rs'
+begin_line = 7
+hint = 'let witness = || function_marked_deprecated::function_to_deprecated();'
```
obi1kenobi, line 9: This raises an interesting point: we don't actually need the closure if the function takes no arguments.

Can we make the template output a hint of `function_marked_deprecated::function_to_deprecated();` if the function takes no arguments?

sandptel: Done

obi1kenobi (approved):

PR #1214, opened 2025-03-23
Chore: Add witness hint for `enum_tuple_variant_field_missing ` lint

GlitchlessCode:
Adds a witness hint for the `enum_tuple_variant_field_missing ` lint.

## Example
`enum_tuple_variant_field_missing ` outputs the following witness hint for the `./test_crates/enum_tuple_variant_field_missing/` test.
```rust
match value {
    enum_tuple_variant_field_missing::PublicEnum::TupleVariantWithMissingField(_, _, destructure) => (),
    _ => (),
}
```
This compiles on the old version, but not on the new due to the now missing field in the tuple destructuring.

## Commit Notes
+ Added witness hint for `enum_tuple_variant_field_missing`
+ Added new query for `baseline_field_names` for extra witness data

## Comments
I could use some feedback on this one. I wasn't sure what to call the problem field, so I just called it `destructure`. Let me know if there would be a better name to call it, if anything at all. Second, in the tests there was something relating to `#[doc(hidden)]`, which I wasn't too sure about if it was something I had to worry about or not. If someone could enlighten me on that, it would be appreciated.


GlitchlessCode:
Really just tying up a lot of loose ends today, this one should be good to go too I think!

obi1kenobi (approved):

PR #1216, opened 2025-03-24
Chore: Add witness hint for `enum_tuple_variant_field_marked_deprecated` lint 

GlitchlessCode:
Adds a witness hint for the `enum_tuple_variant_field_marked_deprecated` lint.

## Example
`enum_tuple_variant_field_marked_deprecated` outputs the following witness hint for one of the `./test_crates/enum_tuple_variant_field_marked_deprecated/` tests.
```rust
match value {
    enum_tuple_variant_field_marked_deprecated::TupleVariantEnum::TupleVariant (_, to_be_deprecated, _) => (),
    _ => (),
}
```
This compiles on both versions, but demonstrates the issue with the deprecated field, and would emit an error if compiled.

## Commit Notes
+ Added witness hint for `enum_tuple_variant_field_marked_deprecated`
+ Added new parts to query for `enum_tuple_variant_field_marked_deprecated` to fetch relevant data for the witness

## Comments
Once again, similar to #1214 I wasn't 100% sure on what to call the bound variable (`to_be_deprecated`) when destructuring the tuple, but I think this mostly works. Let me know if you think the name should get changed!


obi1kenobi:
> Once again, similar to #1214 I wasn't 100% sure on what to call the bound variable (`to_be_deprecated`) when destructuring the tuple, but I think this mostly works. Let me know if you think the name should get changed!

We usually call the most-directly-involved value `witness`, so I'd maybe choose that. The naming consistency will help users know where to look in the code example.

src/lints/enum_tuple_variant_field_marked_deprecated.ron
```diff
@@ -31,6 +31,10 @@ SemverQuery(
                                     public_api_eligible @filter(op: "=", value: ["$true"])
                                     deprecated @filter(op: "!=", value: ["$true"])
                                 }
+
+                                field @fold {
+                                    baseline_field_names: name @output
+                                }
```
obi1kenobi, line 37: In a tuple variant these are just always going to be `0, 1, 2, ...` since the fields are purely positional, and not otherwise named.

Consider perhaps just counting the fields instead of outputting a (potentially large) number of known-consecutive numbers? It would make the query a bit more efficient and a bit more obvious.

As another idea, it might be good to mention that this info is used for the witness hint and isn't otherwise necessary for the detection of the issue itself.
```
                                # Figure out how many fields exist in the variant
                                # so we can generate a proper witness hint here.
                                field @fold @transform(op: "count") @output(name: "field_count")
```

src/lints/enum_tuple_variant_field_marked_deprecated.ron
```diff
@@ -79,4 +83,10 @@ SemverQuery(
     },
     error_message: "A field in an enum's tuple variant is now #[deprecated]. Downstream crates will get a compiler warning when accessing this field.",
     per_result_error_template: Some("field {{enum_name}}::{{variant_name}}.{{field_name}} in {{span_filename}}:{{span_begin_line}}"),
+    witness: (
+        hint_template: r#"match value {
+    {{ join "::" path }}::{{ variant_name }} ({{#each baseline_field_names }}{{#if (eq this ../field_name)}}to_be_deprecated{{else}}_{{/if}}{{#unless @last }}, {{/unless}}{{/each}}) => (),
+    _ => (),
+}"#,
```
obi1kenobi, line 90: Excellent use of templating, well done!

Adding my suggestion to name the deprecated field `witness`:
```suggestion
        hint_template: r#"match value {
    {{ join "::" path }}::{{ variant_name }} ({{#each baseline_field_names }}{{#if (eq this ../field_name)}}witness{{else}}_{{/if}}{{#unless @last }}, {{/unless}}{{/each}}) => (),
    _ => (),
}"#,
```

obi1kenobi (commented):
Really nice work! In principle we could merge this right away, but there are a couple of very easy opportunities to polish up the code so it really shines. If you have a second, I'd love to get those tweaks in too!

src/lints/enum_tuple_variant_field_marked_deprecated.ron
GlitchlessCode: Genuinely forgot I could do this, thanks for the reminder, probably able to use this in some of the other `enum_tuple_variant_...` lint's.

GlitchlessCode:
I think this has resolved all the aforementioned improvements? If it hasn't please do let me know!

PR #1218, opened 2025-03-24
Chore: Add witness hint for `enum_variant_added` lint

GlitchlessCode:
Adds a witness hint for the `enum_variant_added` lint.

## Example
`enum_variant_added` outputs the following witness hint for the `./test_crates/enum_variant_added/` test.
```rust
match value { 
    enum_variant_added::EnumWithNewVariant::OldVariant => (),
    enum_variant_added::EnumWithNewVariant::OldStructVariant { .. } => (),
    enum_variant_added::EnumWithNewVariant::OldTupleVariant(..) => (),
}
```
This compiles on the old version, but not on the new due to the lack of exhaustiveness on the match cases. The lack of a catch-all case causes the addition of a new variant to cause a compilation error.

## Commit Notes
+ Added witness hint for enum_variant_added
+ Added new parts to query for enum_variant_added to fetch relevant data for the witness
+ Update test crates to allow for more exhaustive witness format testing

## Comments
Two questions and possible areas for bigger improvements:
1. Is there a better way to combine `baseline_variant_names` and `baseline_variant_kinds`? They both form lists, but each list is separate from the other. Now, each should still always form the same length of list as the other, as they are tied to the same `variant` query, but if there is a way to combine/tuple them together, that would be good to know so I can make that adjustment.
2. Is it okay to modify tests like I've done here? It doesn't change the final result of the test, it all runs the same, I just added some new variants to the tested enum so that I can test the witness printing unit-like, struct-like, and tuple-like enum variants.

### DIFF
--- a/src/lints/enum_variant_added.ron
+++ b/src/lints/enum_variant_added.ron
@@ -46,6 +46,12 @@ SemverQuery(
                         variant @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
                             name @filter(op: "=", value: ["%variant_name"])
                         }
+
+                        variant @fold {
+                            baseline_variant_kinds: __typename @output
+
+                            baseline_variant_names: name @output
+                        }
                     }
                 }
             }
@@ -59,4 +65,9 @@ SemverQuery(
     },
     error_message: "A publicly-visible enum without #[non_exhaustive] has a new variant.",
     per_result_error_template: Some("variant {{enum_name}}:{{variant_name}} in {{span_filename}}:{{span_begin_line}}"),
+    witness: (
+        hint_template: r#"match value { {{#each baseline_variant_kinds }}
+    {{ join "::" ../path }}::{{lookup ../baseline_variant_names @index}}{{#if (eq this "StructVariant")}} { .. }{{/if}}{{#if (eq this "TupleVariant")}}(..){{/if}} => (),{{/each}}
+}"#,
+    ),
 )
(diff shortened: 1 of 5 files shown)


obi1kenobi (approved):

obi1kenobi:
Excellent work 💯 

> ## Comments
> Two questions and possible areas for bigger improvements:
> 
> 1. Is there a better way to combine `baseline_variant_names` and `baseline_variant_kinds`? They both form lists, but each list is separate from the other. Now, each should still always form the same length of list as the other, as they are tied to the same `variant` query, but if there is a way to combine/tuple them together, that would be good to know so I can make that adjustment.

Unfortunately no, at the moment Trustfall only supports what you've used here. Combining the outputs as you describe is a feature on our wish list, but it's surprisingly tricky to design due to having many non-obvious edge cases.

If you might be interested in taking a crack at building that feature (and would be willing to dedicate 3-6 months to work up to it + 3-6 months to actually build and test it), I'd be very happy to support you in that. You definitely have the right combination of coding skill, attention to detail, and tenacity to figure tricky things out to pull it off! That means it will merely be very challenging to build, as opposed to being impossible :)

> 2. Is it okay to modify tests like I've done here? It doesn't change the final result of the test, it all runs the same, I just added some new variants to the tested enum so that I can test the witness printing unit-like, struct-like, and tuple-like enum variants.

Yes! Good call on noticing that the test coverage was insufficient and taking the initiative to improve it. I love to see it!

PR #1219, opened 2025-03-24
Feat: Add custom `repeat` handlebars helper

GlitchlessCode:
Adds a custom handlebars block helper called `repeat`. The goal of this helper is to allow one to iterate a value `n` number of times. The syntax of `repeat` is similar to that of the built-in `each` helper. It assigns to the local block variables `@index`, `@first`, and `@last`. `@index` is a number, `@first` and `@last` are booleans.

## Example
The following example handlebars template assumes the existence of an outputted value called `field_count`, which is a `Number` type representing the number of fields on an enum variant (for example).
```
"path_to_enum::ExampleEnum::ExampleVariant({{#repeat field_count }}_{{#unless @last }}, {{/unless}}{{/times}})"
```
This template string would produce the following output string for an input tuple enum with 3 fields:
```
"path_to_enum::ExampleEnum::ExampleVariant(_, _, _)"
```

## Commit Notes
+ Adds a #repeat block helper, similar to the built-in #each helper

## Comments
This PR is to support the progress of #1212, #1214, and #1216. I'm creating this because as far as I am aware, there is not currently a way to iterate based on a Number in the existing set of handlebars helpers. Let me know if there's any improvements I can make to this!


src/templating.rs
```diff
@@ -59,12 +62,81 @@ handlebars_helper!(multiple_spans: |files: Value, begin_line_numbers: Value| {
     }
 });
 
+// a helper to loop n number of times. similar to #each, has @index, @first, and @last. does not set `this`
+// written without the handlebars_helper! for block access
+// this is largely based on the logic within #each itself
+fn times<'reg, 'rc>(
```
obi1kenobi, line 68:
```suggestion
fn repeat<'reg, 'rc>(
```

src/templating.rs
```diff
@@ -59,12 +62,81 @@ handlebars_helper!(multiple_spans: |files: Value, begin_line_numbers: Value| {
...
+    helper: &Helper<'rc>,
+    registry: &'reg Handlebars<'reg>,
+    ctx: &'rc Context,
+    render_ctx: &mut RenderContext<'reg, 'rc>,
+    output: &mut dyn Output,
+) -> Result<(), RenderError> {
+    let value = helper
+        .param(0)
+        .ok_or(RenderErrorReason::ParamNotFoundForIndex("times", 0))?;
```
obi1kenobi, line 77:
```suggestion
        .ok_or(RenderErrorReason::ParamNotFoundForIndex("repeat", 0))?;
```

src/templating.rs
```diff
@@ -59,12 +62,81 @@ handlebars_helper!(multiple_spans: |files: Value, begin_line_numbers: Value| {
...
+}
+
 pub(crate) fn make_handlebars_registry() -> Handlebars<'static> {
     let mut registry = Handlebars::new();
     registry.set_strict_mode(true);
     registry.register_helper("lowercase", Box::new(lowercase));
     registry.register_helper("join", Box::new(join));
     registry.register_helper("unpack_if_singleton", Box::new(unpack_if_singleton));
     registry.register_helper("multiple_spans", Box::new(multiple_spans));
+    registry.register_helper("times", Box::new(times));
```
obi1kenobi, line 140:
```suggestion
    registry.register_helper("repeat", Box::new(repeat));
```

obi1kenobi (approved):
Looks good! Will rename to `repeat` like we chatted, let me see if I can edit your PR from the web UI without breaking it 🤞

PR #1236, opened 2025-03-25
Feat: Add custom `to_string` handlebars helper

GlitchlessCode:
Adds a `to_string` custom handlebars helper.

## Comments
There's not much to show for this one, it just converts a Number or Bool to a String Value. This isn't useful most of the time, as templating converts a variable to a string anyways. It's main use case is for comparing a number and a string. This is currently needed in #1214 and #1216, as although we can now iterate using the `repeat` helper, it produces Number Values, while `field_name`, which `@index` is compared against, is a String Value, which will always fail the `eq` test. This helper exists to resolve that issue.

### DIFF
--- a/src/templating.rs
+++ b/src/templating.rs
@@ -62,6 +62,17 @@ handlebars_helper!(multiple_spans: |files: Value, begin_line_numbers: Value| {
     }
 });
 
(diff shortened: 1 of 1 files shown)


obi1kenobi:
Love the explanation of why this PR is important — super easy decision to merge given that context!

Issue #1207, opened 2025-03-22
Consider the minimum possible SemVer bump when selecting lints for "no version change" runs


jonhoo:
### Which lint or lints are the issue

`inherent_method_must_use_added`

### Known issues that might be causing this

- [ ] Is the flagged item defined in another crate, or a re-export of such an item? ([#638](https://github.com/obi1kenobi/cargo-semver-checks/issues/638))

### Steps to reproduce the bug with the above code

Trigger any `required_update: Minor` lint in a crate versioned at `0.1.0`.

### Actual Behaviour

The `required_update: Minor` check fails with:
```
semver requires new minor version: 0 major and 1 minor checks failed
```

### Expected Behaviour

No `required_update: Minor` check should fail.

### Verbose Lint Output

```

--- failure inherent_method_must_use_added: inherent method #[must_use] added ---

Description:
An inherent method is now #[must_use]. Downstream crates that did not use its return value will get a compiler lint.
        ref: https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.40.0/src/lints/inherent_method_must_use_added.ron

Failed in:
  method fantoccini::client::RawRequestBuilder::map_request in /home/jon/dev/minor/fantoccini/src/client.rs:320
```

### Generated System Information

System information:
-------------------
#### Software version

cargo-semver-checks 0.40.0

#### Operating system

- OS: Linux (Arch Linux rolling)
- Kernel: 6.13.5-arch1-1

#### Command-line

```bash
/home/jon/.cargo/bin/cargo-semver-checks semver-checks --bugreport 
```

#### cargo version

```
> cargo -V 
cargo 1.86.0-beta.6 (adf9b6ad1 2025-02-28)
```

#### Compile time information

- Profile: release
- Target triple: x86_64-unknown-linux-gnu
- Family: unix
- OS: linux
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2
- Host: x86_64-unknown-linux-gnu

### Additional Context

MAINTAINER EDIT: See [explanation here](https://github.com/obi1kenobi/cargo-semver-checks/issues/1207#issuecomment-2745331161) for full details; the below text is a summary. I'm leaving the original text of this issue below.

[...]

PR #1237, opened 2025-03-25
feat: check for minimum possible semver bump

Frank-III:
closes: #1207 

not fully confident in this approach, please review it!

### DIFF
--- a/src/check_release.rs
+++ b/src/check_release.rs
@@ -32,67 +32,70 @@ enum VersionChangeKind {
     Minimum,
 }
 
+/// Classifies the minimum semantic version change between two versions.
+/// It would panic if either version is not a valid semver.
 fn classify_minimum_semver_version_change(
-    current_version: Option<&str>,
-    baseline_version: Option<&str>,
-) -> Option<VersionChange> {
-    if let (Some(baseline), Some(current)) = (baseline_version, current_version) {
-        let baseline_version =
-            semver::Version::parse(baseline).expect("baseline not a valid version");
-        let current_version = semver::Version::parse(current).expect("current not a valid version");
+    current_version: &str,
+    baseline_version: &str,
+) -> VersionChange {
+    let baseline_version =
+        semver::Version::parse(baseline_version).expect("baseline not a valid version");
+    let current_version =
+        semver::Version::parse(current_version).expect("current not a valid version");
 
-        // Check if versions are identical (ignoring build metadata)
-        if baseline_version.cmp_precedence(&current_version) == Ordering::Equal {
-            if !current_version.pre.is_empty() {
-                // we're on a pre-release: the minimum "next" change is major
-                return Some(VersionChange {
-                    level: ActualSemverUpdate::Major,
-                    kind: VersionChangeKind::Minimum,
-                });
-            }
-            return Some(get_minimum_version_change(&current_version));
+    // Check if versions are identical (ignoring build metadata)
+    if baseline_version.cmp_precedence(&current_version) == Ordering::Equal {
+        if !baseline_version.pre.is_empty() {
+            // the baseline is a pre-release: the minimum "next" change is major
+            return VersionChange {
+                level: ActualSemverUpdate::Major,
+                kind: VersionChangeKind::Minimum,
+            };
         }
+        return get_minimum_version_change(&current_version);
+    }
 
-        // From the cargo reference:
-        // > Initial development releases starting with "0.y.z" can treat changes
-        // > in "y" as a major release, and "z" as a minor release.
-        // > "0.0.z" releases are always major changes. This is because Cargo uses
-        // > the convention that only changes in the left-most non-zero component
-        // > are considered incompatible.
-        // https://doc.rust-lang.org/cargo/reference/semver.html
-        let update_kind = if baseline_version.major != current_version.major {
+    // From the cargo reference:
+    // > Initial development releases starting with "0.y.z" can treat changes
+    // > in "y" as a major release, and "z" as a minor release.
+    // > "0.0.z" releases are always major changes. This is because Cargo uses
+    // > the convention that only changes in the left-most non-zero component
+    // > are considered incompatible.
+    // https://doc.rust-lang.org/cargo/reference/semver.html
+    let update_kind = if baseline_version.major != current_version.major {
+        ActualSemverUpdate::Major
+    } else if baseline_version.minor != current_version.minor {
+        if current_version.major == 0 {
             ActualSemverUpdate::Major
-        } else if baseline_version.minor != current_version.minor {
-            if current_version.major == 0 {
+        } else {
+            ActualSemverUpdate::Minor
+        }
+    } else if baseline_version.patch != current_version.patch {
+        if current_version.major == 0 {
+            if current_version.minor == 0 {
                 ActualSemverUpdate::Major
             } else {
                 ActualSemverUpdate::Minor
             }
-        } else if baseline_version.patch != current_version.patch {
-            if current_version.major == 0 {
-                if current_version.minor == 0 {
-                    ActualSemverUpdate::Major
-                } else {
-                    ActualSemverUpdate::Minor
-                }
-            } else {
-                ActualSemverUpdate::Patch
-            }
-        } else if baseline_version.pre != current_version.pre {
-            // > A pre-release version indicates that the version is unstable and might not satisfy
-            // > the intended compatibility requirements as denoted by its associated normal version
-            // https://semver.org/#spec-item-9
-            ActualSemverUpdate::Major
         } else {
-            unreachable!("Versions are identical.");
-        };
-
-        Some(VersionChange {
-            level: update_kind,
-            kind: VersionChangeKind::Actual,
-        })
+            ActualSemverUpdate::Patch
+        }
+    } else if baseline_version.pre != current_version.pre {
+        // > A pre-release version indicates that the version is unstable and might not satisfy
+        // > the intended compatibility requirements as denoted by its associated normal version
+        // https://semver.org/#spec-item-9
+        ActualSemverUpdate::Major
     } else {
-        None
+        unreachable!(
+            "versions have identical major.minor.patch components, but did not \
+                register as equal when compared: \
+                {current_version:?} vs {baseline_version:?}"
+        );
+    };
+
+    VersionChange {
+        level: update_kind,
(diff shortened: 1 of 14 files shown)


obi1kenobi:
I'm sorry for failing to get this reviewed and merged into this month's release :( I've had an overwhelming amount of work in both personal and work contexts, and I hope to have *a touch* more time to dedicate here in coming months.

I'm very excited about both this PR in particular and about your ongoing contributions to `cargo-semver-checks` in general! Please don't take this delay to mean otherwise!

Frank-III:
> I'm sorry for failing to get this reviewed and merged into this month's release :( I've had an overwhelming amount of work in both personal and work contexts, and I hope to have _a touch_ more time to dedicate here in coming months.
> 
> I'm very excited about both this PR in particular and about your ongoing contributions to `cargo-semver-checks` in general! Please don't take this delay to mean otherwise!

No worries! I know you are quite busy traveling around and will be presenting at RustWeek, looking forward to your talk!

obi1kenobi:
I'm very excited about it — I think it'll be a really good talk! 🤞

src/check_release.rs
```diff
@@ -189,17 +231,23 @@ pub(super) fn run_check_release(
...
             config
                 .shell_warn(
                     "Could not determine whether crate version changed. Assuming no change.",
                 )
                 .expect("print failed");
-            ActualSemverUpdate::NotChanged
+            VersionChange {
+                level: ActualSemverUpdate::NotChanged,
+                kind: VersionChangeKind::Actual,
+            }
```
obi1kenobi, line 248: I'm a bit worried about whether we might hit the same issue in this code path. This code is pretty old and I don't remember when we might fail to determine the crate version change.

Could you look into it, document what you find in a comment here, and decide if we should apply the "minimum semver change" heuristic here as well or not?

src/check_release.rs
```diff
@@ -482,115 +536,211 @@ mod test {
...
     #[test]
     fn classify_pre_same() {
         let baseline = Some("1.0.0-alpha.0");
         let current = Some("1.0.0-alpha.0");
-        let expected = Some(ActualSemverUpdate::NotChanged);
-        let actual = classify_semver_version_change(baseline, current);
+        let expected = Some(VersionChange {
+            level: ActualSemverUpdate::Patch,
+            kind: VersionChangeKind::Minimum,
+        });
```
obi1kenobi, line 634: I think this might be incorrect. Compare this to lines 643-646 below: bumping any component of a pre-release is major relative to that pre-release (but not relative to other stable releases in that range, see lines around 667).

tests/rustdoc_edge_cases.rs
```diff
@@ -233,9 +233,4 @@ fn release_type_flag_minor() {
         ])
         .assert()
         .failure();
-    let mut cmd = Command::cargo_bin("cargo-semver-checks").unwrap();
-    cmd.current_dir("test_crates/type_marked_deprecated/new")
-        .args(["semver-checks", "check-release", "--baseline-root=../old"])
-        .assert()
-        .failure();
```
obi1kenobi, line 240: Why are we deleting this test?

At minimum, the comment on line 225 needs updating. If possible & reasonable, I'd like to preserve this test and just update it to test the new behavior.

obi1kenobi (commented):
I think this is a solid move in the right direction. A bit more polish and this should be good to go. Thank you for putting it together, and sorry for the wait!

Frank-III:
> I think this is a solid move in the right direction. A bit more polish and this should be good to go. Thank you for putting it together, and sorry for the wait!

no worries at all! I would address these issues later today!

obi1kenobi:
You probably know this already, but just in case (since GitHub doesn't do a great job of highlighting it):

When this is ready for another review, make sure to hit the "re-request review" button in the reviewer pane of the PR:
screenshot showing the "re-request review" button, shaped like two arrows forming a cycle, in the PR reviewers pane

tests/rustdoc_edge_cases.rs
Frank-III: Yeah, I agree., I should not delete this.Thee reason I delete it before is that its behavior it is the same as the above test, but yeah we should keep it for clarity
```rs
    let mut cmd = Command::cargo_bin("cargo-semver-checks").unwrap();
    cmd.current_dir("test_crates/type_marked_deprecated/new")
        .args([
            "semver-checks",
            "check-release",
            "--baseline-root=../old",
            "--release-type=minor",
        ])
        .assert()
        .success();
```

src/check_release.rs
Frank-III: yeah I think so! When we are in a pre-release version, the minimum possible change would be a major, not a patch.

src/check_release.rs
Frank-III: I updated the comment to cover all possible cases that would be encountered when deciding which minimal server bump is, let me know if it is clear or not!

Frank-III:
> You probably know this already, but just in case (since GitHub doesn't do a great job of highlighting it):
> 
> When this is ready for another review, make sure to hit the "re-request review" button in the reviewer pane of the PR: screenshot showing the "re-request review" button, shaped like two arrows forming a cycle, in the PR reviewers pane

just wanted to address your comments before I click the button!

src/check_release.rs
```diff
@@ -188,18 +236,34 @@ pub(super) fn run_check_release(
...
 
+    // Determine the version change between baseline and current versions.
+    // This can happen in three ways:
+    // 1. User explicitly specified a release type (--release-type flag)
+    // 2. We can determine it from the version strings
+    // 3. We fall back to assuming no change
+    //
+    // Note: classify_minimum_semver_version_change() will return None in two cases:
+    // - When either baseline_version or current_version is None (unknown version)
+    // - When the versions can't be parsed as valid semver
```
obi1kenobi, line 247: I think failing to parse the versions as valid semver will panic, actually. There seems to be an `expect()` there.

Since we're making `classify_minimum_semver_version_change()` more complex, consider doing a bit of refactoring to make this more clear. How about we change that function to take two `&str` values, and only call it on line 254 below if both `Option<&str>` values are `Option::Some`? Then the "returns `None` if a version is `None`" becomes obvious. A doc comment on the function could mention that it panics if the versions are not valid SemVer.

src/check_release.rs
```diff
@@ -15,15 +16,43 @@ use crate::{
...
         let current_version = semver::Version::parse(current).expect("current not a valid version");
 
+        // Check if versions are identical (ignoring build metadata)
+        if baseline_version.cmp_precedence(&current_version) == Ordering::Equal {
+            if !current_version.pre.is_empty() {
+                // we're on a pre-release: the minimum "next" change is major
+                return Some(VersionChange {
+                    level: ActualSemverUpdate::Major,
+                    kind: VersionChangeKind::Minimum,
+                });
```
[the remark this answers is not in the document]
obi1kenobi: This would have been what I would have expected here:
```suggestion
            if !baseline_version.pre.is_empty() {
                // the baseline is a pre-release: the minimum "next" change is major
                return Some(VersionChange {
                    level: ActualSemverUpdate::Major,
                    kind: VersionChangeKind::Minimum,
                });
```

tests/rustdoc_edge_cases.rs
```diff
@@ -233,9 +233,12 @@ fn release_type_flag_minor() {
         ])
         .assert()
         .failure();
+
+    // Running the same command without the flag should pass.
+    // This is because the detected minimum version change is semver-minor,
```
obi1kenobi, line 238: Actually, can we change the `type_marked_deprecated` test crates' versions to 1.0.0 instead, so that this test continues to fail like the comment originally said? I think that's more in the spirit of this test.

If you don't mind, I'd love a parallel test to this one that uses the `function_marked_deprecated` test crates (which are currently also v0.1.0) that ensures the `.success()` case that is the current edit here.

obi1kenobi (commented):
Just some nitpicks, otherwise I think this is good.

If you don't have time or energy to apply them, I'm happy to merge this and apply the suggested tweaks myself afterward.

Thanks again for putting this together!

src/check_release.rs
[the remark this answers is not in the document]
Frank-III: >Can you see if you can trigger a bug based on the fact that we're checking current_version.pre and not baseline_version.pre here? Or are they guaranteed to be equal because of the cmp_precedence() call? 

yeah, exactly this case.

```rs
baseline_version.cmp_precedence(&current_version)
// equivalent to 
Ord::cmp(
    &(self.major, self.minor, self.patch, &self.pre),
    &(other.major, other.minor, other.patch, &other.pre),
)
```

tests/lib_tests.rs
```diff
@@ -12,7 +12,7 @@ fn major_required_bump_if_breaking_change() {
     let (_crate_name, crate_report) = report.crate_reports().iter().next().unwrap();
     let required_bump = crate_report.required_bump().unwrap();
     assert_eq!(required_bump, ReleaseType::Major);
-    assert_eq!(crate_report.detected_bump(), ActualSemverUpdate::NotChanged);
+    assert_eq!(crate_report.detected_bump(), ActualSemverUpdate::Minor);
```
obi1kenobi, line 15:
```suggestion
    // The "old" and "new" crates have the same version.
    // The detected bump is the minimum-possible SemVer bump in a new release.
    // Since the crates are v0.1.0, the minimum possible bump is minor.
    assert_eq!(crate_report.detected_bump(), ActualSemverUpdate::Minor);
```

tests/rustdoc_edge_cases.rs
```diff
@@ -239,3 +239,14 @@ fn release_type_flag_minor() {
         .assert()
         .failure();
 }
+
+/// This test ensures that minimum semver version changes for 0.y.z versions
+/// are detected correctly without the need for the `--release-type` flag.
```
obi1kenobi, line 244:
```suggestion
/// This test ensures that minimum semver version changes for 0.y.z versions
/// are detected correctly without the need for the `--release-type` flag.
/// In this test, the crate has SemVer-minor lints, and the minimum possible
/// SemVer bump is minor (due to the 0.y.z version), so no lints are raised.
```

obi1kenobi (approved):
Excellent work! Thank you!

I'm adding a couple of extra comments to prod our memory better the next time we need to read these tests, and then merging + closing the linked issue.

Issue #1223, opened 2025-03-25
`ImplOwner` method's receiver changes from `&self` to `&mut self`


obi1kenobi:

KhaledSayed04:
For this issue, what should happen when there is another method with the same name still takes immutable reference? the lint shouldn't report this as a breaking change? or it should detect this by the order & type & number of parameters?

Old code:
`impl Foo {
    fn foo(&self) {}
    fn foo(&self, x: i32) {}
}`

new code:
`impl Foo {
    fn foo(&mut self) {} 
    fn foo(&self, x: i32) {}
}`

obi1kenobi:
I believe the proposed new code wouldn't compile: that looks to me like a
duplicate definition compile error in today's Rust.

Good question though! Thinking through edge cases like this is exactly the
energy we need in this project.

KhaledSayed04:
Is this a valid id for this issue? "method_receiver_ref_mutability_increased" or "method_receiver_self_ref_mutability_increased" to differentiate it from the case where the kind is Box<Self> or any other kind?

KhaledSayed04:
I opened a pull request for this issue. I would appreciate it if you could take a look and give me your feedback.

obi1kenobi:
> Is this a valid id for this issue?

Yeah, this is perfect :+1:

> "method_receiver_ref_mutability_increased" or "method_receiver_self_ref_mutability_increased" to differentiate it from the case where the kind is Box or any other kind?

I find "mutability increased" confusing, because it isn't a continuous spectrum. It's either mutable or not. Naming is hard :)

Another question is whether we want to differentiate between kind `Self` and other ones like `Box` or `Rc`. I think the current way you've written the lint, it will only look for kind `Self`, so changing `self: &Rc<Self>` to `self: &mut Rc<Self>` won't be caught here even though it's also a `&` to `&mut` change. Is this on purpose?

I'd recommend trying to come up with a plan for which cases end up in which lints, then suggest lint names for all cases in this lint group so we can make sure they are all consistent with each other and with other existing lints. I suspect that handling `&Rc<Self>` to `&mut Rc<Self>` will end up being best in this lint and not elsewhere, but I'm not 100% sure and I'm willing to entertain other ideas backed up by good reasons :)

KhaledSayed04:
>I find "mutability increased" confusing, because it isn't a continuous spectrum. It's either
>mutable or not. Naming is hard   :)

[...]

PR #1243, opened 2025-04-01
Add type_method_receiver_ref_now_mutable lint

KhaledSayed04:
Add type_method_receiver_ref_now_mutable lint to solve #1223 issue.

### DIFF
--- a/src/lints/type_method_receiver_ref_now_mutable.ron
+++ b/src/lints/type_method_receiver_ref_now_mutable.ron
@@ -1,0 +1,84 @@
+SemverQuery(
+    id: "type_method_receiver_ref_now_mutable",
+    human_readable_name: "type method's receiver now requires mutable reference instead of immutable reference",
+    description: "A type method's receiver changed from '&self' to '&mut self'.",
+    required_update: Major,   
+    lint_level: Deny,   
+    reference_link: "https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#mutable-references",
+    query: r#"
+    {
+        CrateDiff {
+            baseline {
+                item {
+                    ... on ImplOwner {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        name @output
+                        owner_type: __typename @tag @output
+                        
+                        importable_path {
+                            path @tag @output
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        inherent_impl {
+                            public_api_eligible @filter(op: "=", value: ["$true"])
+
+                            method {  
+                                method_name: name @output @tag
+                                public_api_eligible @filter(op: "=", value: ["$true"])
+  
+                                receiver {
+                                    by_reference @filter(op: "=", value: ["$true"])
+                                    method_receiver_kind: kind @tag @output
+                                } 
+
+                                span_: span @optional {
+                                    filename @output
+                                    begin_line @output
+                                    end_line @output
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            current {
+                item {
+                    ... on ImplOwner {
+                        visibility_limit @filter(op: "=", value: ["$public"])   
+                        __typename @filter(op: "=", value: ["%owner_type"])
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+                          
+
+                        inherent_impl {
+                            public_api_eligible @filter(op: "=", value: ["$true"])
+
+                            method { 
+                                name @filter(op: "=", value: ["%method_name"])  
+                                public_api_eligible @filter(op: "=", value: ["$true"])
+                                
+                                receiver {
+                                    by_mut_reference @filter(op: "=", value: ["$true"])
+                                    kind @filter(op: "=", value: ["%method_receiver_kind"])
+                                } 
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public", 
+        "true": true, 
+    },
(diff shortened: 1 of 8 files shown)


KhaledSayed04:
I'm not sure if these may be more suitable for reference_link:
1. https://doc.rust-lang.org/book/ch05-03-method-syntax.html#defining-methods
2. https://doc.rust-lang.org/reference/items/associated-items.html#methods

src/lints/type_method_receiver_ref_now_mutable.ron
```diff
@@ -0,0 +1,84 @@
...
+                            public_api_eligible @filter(op: "=", value: ["$true"])
+
+                            method {  
+                                method_name: name @output @tag
+                                public_api_eligible @filter(op: "=", value: ["$true"])
+  
+                                receiver {
+                                    by_reference @filter(op: "=", value: ["$true"])
+                                    method_receiver_kind: kind @tag @output
+                                }
```
Frank-III, line 33: should avoid extra space

```suggestion
                                }
```

src/lints/type_method_receiver_ref_now_mutable.ron
```diff
@@ -0,0 +1,84 @@
...
+                item {
+                    ... on ImplOwner {
+                        visibility_limit @filter(op: "=", value: ["$public"])   
+                        __typename @filter(op: "=", value: ["%owner_type"])
+
+                        importable_path {
+                            path @filter(op: "=", value: ["%path"])
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
```
Frank-III, line 55: remove extra new-line
```suggestion
```

src/lints/type_method_receiver_ref_now_mutable.ron
```diff
@@ -0,0 +1,84 @@
...
+                            public_api_eligible @filter(op: "=", value: ["$true"])
+
+                            method { 
+                                name @filter(op: "=", value: ["%method_name"])  
+                                public_api_eligible @filter(op: "=", value: ["$true"])
+                                
+                                receiver {
+                                    by_mut_reference @filter(op: "=", value: ["$true"])
+                                    kind @filter(op: "=", value: ["%method_receiver_kind"])
+                                }
```
Frank-III, line 67: could you clean all of these extra space?
```suggestion
                                }
```

src/lints/type_method_receiver_ref_now_mutable.ron
```diff
@@ -0,0 +1,84 @@
...
+                            path @tag @output
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        inherent_impl {
+                            public_api_eligible @filter(op: "=", value: ["$true"])
+
+                            method {  
+                                method_name: name @output @tag
+                                public_api_eligible @filter(op: "=", value: ["$true"])
```
Frank-III, line 28: it seems that we are not limit it to just public method:
```suggestion
```

src/lints/type_method_receiver_ref_now_mutable.ron
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }"#,
+    arguments: {
+        "public": "public", 
+        "true": true, 
+    },
+    error_message: "A type method's receiver now requires a mutable reference instead of immutable one. Uses of this method that supplied the previous mutability state will be broken.",
+    per_result_error_template: Some("method {{join \"::\" path}}::{{method_name}} in {{span_filename}}:{{span_begin_line}}"), 
+    witness: (
+        hint_template: r#"make the reference mutable for method '{{ join "::" path }}::{{ method_name }}'"#
```
Frank-III, line 82: this witness is wrong.

Frank-III (commented):
The core logic is correct, and I know you probably use LLM for the lint, but could you take a detailed look at other lints before submitting it?

Also, you should notice that when testing `ImplOwner`, we usually test struct/enum/union, could you add tests for all of them?

src/lints/type_method_receiver_ref_now_mutable.ron
suaviloquence: See docs for adding a witness here. https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md#hint-templates

Does this help clarify what this field should be? I'd love to improve the docs if they're unclear.

src/lints/type_method_receiver_ref_now_mutable.ron
KhaledSayed04: But the external files/modules can't call the private methods so won't be effected, right?

src/lints/type_method_receiver_ref_now_mutable.ron
KhaledSayed04: is this good?
witness: (
        hint_template: r#"let x: {{join "::" path}} = ...; x.{{ method_name }}(...);"#
    )

KhaledSayed04:
>and I know you probably use LLM for the lint

For the written code no. I may chat with it to brainstorm. 
I read the note mentioned at the contributing guide "We have a zero-tolerance policy for low-quality AI-generated contributions.", and sure I don't want to be banned. 
I'm just striving to engage fast to meet GSoC application deadline.

src/lints/type_method_receiver_ref_now_mutable.ron
suaviloquence: I'm not familiar enough with the receiver schema that Frank added to answer your second question, but in my opinion, something that only has an `&T` (so cant call methods on `&mut T`) may be clearer as a witness:

```rs
fn downstream(x: &path::to::Struct) {
    x.method(...);
}
```

because it shows greater breakage than "you just have to add `mut` to the `let`"

PR #1249, opened 2025-04-14
Chore: Add witness hint for `struct_missing` lint

GlitchlessCode:
Adds a witness hint for the `struct_missing` lint.

## Example
`struct_missing` outputs the following witness hint for one of the `./test_crates/struct_missing/` tests.
```rust
use struct_missing::WillBeRemovedStruct;
```
This compiles on the old version, but not on the new as importing from a non-existant path is not possible.

## Commit Notes
+ Added witness hint for struct_missing

## Comments
I figured just using a `use` statement would be enough to get across that the breaking change is the now missing path. Do you think this would work fine, or would it be preferable to use a different form of hint?


obi1kenobi:
Oh, unfortunately there's already an open PR (#1206) with some discussion on it that we've just failed to merge in time, in part due to me being overwhelmed in the last few weeks.

Sorry about that!

GlitchlessCode:
Oh, whoops! I thought I checked but I guess not thoroughly enough. My apologizes. PR Closed!

PR #1257, opened 2025-04-22
Drop support for Rust 1.84. New MSRV is 1.85.

obi1kenobi:


.github/workflows/ci.yml
```diff
@@ -1886,7 +1886,7 @@ jobs:
       - name: Check output
         run: |
           cd semver
-          EXPECTED="$(echo -e "error: rustc version is not high enough: >=1.84.0 needed, got 1.83.0")"
+          EXPECTED="$(echo -e "error: rustc version is not high enough: >=1.85.0 needed, got 1.84.0")"
```
obi1kenobi, line 1889:
```suggestion
          EXPECTED="$(echo -e "error: rustc version is not high enough: >=1.85.0 needed, got 1.84.1")"
```

PR #1269, opened 2025-05-16
Enable environment variable overrides for git repo discovery

PigeonF:
This enables overriding the git repository with the `GIT_DIR` environment variable. I was looking for this feature when working on a repository cloned with [jujutsu][], though I am sure there are other workflows for which overriding the git repository path can be useful.

[jujutsu]: https://github.com/jj-vcs/jj

This enables use with [jujutsu][] without a colocated repository.

```console
GIT_DIR="$(jj root)/.jj/repo/store/git" cargo semver-checks --package "foo" --baseline-rev "foo/v0.1.0"
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1473,6 +1473,96 @@ jobs:
           key: ${{ runner.os }}-${{ steps.toolchain.outputs.cachekey }}-${{ hashFiles('semver/**/Cargo.lock') }}-${{ github.job }}-rustdoc
           path: subject-current/target/semver-checks/cache
 
+  run-on-bare-repository:
+    # Test that overwriting the git repository detection with `GIT_DIR` works as expected by
+    # running on a bare clone of a repository.
+    name: "Semver: bare repository"
+    runs-on: ubuntu-latest
+    needs:
+      - build-binary
+    steps:
+      - name: Checkout cargo-semver-checks
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: true
+          path: "semver"
+
+      - name: Checkout ref-slice fork major change
+        uses: actions/checkout@v4
+        with:
+          # NOTE: We cannot checkout a bare repository with the actions/checkout action
+          # (https://github.com/actions/checkout/issues/603), so we checkout a "normal" version
+          # and clone the local checkout into a bare repository.
+          fetch-depth: 0
+          persist-credentials: false
+          repository: "mgr0dzicki/cargo-semver-action-ref-slice"
+          ref: "major_change"
+          path: "subject-checkout"
+
+      - name: Setup bare repository worktree
+        run: |
+          set -euo pipefail
+          git clone --bare subject-checkout subject-bare
+          mkdir subject
+          git --git-dir subject-bare --work-tree subject checkout major_change .
+          test ! -e subject/.git
+
+      - name: Install rust
+        id: toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          cache: false
+          rustflags: ""
+
+      - name: Restore binary
+        id: cache-binary
+        uses: actions/cache/restore@v4
+        with:
+          path: bins/
+          key: bins-${{ runner.os }}-${{ github.run_id }}-${{ github.run_attempt }}
+          fail-on-cache-miss: true
+
+      - name: Restore rustdoc
+        id: cache
+        uses: actions/cache/restore@v4
+        with:
+          key: ${{ runner.os }}-${{ steps.toolchain.outputs.cachekey }}-${{ hashFiles('semver/**/Cargo.lock') }}-${{ github.job }}-rustdoc
+          path: subject/target/semver-checks/cache
+
+      - name: Restore cargo index and rustdoc target dir
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: |
+            subject/target/semver-checks/git-v1_2_1/
+            subject/target/semver-checks/local-ref_slice-1_2_1-01666ec060466c14/
+
+      - name: Run semver-checks with --baseline-rev
+        continue-on-error: true
+        run: |
+          cd semver
+          set -euo pipefail
+          GIT_DIR=../subject-bare ../bins/cargo-semver-checks semver-checks check-release --manifest-path="../subject/Cargo.toml" --baseline-rev=v1.2.1 --verbose | tee output
+          touch unexpectedly_did_not_fail
+
+      - name: Check whether it failed
+        run: |
+          cd semver
+          if [ -f unexpectedly_did_not_fail ]; then exit 1; else exit 0; fi
+
+      - name: Check output
+        run: |
+          cd semver
+          EXPECTED="$(echo -e "--- failure function_missing: pub fn removed or renamed ---")"
+          RESULT="$(grep failure output | sort)"
+          diff <(echo "$RESULT") <(echo "$EXPECTED")
+
+      - name: Save rustdoc
+        uses: actions/cache/save@v4
+        if: steps.cache.outputs.cache-hit != 'true'
+        with:
+          key: ${{ runner.os }}-${{ steps.toolchain.outputs.cachekey }}-${{ hashFiles('semver/**/Cargo.lock') }}-${{ github.job }}-rustdoc
+          path: subject/target/semver-checks/cache
+
   run-on-ref-slice-fork:
     # Simulate testing an as-yet-unreleased crate version without an explicit baseline,
     # i.e. with the expectation that the largest crates.io version is used as a baseline.
--- a/README.md
+++ b/README.md
@@ -127,6 +127,38 @@ Custom registries are not currently supported
 ([#160](https://github.com/obi1kenobi/cargo-semver-checks/issues/160)), so crates published on
 registries other than crates.io should use one of the other approaches of generating the baseline.
 
+#### Git repository detection
+
+When looking up a git revision with `--baseline-rev`,
+`cargo-semver-checks` will walk up the current directory until it finds a `.git/` directory to resolve the revision and extract the corresponding worktree.
+You can use the following environment variables to influence the git repository detection:
+
+- [`GIT_CEILING_DIRECTORIES`]: a colon (`:`) separated list of absolute paths which `cargo-semver-checks` should not search for the `.git/` directory.
+  This can be used to prevent `cargo-semver-checks` from searching slow network mounted directories.
+  This environment variable cannot be used to prevent searching the current working directory,
+  or a directory explicitly set with [`GIT_DIR`].
+  If this is not set,
+  `cargo-semver-checks` will search the current working directory and any of its parents (up to the root directory).
+- [`GIT_DIR`]: explicitly set the location of the `.git/` directory
+  (if the value is a relative path, it is resolved from the current working directory).
+  If this is not set,
+  `cargo-semver-checks` will search the current working directory and any of its parents for the `.git/` directory.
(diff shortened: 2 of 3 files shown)


obi1kenobi:
Apologies for the delay, I've been working through the backlog of issues and emails that built up during my trip around RustWeek. Thanks for opening the PR!

I'm afraid I haven't had time to look into jujutsu yet, so I don't quite understand the motivation for this feature. If it's generally useful for both jujutsu and other use cases, we should probably not only support it but also call it out in the README. I'm not opposed to adding the feature — I just want to understand what we're promising to support :)

Would you mind adding a section to the README that appropriately describes the motivation and proper use of the new feature?

A test that makes sure this functionality continues to work would also be lovely.

PigeonF:
No worries about the delayed response :)

Jujutsu stores it's data in `.jj` instead of `.git`. Part of that data is basically an equivalent of the `.git` directory (in `.jj/repo/store/git`). This directory can be read by git no problem, the only issue is getting git to read this directory instead of the default `.git` directory. One way to overwrite the path git looks at is the [`GIT_DIR`](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#_repository_locations) environment variable - this PR enables the same behaviour for `gix`.

Basically, this change keeps the old git repository detection logic the same, but allows changing the behaviour with [some environment variables](https://docs.rs/gix/latest/gix/struct.ThreadSafeRepository.html#method.discover_with_environment_overrides_opts):

- `GIT_DIR` to override the logic of walking up the parent directories until a `.git` directory is found
- `GIT_DISCOVERY_ACROSS_FILESYSTEM` to enable/disable discovery across file systems (this currently defaults to `false`, so this would enable support for some setups where part of the repository are in different file systems)
- `GIT_CEILING_DIRECTORIES` to stop discovering git directories after a certain level

There are probably reasons other than integrating with jujutsu that overriding the `GIT_DIR` enables (such as [bare repositories](https://www.saintsjd.com/2011/01/what-is-a-bare-git-repository/)), but since I am not using them, I cannot comment further 😅

I am open to adding a section to the readme, as well as adding tests. I tried looking through the tests to see if there were any existing tests for the [`RustdocFromGitRevision`](https://github.com/obi1kenobi/cargo-semver-checks/blob/e6bb9c550a5a563269fb891b1d584a67354f5252/src/rustdoc_gen.rs#L469-L494), but could not find any. Did I miss them somewhere, or should I add them from scratch? If so, do you have any thoughts on how to test them? The "most correct" way is probably to clone a fresh checkout of this repository, but that would add a test-time dependency on git. Testing with the currently-checked-out repository is possible, but I don't see how to non-destructively test that overriding `GIT_DIR` works (since a proper test should check that discovery works at a location other than `.git`, which would involve changing the current checkout).

P.S.: If you ever want to take a closer look at jujutsu, I found [steve klabnik's tutorial](https://steveklabnik.github.io/jujutsu-tutorial/) very informative

obi1kenobi:
Thanks, that was super useful context! I'm onboard with adding this.

For testing, we test `--baseline-rev` via GitHub Actions jobs, not in `cargo test`, since that proved to be the easiest way to get it tested in a fashion that accurately represents real-world use.

Our GitHub Action repo [has some tests using a small custom repo](https://github.com/obi1kenobi/cargo-semver-checks-action/blob/de90d0688b68824ce9d50523ae0a4066961d8d0b/.github/workflows/test-action.yml#L32-L56) that we've prepared to have both minor and major breaking changes. It might be reasonable to add a new GitHub Actions job that clones that repo, moves its `.git` directory to `.jj/...` to simulate `jj` use, and then attempts to use `cargo-semver-checks` with `GIT_DIR`. Probably best to test the case where a major breaking change is found: check for non-zero exit and grep for the expected-to-fail lint name in the output to ensure it wasn't something else that broke.

What do you think?

PigeonF:
Sounds good! I will probably get around to this tomorrow 👍

obi1kenobi:
Awesome! No rush & thank you!

.github/workflows/ci.yml
```diff
@@ -1473,6 +1473,95 @@ jobs:
...
+        with:
+          key: ${{ runner.os }}-${{ steps.toolchain.outputs.cachekey }}-${{ hashFiles('semver/**/Cargo.lock') }}-${{ github.job }}-rustdoc
+          path: subject/target/semver-checks/cache
+
+      - name: Restore cargo index and rustdoc target dir
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: |
+            subject/target/semver-checks/git-master/
+            subject/target/semver-checks/local-ref_slice-1_2_1-01666ec060466c14
```
PigeonF: I have copied this path from my local paths, but I am not sure where the hash comes from

PigeonF:
I have added a test for `GIT_DIR`, and added a section to the readme. Let me know if I should change something :)

.github/workflows/ci.yml
obi1kenobi: It's a hash of the enabled features of the crate, to ensure we don't accidentally use a rustdoc JSON that has different features and therefore a different public API.

README.md
```diff
@@ -127,6 +127,26 @@ Custom registries are not currently supported
 ([#160](https://github.com/obi1kenobi/cargo-semver-checks/issues/160)), so crates published on
 registries other than crates.io should use one of the other approaches of generating the baseline.
 
+#### Git repository detection
+
+When looking up a git revision with `--baseline-rev`,
+`cargo-semver-checks` will walk up the current directory until it finds a `.git/` directory to resolve the revision and extract the corresponding worktree.
+You can use the following environment variables to influence the git repository detection
```
obi1kenobi:
```suggestion
You can use the following environment variables to influence the git repository detection:
```

README.md
```diff
@@ -127,6 +127,26 @@ Custom registries are not currently supported
...
+
+- [`GIT_CEILING_DIRECTORIES`] to stop automatic git repository discovery when reaching certain directories.
```
obi1kenobi: The "certain directories" phrase is a bit confusing. Could you offer a bit of intuition to the reader as to when they might use this?

Also, what's the default value / behavior here?

README.md
```diff
@@ -127,6 +127,26 @@ Custom registries are not currently supported
...
+- [`GIT_DIR`] to explicitly set the location of the `.git/` directory.
```
obi1kenobi: Does this have to be an absolute path, or would a relative path do? If relative, is it relative to the `pwd` or something else?

Also, what's the default value / behavior here?

obi1kenobi (commented):
The test and implementation look great! I'd love to get just a touch more polish on the docs and it'll be ready to merge and ship 🚀

PigeonF:
I have updated the documentation to be more explicit; the previous version was more terse, because this basically duplicates what the linked git documentation says. Let me know if the new wording is better than the old one :)

P.S.: I force-pushed the changes, if you prefer new commits for review instead, let me know!

README.md
```diff
@@ -127,6 +127,38 @@ Custom registries are not currently supported
...
+- [`GIT_CEILING_DIRECTORIES`]: a colon (`:`) separated list of absolute paths which `cargo-semver-checks` should not search for the `.git/` directory.
+  This can be used to prevent `cargo-semver-checks` from searching slow network mounted directories.
+  This environment variable cannot be used to prevent searching the current working directory,
+  or a directory explicitly set with [`GIT_DIR`].
+  If this is not set,
+  `cargo-semver-checks` will search the current working directory and any of its parents (up to the root directory).
+- [`GIT_DIR`]: explicitly set the location of the `.git/` directory
+  (if the value is a relative path, it is resolved from the current working directory).
+  If this is not set,
+  `cargo-semver-checks` will search the current working directory and any of its parents for the `.git/` directory.
```
obi1kenobi, line 145: `GIT_DIR` seems like the option we'd expect to be most commonly used, right?

If so, can we put it in the top spot? It feels like describing `GIT_CEILING_DIRECTORIES` will flow better in the second spot anyway, because if `GIT_DIR` isn't set it's presumably _then_ that we kick off a search based on parameters that `GIT_CEILING_DIRECTORIES` influences. If so, let's document that setting `GIT_CEILING_DIRECTORIES` has no effect if `GIT_DIR` is set too.

README.md
```diff
@@ -127,6 +127,38 @@ Custom registries are not currently supported
...
+- [`GIT_DISCOVERY_ACROSS_FILESYSTEM`]: enable git repository detection across filesystems.
+  When automatically searching for the `.git/` directory,
+  `cargo-semver-checks` will not cross filesystem boundaries.
```
obi1kenobi:
```suggestion
  By default, `cargo-semver-checks` will not cross filesystem
  boundaries when searching for the `.git/` directory.
```

README.md
```diff
@@ -127,6 +127,38 @@ Custom registries are not currently supported
 ([#160](https://github.com/obi1kenobi/cargo-semver-checks/issues/160)), so crates published on
 registries other than crates.io should use one of the other approaches of generating the baseline.
 
+#### Git repository detection
```
obi1kenobi, line 130: Imagine you came here to figure out how to use `cargo-semver-checks` with jujutsu. `Git repository detection` is probably not the section title where you'd expect to find the answer.

Consider adding a
```
#### Use with `jujutsu`
```
subsection? Also consider adding a `jujutsu` entry to the FAQ?

README.md
```diff
@@ -127,6 +127,38 @@ Custom registries are not currently supported
...
+
+[`GIT_CEILING_DIRECTORIES`]: https://git-scm.com/docs/git.html#Documentation/git.txt-codeGITCEILINGDIRECTORIEScode
+[`GIT_DIR`]: https://git-scm.com/docs/git.html#Documentation/git.txt-codeGITDIRcode
+[`GIT_DISCOVERY_ACROSS_FILESYSTEM`]: https://git-scm.com/docs/git.html#Documentation/git.txt-codeGITDISCOVERYACROSSFILESYSTEMcode
+
+Setting the `GIT_DIR` allows using `cargo-semver-checks` with a non-colocated [jujutsu](https://jj-vcs.github.io/jj/latest/) repository.
+
+```console
+GIT_DIR="$(jj root)/.jj/repo/store/git" cargo semver-checks --package "foo" --baseline-rev "foo/v0.1.0"
+```
```
obi1kenobi:
```suggestion
#### Use with jujutsu

Setting the `GIT_DIR` allows using `cargo-semver-checks` with a non-colocated [jujutsu](https://jj-vcs.github.io/jj/latest/) repository.

```console
GIT_DIR="$(jj root)/.jj/repo/store/git" cargo semver-checks --package "foo" --baseline-rev "foo/v0.1.0"
```
```

obi1kenobi (commented):
We squash-merge all PRs into a single commit to make any required reverts easy, so rebasing / force-pushing has limited value. GitHub's review flow isn't friendly to history-rewrites since it loses past comment state, so it's a bit easier for me to review if you just add new commits with tweaks.

Anyway, a couple of micro-nitpicks and tiny suggestions — hopefully all resolvable within <5min. I appreciate all the work you've done here! If you're tired of wordsmithing, I don't mind applying the changes myself when I get the chance — just let me know.

PigeonF:
I don't mind updating the documentation again, but you have more experience and a better vision of what you want the changes to look like (especially for something as important and user-facing as the readme), so I would prefer if you do it. Writing good documentation is one of my weaker skills, so I will gladly take this as a learning experience on how you would document the changes from this PR :)

obi1kenobi (approved):
Thanks! I'll merge this PR and then iterate on it separately, just to keep things simpler. I'm not great at docs either — we're all just doing our best out here :)

PR #1276, opened 2025-06-01
[Witness Generation] Add unstable witness program flag

GlitchlessCode:
# Commit Notes
+ Added new unstable flag `witnesses`
+ Flag requires output path `PathBuf` for target witness output

# Comments
Feels good to get started! This just adds an unstable flag as was suggested on the rust-lang Zulip to gate the witness generation system. Decided to make the flag require an output directory, since that is mandatory for the witness system to work anyways. 

Nothing is done with the flag as of right now, it just exists.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -339,6 +339,10 @@ struct UnstableOptions {
     /// Enable printing witness hints, examples of potentially-broken downstream code.
     #[arg(long, hide = true)]
     witness_hints: bool,
+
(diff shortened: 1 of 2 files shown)


GlitchlessCode:
I forgot to update the integration snapshots, I'll go fix that :face_exhaling:

obi1kenobi (approved):
Sorry for the delay!

Issue #166, opened 2022-10-31
Crates published on a custom registry cannot be checked using default settings


obi1kenobi:
By default, `cargo semver-checks check-release` attempts to look up the most recent previous version of the crate on crates.io. This won't work for crates that are only published on a custom registry.

The README has [a decent workaround](https://github.com/obi1kenobi/cargo-semver-check#does-the-crate-im-checking-have-to-be-published-on-cratesio) for now, but it's still a workaround and it would be more ergonomic to natively support custom registries at some point.

j-baker:
Hi! I can't commit to working on this as of right now, but maybe to nudge progress along I'd like to lay out a proposal for a design here.
In particular, a simple intuition regarding ergonomic UX would be to match Cargo's behaviour in publish; that is to say, to implement approximately:

1. Accept a registry name with `--registry foo`. Look that registry up in Cargo configs, possibly with the cargo-config2 crate.
2. If not specified and there is a package.publish field in Cargo.toml with a single registry, publish to that registry.
3. Otherwise, use the default registry.

obi1kenobi:
I think there are a few more edge cases to tackle, but in broad strokes that makes sense.

It's my impression that nearly all use cases that require a custom registry are related to commercial activity. I'd love it if the companies publishing crates to private registries sponsored `cargo-semver-checks` development!

Right now, I have to be extremely aggressive about limiting the maintenance cost of this project. That means I can't afford to include features like this one that run a risk of causing a high maintenance and support burden ("why doesn't it work with my registry?") and risk burning me out in the process. I'd love to make this a self-sustaining project with a stable income, instead of a thing I do in spare time on nights and weekends.

If you work at a company that uses `cargo-semver-checks`, and especially one that wants to use it with a custom registry, please talk to them about [sponsoring its development](https://github.com/sponsors/obi1kenobi). If you aren't sure how to approach that conversation, or if your company is interested in any additional services, please give me a ping!

PR #1286, opened 2025-06-09
Support for custom registries

### LINKED ISSUE #160: Custom registry does not seem supported

kornelski:
Fixes #160 and #166

`tame-index` doesn't support authentication (https://github.com/EmbarkStudios/tame-index/issues/50), so I've added two workarounds for this:
 * an env var `CARGO_REGISTRIES_{name}_TOKEN` can be set (the same one as used by Cargo). This will make authenticated requests work directly, with good performance.
 * as a fallback, I run `cargo info --registry {name} {crate}` which makes Cargo fetch the crate and cache it, and then `tame-index` can fish it out of the Cargo cache. This is slower, but supports all the fancy Cargo credential helpers.

The loader abstraction assumed there was only a single `RustdocGenerator` for all baseline crates and eagerly created an index instance. It's possible to have a workspace that publishes to multiple different registries, so I've made generators instantiated lazily as needed.


obi1kenobi:
Hey @kornelski, thanks for putting this together!

Any chance you might be up for a chat over Zulip one of these days? It's clear this PR and #1288 weren't easy to put together, and I'd love to know more about your use case.

My primary concern with supporting custom registries is one of maintainability. `cargo-semver-checks` is always dancing on the absolute limit of maintainability, and custom registries support risks massively increasing the support burden — almost always from companies using `cargo-semver-checks` in a commercial context but not currently funding it.

If we can find a way to get companies that would benefit from custom registry support to give back to the project in a meaningful way, I'd be thrilled to merge this and support it going forward. But if we can't make it a win-win, then I think I'd rather spend the limited "maintainability tokens" on things that benefit the entire Rust community instead.

kornelski:
I'll see what I can do. In the meantime, #1289 is the groundwork that should _reduce_ maintenance burden of that code.

PR #1289, opened 2025-06-10
Collect crates before running reports

kornelski:
This refactoring doesn't alter functionality, but deduplicates report-generation code in `check_release`. 

Previously, two kinds of data sources would run almost-identical report generation code. Now the reports are built in one place.

Moving some code to a couple of helper functions make the check loop less spaghettified.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -429,7 +429,7 @@ impl Check {
             };
         }
 
-        let crates_to_check: Vec<CrateToCheck> = match &self.current.source {
+        let crates_to_check: Vec<CrateToCheck<'_>> = match &self.current.source {
             RustdocSource::Rustdoc(_)
             | RustdocSource::Revision(_, _)
             | RustdocSource::VersionFromRegistry(_) => {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -423,79 +429,43 @@ impl Check {
...
-        let current_loader = self.get_rustdoc_generator(config, &self.current.source)?;
-        let baseline_loader = self.get_rustdoc_generator(config, &self.baseline.source)?;
-
-        // Create a report for each crate.
-        // We want to run all the checks, even if one returns `Err`.
-        let all_outcomes: Vec<anyhow::Result<(String, Option<CrateReport>)>> = match &self
-            .current
-            .source
-        {
+        let crates_to_check: Vec<CrateToCheck> = match &self.current.source {
```
obi1kenobi: I find omitted lifetimes to be a bit of a footgun, so I'd love to acknowledge them even with just `'_`:
```suggestion
        let crates_to_check: Vec<CrateToCheck<'_>> = match &self.current.source {
```

obi1kenobi (approved):
Thank you!

PR #1290, opened 2025-06-10
Don't change current_dir when running cargo

kornelski:
Cargo has a quirk that search for `.cargo/config.toml` is relative to the current directory Cargo has been launched from, which may end up being completely unrelated to the workspace's or crate's directory.

Annoyingly, this means that `cargo build` and `cargo build --manifest-path=some/dir/elsewhere/Cargo.toml` can have a significantly different behavior.

Some projects take advantage of that, e.g. Linux distros may want to override build scripts without modifying the source tree, so they'll build from another dir with the desired config. This also can go the other way, and a project may require `rustflags` set in its `.cargo/config.toml` to build properly.

So it's safest to keep the current dir as-is. `cargo-semver-checks` mostly does this correctly, except this one.

### DIFF
--- a/src/data_generation/generate.rs
+++ b/src/data_generation/generate.rs
@@ -84,7 +84,6 @@ pub(super) fn generate_rustdoc(
             crate_name,
             version,
             request,
-            build_dir,
             placeholder_manifest_path.as_path(),
             &settings,
         ) {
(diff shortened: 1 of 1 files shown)


obi1kenobi:
Thanks!

Issue #1262, opened 2025-05-09
Allow skipping version metadata when specifying `--baseline-version`


maia-s:
### Steps to reproduce the bug with the above code

If a version of a crate has version metadata, that metadata has to be provided to `--baseline-version`, or it will error. For example, this fails:
```
> DOCS_RS=1 cargo semver-checks --baseline-version 0.4.7
    Building sdl3-sys v0.4.9+SDL3-3.2.10 (current)
       Built [   1.106s] (current)
     Parsing sdl3-sys v0.4.9+SDL3-3.2.10 (current)
      Parsed [   0.032s] (current)
error: Version sdl3-sys of crate 0.4.7 not found in registry
```
(also, the crate name and version are swapped in the error message)

But this succeeds:
```
> DOCS_RS=1 cargo semver-checks --baseline-version 0.4.7+SDL3-3.2.10
    Building sdl3-sys v0.4.9+SDL3-3.2.10 (current)
       Built [   1.431s] (current)
     Parsing sdl3-sys v0.4.9+SDL3-3.2.10 (current)
      Parsed [   0.033s] (current)
     Parsing sdl3-sys v0.4.7+SDL3-3.2.10 (baseline, cached)
      Parsed [   0.040s] (baseline)
    Checking sdl3-sys v0.4.7+SDL3-3.2.10 -> v0.4.9+SDL3-3.2.10 (minor change)
     Checked [   0.148s] 128 checks: 128 pass, 36 skip
     Summary no semver update required
    Finished [   2.160s] sdl3-sys
```

### Actual Behaviour

Version metadata is required

### Expected Behaviour

Version metadata is ignored by cargo, so `cargo-semver-checks` shouldn't require it.

### Generated System Information

System information:
-------------------
#### Software version

cargo-semver-checks 0.41.0 (793f25c)

#### Operating system

- OS: macOS 15.4.1 Sequoia
- Kernel: 24.4.0

#### Command-line

```bash
/Users/maia/.cargo/bin/cargo-semver-checks semver-checks --bugreport
```

#### cargo version

```
> cargo -V
cargo 1.88.0-nightly (d811228b1 2025-04-15)
```

#### Compile time information

- Profile: release
- Target triple: aarch64-apple-darwin
- Family: unix
- OS: macos
- Architecture: aarch64
- Pointer width: 64
- Endian: little
[...]

PR #1293, opened 2025-06-13
chore: change version comparison logic in RustdocGenerator

Frank-III:
close: #1262


obi1kenobi (commented):
The change looks great!

Would you be able to add a test for it? For example, a GitHub Actions workflow that checks out the project that found this issue, reproduces the original invocation that used to fail, and ensures it succeeds? There should be a few examples of similar uses of other projects that you should be able to follow.

obi1kenobi:
I noticed the test is being stubborn and hard to get working. If it's not coming together, feel free to ping the OP in the related issue — I'm sure they'd know what to do.

PR #1330, opened 2025-06-18
[Witness Generation]  Add `InheritedValue` inheritance

GlitchlessCode:
# Commit Notes
+ Adds the `InheritOutput` variant to `InheritedValue`
+ Adds a function to inherit prior values for witness generation
# Comments
Technically this does the inheritance for both arguments and outputs, which is needed for witness querying systems. I'm not happy with the cloning I put, that was just to get something in and working. Also, I'd prefer to use an `Iterator` instead of a for loop (allows use of `BTreeMap::from_iter` which would be nice), just makes it nicer to read in my mind. Generally just want some feedback on this, but overall I'm pretty happy with this as at least a first iteration of the value inheritance.

Going to put this as a draft PR at least for now, since it doesn't really *need* to be merged right away and I'd like to improve it. This is just getting something hammered out. If you want to merge this anyways, just so we have something, that also works.

Also, not entirely happy with all my variable names, function names, etc. feel free to modify/suggest away.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -2,7 +2,7 @@ use std::collections::BTreeMap;
 
 use ron::extensions::Extensions;
 use serde::{Deserialize, Serialize};
-use trustfall::TransparentValue;
+use trustfall::{FieldValue, TransparentValue};
 
 use crate::ReleaseType;
 
@@ -313,6 +313,49 @@ pub struct WitnessQuery {
     pub arguments: BTreeMap<String, InheritedValue>,
 }
 
+impl WitnessQuery {
+    /// Returns [`arguments`](Self::arguments), mapping the [`InheritedValue`]s from
+    /// the given argument and output maps. The argument map is the map of arguments
+    /// used in the previous query. The output map is the map of output [`FieldValue`]s
+    /// from the previous query.
+    ///
+    /// Fails with an [`anyhow::Error`] if the requested inheritance keys is missing.
+    pub fn inherit_arguments_from(
+        self,
+        argument_map: &BTreeMap<String, TransparentValue>,
+        output_map: &BTreeMap<std::sync::Arc<str>, FieldValue>,
+    ) -> anyhow::Result<BTreeMap<String, TransparentValue>> {
+        let mut mapped = BTreeMap::new();
+
+        // Ideally this should be an iterator method chain, not a for loop
+        // but this works for now
+        for (key, value) in self.arguments {
+            let mapped_value = match value {
+                // Inherit an argument
+                InheritedValue::Inherited { inherit } => argument_map
+                    .get(&inherit)
+                    .map(Clone::clone) // TODO: Remove this clone if possible
+                    .ok_or(anyhow::anyhow!(
+                        "inherited argument key `{inherit}` does not exist"
+                    ))?,
+                // Inherit an output
+                InheritedValue::InheritOutput { from } => output_map
+                    .get(from.as_str())
+                    .map(Clone::clone) // TODO: Remove this clone if possible
+                    .map(Into::into)
+                    .ok_or(anyhow::anyhow!(
+                        "inherited output key `{from}` does not exist"
+                    ))?,
+                // Set a constant
+                InheritedValue::Constant(value) => value,
+            };
+            mapped.insert(key, mapped_value);
+        }
+
+        Ok(mapped)
+    }
+}
+
 /// Represents either a value inherited from a previous query, or a
 /// provided constant value.
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


GlitchlessCode:
Also I would like to add unit tests for the new system eventually, currently just added the system.

GlitchlessCode:
Also glad to see clippy found my clone as well. I definitely will be reworking that eventually. As I said, it's just something that works right now.

GlitchlessCode:
One other thought I just had:
The method takes `self` rather than `&self`. I originally had it taking `&self` but then ran into other problems and that was the easiest solution. Realizing now however, that this breaks the use of the `WitnessGeneration` struct elsewhere since it consumes itself. That 100% needs to change back to taking `&self`. That's why we draft things I guess :P

src/query.rs
```diff
@@ -313,13 +313,58 @@ pub struct WitnessQuery {
     pub arguments: BTreeMap<String, InheritedValue>,
 }
 
+impl WitnessQuery {
+    /// Returns [`arguments`](Self::arguments), mapping the [`InheritedValue`]s from
+    /// the given argument and output maps. The argument map is the map of arguments
+    /// used in the previous query. The output map is the map of output [`FieldValue`]s
+    /// from the previous query.
+    ///
+    /// Fails with an [`anyhow::Error`] if the requested inheritance keys is missing.
```
obi1kenobi:
```suggestion
    /// Fails with an [`anyhow::Error`] if the requested inheritance keys are missing.
```

src/query.rs
```diff
@@ -313,13 +313,58 @@ pub struct WitnessQuery {
...
+    pub fn inherit_arguments_from(
+        self,
+        argument_map: &BTreeMap<String, TransparentValue>,
+        output_map: &BTreeMap<std::sync::Arc<str>, FieldValue>,
+    ) -> anyhow::Result<BTreeMap<String, TransparentValue>> {
+        let mut mapped = BTreeMap::new();
+
+        // Ideally this should be an iterator method chain, not a for loop
+        // but this works for now
```
obi1kenobi: FWIW I don't think an iterator vs for-loop makes any difference for us. We'd fully exhaust the iterator immediately to construct the map, so there's no real advantage one way or the other.

I'd say keep an eye on the big picture: high-quality witnesses delivered to our users as soon as we can. The users will neither care about nor notice the difference between the iterator and the for loop here :)

src/query.rs
```diff
@@ -313,13 +313,58 @@ pub struct WitnessQuery {
...
+        for (key, value) in self.arguments {
+            let mapped_value = match value {
+                // Inherit an argument
+                InheritedValue::Inherited { inherit } => argument_map
+                    .get(&inherit)
+                    .map(Clone::clone) // TODO: Remove this clone if possible
```
obi1kenobi: Don't worry about cloning in code paths around witnesses.

Zooming out: we just ran a very expensive lint query over hundreds of MB of data, and it found problems to look into. Our users will be very happy to spend an extra sub-millisecond cloning the `String` if it means they avoid SemVer breakage.

src/query.rs
```diff
@@ -313,13 +313,58 @@ pub struct WitnessQuery {
...
+
 /// Represents either a value inherited from a previous query, or a
 /// provided constant value.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(untagged, deny_unknown_fields)]
 pub enum InheritedValue {
     /// Inherit the value from the previous output whose name is the given `String`.
     Inherited { inherit: String },
+    /// Inherit the value from the previous lint's output by whichever @output name is the given `String`.
+    InheritOutput { from: String },
```
obi1kenobi: I'm a bit confused about how these two variants are different. They both seem to inherit from an `@output` in the lint query with the contained name.

GlitchlessCode: Huh I could have sworn that previously said something about inheriting arguments not outputs. Basically the theory was to have one for inheriting from the original map of arguments, the other from the map of outputs. Honestly I can just simplify it down to the one, I don't really see a case where inheriting an argument would be necessary? Mostly though, yeah I agree with you, idk what I was doing there 😅

GlitchlessCode:
Apologies for the delay to get that in, had some real life stuff come up, but that's taken care of

obi1kenobi:
This looks good to me, I'm happy to merge it — I'll let you move it out of "draft" state first, in case you wanted to iterate on it some more first. The failing CI check is unrelated, seems like a blip in the APIs we use.

GlitchlessCode:
I think it's in a good state now or at least good enough. Moving out of draft

obi1kenobi:
Also ignore nightly for a bit. I need to add support for a new rustdoc JSON format, and it'll keep failing until I get that in.

PR #1342, opened 2025-06-25
[Witness Generation] Add witness query system & witness text generation

GlitchlessCode:
# Commit Notes
+ Add new `witness_gen` module
+ Add logic to run witness queries
+ Witness query data is unused as of right now

# Comments
This took longer than I thought it would. Not the programming itself but life has not been on the side of my time lately.

Either way, this adds a new `witness_gen` module which I plan to use for a lot more of the witness generation system. Here, I just wanted it to separate things out a little bit. Currently, I'm not doing anything with the query result, and in fact, I haven't had the opportunity to test if it even does *anything* but I'll try to quickly test that next, I'd like to know it's working before going on to generating the actual witnesses from this data. Also, witness queries that fail for some reason or another currently keep their own errors, I'd rather design the system to allow as many witnesses to be generated as possible, and then simply report error/gather diagnostics for any query or witness that fails. Unlike the lint queries themselves, it arguably should not fail hard, and should allow the rest of `c-s-c` to complete, it should just report any errors that occur to the user, and prompt them to report it as an issue. At least, that's my opinion on it, thoughts?

Additionally, as I state in one of the comments, if no witness query exists, we ideally want to forward existing lint_results, so those values can still be used in a witness that may not need a query. Though looking at this now, it might be optimal to add another case to the `match` for specifically where there *is* a witness to generate and *not* a witness query present to do that, and then on the catchall, instead just get filtered out (switching to filter_map instead of just map would allow this). However, this does still work, it's just a possible improvement for later.

### DIFF
--- a/src/check_release.rs
+++ b/src/check_release.rs
@@ -8,10 +8,9 @@ use anyhow::Context;
 use clap::crate_version;
 use itertools::Itertools;
 use rayon::prelude::*;
-use trustfall::TransparentValue;
+use trustfall::{FieldValue, TransparentValue};
 
 use crate::data_generation::DataStorage;
-use crate::query::QueryResults;
 use crate::witness_gen;
 use crate::{
     query::{ActualSemverUpdate, LintLevel, OverrideStack, RequiredSemverUpdate, SemverQuery},
@@ -121,7 +120,7 @@ fn get_minimum_version_change(version: &semver::Version) -> VersionChange {
 fn print_triggered_lint(
     config: &mut GlobalConfig,
     semver_query: &SemverQuery,
-    results: Vec<QueryResults>,
+    results: Vec<BTreeMap<Arc<str>, FieldValue>>,
     witness_generation: &WitnessGeneration,
 ) -> anyhow::Result<()> {
     if let Some(ref_link) = semver_query.reference_link.as_deref() {
--- a/src/check_release.rs
+++ b/src/check_release.rs
@@ -9,6 +9,7 @@ use clap::crate_version;
 use itertools::Itertools;
 use rayon::prelude::*;
 use trustfall::{FieldValue, TransparentValue};
+use trustfall_rustdoc::VersionedRustdocAdapter;
 
 use crate::data_generation::DataStorage;
 use crate::witness_gen;
@@ -230,6 +231,46 @@ fn print_triggered_lint(
     Ok(())
 }
 
+/// Helper function to map a given semver query, and its witness to a value for use in [`run_check_release`] to
+/// to an [`Option`].
+fn map_witness_queries<'query>(
+    semver_query: &'query SemverQuery,
+    lint_results: &Vec<BTreeMap<Arc<str>, FieldValue>>,
+    adapter: &VersionedRustdocAdapter,
+) -> Option<(
+    &'query SemverQuery,
+    Vec<anyhow::Result<BTreeMap<Arc<str>, FieldValue>>>,
+)> {
+    match semver_query.witness {
+        // Don't bother running the witness query unless both a witness query and template exist
+        Some(Witness {
+            witness_template: Some(_),
+            witness_query: Some(ref witness_query),
+            ..
+        }) => {
+            let witness_results = lint_results
+                .iter()
+                .cloned()
+                .map(|lint_result| {
+                    witness_gen::run_witness_query(&adapter, witness_query, lint_result)
+                })
+                .collect_vec();
+            Some((semver_query, witness_results))
+        }
+
+        // If no witness query exists, we still want to forward the existing output
+        Some(Witness {
+            witness_template: Some(_),
+            witness_query: None,
+            ..
+        }) => Some((
+            semver_query,
+            lint_results.iter().cloned().map(Result::Ok).collect_vec(),
+        )),
+        _ => None,
+    }
+}
+
 pub(super) fn run_check_release(
     config: &mut GlobalConfig,
     data_storage: &DataStorage,
@@ -345,31 +386,12 @@ pub(super) fn run_check_release(
         .collect::<anyhow::Result<Vec<_>>>()?;
 
     if let Some(ref _witness_dir) = witness_generation.witness_directory {
-        let _witness_results = all_results
-            .par_iter()
-            .map(
-                |(semver_query, _, lint_results)| match semver_query.witness {
-                    // Don't bother running the witness query unless both a witness query and template exist
-                    Some(Witness {
-                        witness_template: Some(_),
-                        witness_query: Some(ref witness_query),
-                        ..
-                    }) => (
-                        semver_query,
-                        witness_gen::run_witness_queries(&adapter, witness_query, lint_results),
-                    ),
-                    // If no witness query exists, we still want to forward the existing output
-                    _ => (
-                        semver_query,
-                        lint_results
-                            .clone()
-                            .into_iter()
-                            .map(Result::Ok)
-                            .collect_vec(),
-                    ),
-                },
-            )
-            .collect::<Vec<_>>();
+        let _witness_results =
+            all_results
+                .par_iter()
+                .filter_map(|(semver_query, _, lint_results)| {
+                    map_witness_queries(&semver_query, lint_results, &adapter)
+                });
     }
 
     let mut results_with_errors = vec![];
--- a/src/check_release.rs
+++ b/src/check_release.rs
@@ -11,6 +11,8 @@ use rayon::prelude::*;
 use trustfall::{FieldValue, TransparentValue};
 
 use crate::data_generation::DataStorage;
+use crate::witness_gen;
+use crate::Witness;
 use crate::{
     query::{ActualSemverUpdate, LintLevel, OverrideStack, RequiredSemverUpdate, SemverQuery},
     CrateReport, GlobalConfig, ReleaseType, WitnessGeneration,
@@ -341,6 +343,34 @@ pub(super) fn run_check_release(
             Ok((semver_query, time_to_decide, results))
         })
         .collect::<anyhow::Result<Vec<_>>>()?;
+
+    if let Some(ref _witness_dir) = witness_generation.witness_directory {
+        let _witness_results = all_results
+            .par_iter()
+            .map(
+                |(semver_query, _, lint_results)| match semver_query.witness {
+                    // Don't bother running the witness query unless both a witness query and template exist
+                    Some(Witness {
+                        witness_template: Some(_),
+                        witness_query: Some(ref witness_query),
+                        ..
+                    }) => (
+                        semver_query,
+                        witness_gen::run_witness_queries(&adapter, witness_query, lint_results),
+                    ),
+                    // If no witness query exists, we still want to forward the existing output
+                    _ => (
+                        semver_query,
+                        lint_results
+                            .clone()
+                            .into_iter()
+                            .map(Result::Ok)
+                            .collect_vec(),
+                    ),
+                },
+            )
+            .collect::<Vec<_>>();
+    }
 
     let mut results_with_errors = vec![];
     let mut results_with_warnings = vec![];
--- a/src/check_release.rs
+++ b/src/check_release.rs
@@ -8,12 +8,11 @@ use anyhow::Context;
 use clap::crate_version;
 use itertools::Itertools;
 use rayon::prelude::*;
-use trustfall::{FieldValue, TransparentValue};
-use trustfall_rustdoc::VersionedRustdocAdapter;
+use trustfall::TransparentValue;
 
 use crate::data_generation::DataStorage;
+use crate::query::QueryResults;
 use crate::witness_gen;
-use crate::Witness;
 use crate::{
     query::{ActualSemverUpdate, LintLevel, OverrideStack, RequiredSemverUpdate, SemverQuery},
     CrateReport, GlobalConfig, ReleaseType, WitnessGeneration,
@@ -122,7 +121,7 @@ fn get_minimum_version_change(version: &semver::Version) -> VersionChange {
 fn print_triggered_lint(
     config: &mut GlobalConfig,
     semver_query: &SemverQuery,
-    results: Vec<BTreeMap<Arc<str>, FieldValue>>,
+    results: Vec<QueryResults>,
     witness_generation: &WitnessGeneration,
 ) -> anyhow::Result<()> {
     if let Some(ref_link) = semver_query.reference_link.as_deref() {
@@ -229,46 +228,6 @@ fn print_triggered_lint(
     }
 
     Ok(())
-}
-
-/// Helper function to map a given semver query, and its witness to a value for use in [`run_check_release`] to
-/// to an [`Option`].
-fn map_witness_queries<'query>(
-    semver_query: &'query SemverQuery,
-    lint_results: &Vec<BTreeMap<Arc<str>, FieldValue>>,
-    adapter: &VersionedRustdocAdapter,
-) -> Option<(
-    &'query SemverQuery,
-    Vec<anyhow::Result<BTreeMap<Arc<str>, FieldValue>>>,
-)> {
-    match semver_query.witness {
-        // Don't bother running the witness query unless both a witness query and template exist
-        Some(Witness {
-            witness_template: Some(_),
-            witness_query: Some(ref witness_query),
-            ..
-        }) => {
-            let witness_results = lint_results
-                .iter()
-                .cloned()
-                .map(|lint_result| {
-                    witness_gen::run_witness_query(&adapter, witness_query, lint_result)
-                })
-                .collect_vec();
-            Some((semver_query, witness_results))
-        }
-
-        // If no witness query exists, we still want to forward the existing output
-        Some(Witness {
-            witness_template: Some(_),
-            witness_query: None,
-            ..
-        }) => Some((
-            semver_query,
-            lint_results.iter().cloned().map(Result::Ok).collect_vec(),
-        )),
-        _ => None,
-    }
 }
 
 pub(super) fn run_check_release(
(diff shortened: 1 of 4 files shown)


GlitchlessCode:
Is it just me, or are the tests not getting the correct rustdoc?

src/witness_gen.rs
```diff
@@ -0,0 +1,48 @@
...
+/// will result in the result from the [`WitnessQuery`] silently overriding the same key from the
+/// [`SemverQuery`](crate::query::SemverQuery).
+pub(crate) fn run_witness_queries(
+    adapter: &VersionedRustdocAdapter,
+    witness_query: &WitnessQuery,
+    lint_results: &Vec<BTreeMap<Arc<str>, FieldValue>>,
+) -> Vec<anyhow::Result<BTreeMap<Arc<str>, FieldValue>>> {
+    lint_results
+        .clone()
+        .into_iter()
```
obi1kenobi: Consideer `.iter().cloned()` instead since it's more efficient: it only clones the elements without also needing to allocate a whole `Vec` which we will immediately discard.

src/witness_gen.rs
```diff
@@ -0,0 +1,48 @@
...
+        .map(|mut single_result| {
+            let arguments = witness_query.inherit_arguments_from(&single_result)?;
+
+            let witness_results = adapter
+                .run_query(&witness_query.query, arguments)
+                .and_then(|mut query_results| {
+                    let query_result = query_results.next();
+
+                    match query_results.next() {
```
obi1kenobi: In principle this should be fine, but it's probably a bit better to play it safe since it won't cost us anything.

The issue is that not all `impl Iterator` types allow calling `.next()` on them after one `.next()` has already returned `None`. The `Iterator` trait doesn't specify what happens — they might return `None` again, they might `panic!()`, they might return `Some(..)` etc.

An `impl FusedIterator` resolves this and guarantees that once one `.next()` call returns `None`, all subsequent `.next()` calls also do the same.

But here we can just as easily check if the first `.next()` was `None`, raise the error, and avoid the second call altogether.

src/witness_gen.rs
```diff
@@ -0,0 +1,48 @@
...
+/// [`anyhow::Error`] otherwise.
+///
+/// Overlapping output keys between the [`WitnessQuery`] and the [`SemverQuery`](crate::query::SemverQuery)
+/// will result in the result from the [`WitnessQuery`] silently overriding the same key from the
+/// [`SemverQuery`](crate::query::SemverQuery).
+pub(crate) fn run_witness_queries(
+    adapter: &VersionedRustdocAdapter,
+    witness_query: &WitnessQuery,
+    lint_results: &Vec<BTreeMap<Arc<str>, FieldValue>>,
+) -> Vec<anyhow::Result<BTreeMap<Arc<str>, FieldValue>>> {
```
obi1kenobi: Do we actually need this "`Vec` in `Vec` out" interface? I'm not sure what we gain by having this extra layer of abstraction, compared to just having a function that takes one lint result and produces the corresponding witness query output (or an error).

src/check_release.rs
```diff
@@ -342,6 +344,34 @@ pub(super) fn run_check_release(
...
+                    ),
+                    // If no witness query exists, we still want to forward the existing output
+                    _ => (
+                        semver_query,
+                        lint_results
+                            .clone()
+                            .into_iter()
+                            .map(Result::Ok)
+                            .collect_vec(),
+                    ),
```
obi1kenobi: I'm confused by this case. If there's no witness data, why are we cloning `lint_results`? I don't see what we'd do with it.

obi1kenobi (commented):
> Also, witness queries that fail for some reason or another currently keep their own errors, I'd rather design the system to allow as many witnesses to be generated as possible, and then simply report error/gather diagnostics for any query or witness that fails.

This seems reasonable.

> Unlike the lint queries themselves, it arguably should not fail hard, and should allow the rest of `c-s-c` to complete, it should just report any errors that occur to the user, and prompt them to report it as an issue. At least, that's my opinion on it, thoughts?

Also reasonable! Let's do that.

> Additionally, as I state in one of the comments, if no witness query exists, we ideally want to forward existing lint_results, so those values can still be used in a witness that may not need a query. Though looking at this now, it might be optimal to add another case to the `match` for specifically where there _is_ a witness to generate and _not_ a witness query present to do that

I like the latter idea better! The former would be confusing and probably a bit convoluted.

> and then on the catchall, instead just get filtered out (switching to filter_map instead of just map would allow this).

I'm afraid I didn't follow this part.

Overall, I think it'd be useful to have a bit more functionality fleshed out here before we merge this, because for all we know some of the pieces might not quite fit well enough together. Better to find that out ASAP :)

obi1kenobi:
> Is it just me, or are the tests not getting the correct rustdoc?

I'll fix this shortly. I wasn't sure which rustdoc version would end up in beta because there were a half dozen new rustdoc JSON versions in the last week.

I'll push new `trustfall-rustdoc-adapter` and `trustfall-rustdoc` releases and update c-s-c to use them.

src/witness_gen.rs
GlitchlessCode: Fixing that right now, good point, thanks!

src/witness_gen.rs
GlitchlessCode: That's true, not something I considered. I'll definitely just implement that

src/witness_gen.rs
GlitchlessCode: Only reason I personally prefer this is that the location where `run_witness_queries` is located is already heavily indented. I could definitely move it out, and considering how a function is typically used/meant to be used, that would also make sense. I just was not a fan of adding another layer of indentation just to loop the function call. But I think I'll switch it over, cause considering the context of what a function should be, that makes more sense.

src/check_release.rs
GlitchlessCode: This was the thing I was saying above about a "catchall". Technically it's not just if there's no witness data, just if there's no *query*. Either way, fixing this.

src/witness_gen.rs
obi1kenobi: The heavily indented concern is totally valid, I just think the better solution for it is to extract a function out of the _caller_. So instead of
```rust
        let _witness_results = all_results
            .par_iter()
            .map(
                |(semver_query, _, lint_results)| { // lots of stuff
```
we'd have
```rust
        let _witness_results = all_results
            .par_iter()
            .map(
                |(semver_query, _, lint_results)| extracted_fn(semver_query, lint_results))
```

Then this extracted function would have a more obvious function signature, and a few other benefits:
- Inside it, we could in principle run the witness query in parallel for each lint result, by using another `.par_iter()`.
- We wouldn't necessarily need to materialize a `Vec` of the witness results — we could just process each `anyhow::Result<BTreeMap<Arc<str>, FieldValue>>` value in streaming fashion as it comes out of the iterator.

obi1kenobi:
https://github.com/obi1kenobi/cargo-semver-checks/pull/1343 will fix the beta tests, and I believe nightly as well.

GlitchlessCode:
# Commit Notes
 Break out map, fix small issues

+ Extracts the mapping logic for witness queries to their own helper function above `run_check_release`
+ Switches mapping logic to use `filter_map`, allowing dropping of unused lint results
+ Removes unecessary vector collection, maintaining iterator format of witness query results
+ Switch to faster `.iter().cloned()` rather than `.clone().into_iterator()`
+ Fix potential issue with iterator in `run_witness_query`, returning early if `None` found

src/witness_gen.rs
```diff
@@ -1,48 +1,43 @@
...
+                    // If there is an extra query match, we don't know which is the "correct one"
+                    Some(_) => Err(anyhow::anyhow!(
+                        "witness query should match exactly one time, matched multiple times"
+                    )),
+                    None => Ok(query_result),
+                }
+            } else {
+                // If there is no query match, something has gone very wrong
+                Err(anyhow::anyhow!(
+                    "witness query should match exactly one time, matched zero times"
```
obi1kenobi, line 49: Imagine debugging a problem where this branch is hit.

You notice a `cargo-semver-checks` invocation where this error happens. You see: `witness query should match exactly one time, matched zero times` as the output.

What do you do next?

src/check_release.rs
```diff
@@ -230,6 +231,46 @@ fn print_triggered_lint(
...
 
+/// Helper function to map a given semver query, and its witness to a value for use in [`run_check_release`] to
+/// to an [`Option`].
+fn map_witness_queries<'query>(
+    semver_query: &'query SemverQuery,
+    lint_results: &Vec<BTreeMap<Arc<str>, FieldValue>>,
+    adapter: &VersionedRustdocAdapter,
+) -> Option<(
+    &'query SemverQuery,
+    Vec<anyhow::Result<BTreeMap<Arc<str>, FieldValue>>>,
```
obi1kenobi: I'm not entirely sold on returning this `Vec` as opposed to immediately acting on the results.

Consider writing out more of the witness logic (expanding beyond the original scope of the PR) so we can see what fits best?

GlitchlessCode:
Converting to a draft, more work is necessary

GlitchlessCode:
# Commit notes

Add witness text generation, refactor witness system

+ Add `QueryResult` type for simplicity
+ Add `run_witness_checks` function as witness generation entrypoint
+ Add error context
+ Add witness text generation immediately after witness query running to avoid passing witness query data around
+ TODO: Snapshot witness query data
+ Remove witness running logic from `check_release.rs`, replacing with single entrypoint call

# Comments
Took what we discussed about adding more context to the errors, added them as `anyhow` context. Also removed the passing around of vectors of witness query result data, instead immediately generating witness text from handlebars templates. Assuming all is well here, and there may be more work necessary, but I'm pretty happy with it, next step is to take the witness text and actually generate all the witness crates using the computed witness text.

src/query.rs
```diff
@@ -330,12 +332,11 @@ impl WitnessQuery {
...
-                    .map(Into::into)
                     .ok_or(anyhow::anyhow!(
-                        "inherited output key `{inherit}` does not exist"
+                        "inherited output key `{inherit}` does not exist in {source_map:?}"
                     ))?,
                 // Set a constant
-                InheritedValue::Constant(value) => value.clone(),
+                InheritedValue::Constant(value) => value.clone().into(),
             };
             mapped.insert(key.clone(), mapped_value);
```
obi1kenobi: Perhaps as a subsequent PR, can we make this `key.clone()` become an `Arc::clone()` by changing the type signatures around here?

I believe Trustfall supports any `impl Into<Arc<str>>` for the parameter names, and we don't want to keep copying and allocating new strings for each witness evaluation.

src/witness_gen.rs
```diff
@@ -0,0 +1,148 @@
...
+    GlobalConfig, SemverQuery,
+};
+
+/// Runs the witness query of a given [`WitnessQuery`] a given lint query match, and merges the witness query
+/// results with the existing lint results. Each query must match exactly once, and will fail with an
+/// [`anyhow::Error`] otherwise.
+///
+/// Overlapping output keys between the [`WitnessQuery`] and the [`SemverQuery`](crate::query::SemverQuery)
+/// will result in the result from the [`WitnessQuery`] silently overriding the same key from the
+/// [`SemverQuery`](crate::query::SemverQuery).
```
obi1kenobi: Can we make this an error actually? If we imagine cutting this "feature" I think a small number of use cases are mildly inconvenienced, but we also prevent a lot of possible tedious debugging. That seems very worth it, given we're optimizing for easy maintainability in this project.

Ideally, we'd even have a part of our test suite that asserts that output names between the lint query and the witness query are fully disjoint, just by running Trustfall compilation on both and comparing metadata.

src/witness_gen.rs
```diff
@@ -0,0 +1,148 @@
...
+fn run_witness_query(
+    adapter: &VersionedRustdocAdapter,
+    witness_query: &WitnessQuery,
+    mut lint_result: QueryResults,
+) -> Result<QueryResults> {
```
obi1kenobi: I can't say I'm a huge fan of `QueryResults` over an explicit type signature here. For example: is this `BTreeMap<_, _>` or `Vec<BTreeMap<_, _>>`? And what were the key and value types again?

src/witness_gen.rs
```diff
@@ -0,0 +1,148 @@
...
+                }
+            } else {
+                // If there is no query match, something has gone very wrong
+                Err(anyhow::anyhow!(
+                    "witness query should match exactly one time, matched zero times"
+                ))
+            }
+        })
+        .with_context(|| {
+            format!("Error running witness query with input arguments {arguments:?}")
```
obi1kenobi: In the common case, we don't hit this error. So let's not make the common case clone `arguments` just for the uncommon case of this error to be able to print them. We can reconstruct `arguments` if we need it here.

src/witness_gen.rs
```diff
@@ -0,0 +1,148 @@
...
+        }) => {
+            let witness_results = lint_results
+                .iter()
+                .cloned()
+                .map(|lint_result| {
+                    let witness_results = run_witness_query(adapter, witness_query, lint_result)
+                        .with_context(|| {
+                            format!(
+                                "Error running witness query for {}: {}",
+                                semver_query.id, semver_query.human_readable_name
```
obi1kenobi:
```suggestion
                                "Error running witness query for {}",
                                semver_query.id
```
Since our human-readable names aren't very good, and since users generally interact with lints via the `id` — for example when configuring them in `Cargo.toml`.

src/witness_gen.rs
```diff
@@ -0,0 +1,148 @@
...
+                            )
+                        })?;
+                    generate_witness_text(handlebars, witness_template, witness_results)
+                        .with_context(|| {
+                            format!(
+                                "Error generating witness text for witness {}: {}",
+                                semver_query.id, semver_query.human_readable_name
```
obi1kenobi:
```suggestion
                                "Error generating witness text for witness {}",
                                semver_query.id,
```

src/witness_gen.rs
```diff
@@ -0,0 +1,148 @@
...
+            semver_query,
+            lint_results
+                .iter()
+                .cloned()
+                .map(|lint_result| {
+                    generate_witness_text(handlebars, witness_template, lint_result).with_context(
+                        || {
+                            format!(
+                                "Error generating witness text for queryless witness {}: {}",
+                                semver_query.id, semver_query.human_readable_name
```
obi1kenobi:
```suggestion
                                "Error generating witness text for queryless witness {}",
                                semver_query.id
```

src/witness_gen.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+pub(crate) fn run_witness_checks(
+    config: &GlobalConfig,
+    _witness_dir: &Path,
+    adapter: &VersionedRustdocAdapter,
+    lint_results: &[(&SemverQuery, Duration, Vec<QueryResults>)],
+) {
+    // Have to pull out handlebars, since &GlobalConfig cannot be shared across threads
+    let handlebars = config.handlebars();
```
obi1kenobi, line 159: Oh, fascinating. I didn't know the type wasn't thread-safe. We might want to fix this at some point when we have no lower-hanging fruit to pick — but very nice comment in the meantime!

obi1kenobi (commented):
Good stuff! I think just one more polishing pass and we should be able to merge this.

obi1kenobi:
Btw I fixed the broken lint job, so merging the upstream `main` should fix your CI (except `nightly`).

src/witness_gen.rs
GlitchlessCode: Good to know. Personally just had that in for reading ergonomics, but I'll go over that again.

src/witness_gen.rs
GlitchlessCode: Yeah I wasn't particularly happy with silent behaviour either. I'll get an error in there.

src/witness_gen.rs
GlitchlessCode: Could you go more in depth for this one? I get what we don't want, I don't like that clone either, I just wasn't sure how to get rid of it.

obi1kenobi: Oh, just by repeating `witness_query.inherit_arguments_from(&lint_result).expect("failed to construct query arguments while creating an error")` from before, with an `.expect()` instead of an error since we know we shouldn't have errored out there if we got to here.

GlitchlessCode: Oh I see, yeah that makes sense! Thanks Predrag

GlitchlessCode:
# Commit Notes
 Add error for overlap, and small fixes

+ Adds an error for when an output key from a witness query overlaps that of the lint query
+ Removes the QueryResults type alias, reverting back to full type information
+ Avoids cloning arguments on the most common code path
+ Removes human_readable_name use in errors

GlitchlessCode:
Clippy is complaining about a very long type and is recommending breaking something out to a `type`, which is exactly what I just reverted. Any thoughts on what to do?

obi1kenobi:
I would suppress it with `#[expect(clippy::type_complexity)]`. It means well and it's a valid complaint, but:
- this isn't our public API
- the complexity is inherent, not accidental
- we'll probably refactor this code more in the future.

obi1kenobi (approved):

src/witness_gen.rs
```diff
@@ -0,0 +1,165 @@
...
+                        },
+                    )
+                })
+                .collect_vec(),
+        )),
+        _ => None,
+    }
+}
+
+pub(crate) fn run_witness_checks(
```
obi1kenobi, line 152:
```suggestion
#[expect(clippy::type_complexity)]
pub(crate) fn run_witness_checks(
```

obi1kenobi:
I quickly threw in the lint suppression and I'm marking this for merge as soon as CI finishes. Thanks for putting it together — onward! 🚀

src/witness_gen.rs
```diff
@@ -0,0 +1,166 @@
...
+use crate::{
+    query::{Witness, WitnessQuery},
+    GlobalConfig, SemverQuery,
+};
+
+/// Runs the witness query of a given [`WitnessQuery`] a given lint query match, and merges the witness query
+/// results with the existing lint results. Each query must match exactly once, and will fail with an
+/// [`anyhow::Error`] otherwise.
+///
+/// Overlapping output keys between the [`WitnessQuery`] and the [`SemverQuery`](crate::query::SemverQuery)
```
obi1kenobi:
```suggestion
/// Overlapping output keys between the [`WitnessQuery`] and the [`SemverQuery`]
```

PR #1354, opened 2025-07-01
Refactor rustdoc generation, replacing dynamic dispatch system

GlitchlessCode:
# Commit Notes

+ Adds new `generate_data_request` method on `RustdocGenerator` trait
+ Changes `load_rustdoc` method to require optional `CrateDataRequest`
+ Splits rustdoc loading into two steps, which raises up the `CrateDataRequest`, making it accessible elsewhere

# Comments

This almost definitely needs more work, as this was just a crude hashing of what needs to happen. I'm going to put this as a draft for now, just to get feedback, but it's far from perfect in my opinion.

What this *does* do is breaks out the `CrateDataRequest` type from being a deeply internal type to being something accessible at the top level, allowing for easy access to information about the crate being queried. Now, the value returned is optional, since feeding rustdoc directly does not create a `CrateDataRequest`, that's why there's `Option<CrateDataRequest<'_>>` in quite a few places. 

This is mainly a necessary dependency for witness generation, as the generation of a witness crate requires knowing information about where the crate exists, or that it exists on crates.io.

### DIFF
--- a/src/data_generation/request.rs
+++ b/src/data_generation/request.rs
@@ -13,8 +13,8 @@ use super::generate::GenerationSettings;
 use super::progress::{CallbackHandler, ProgressCallbacks};
 
 #[derive(Debug, Clone)]
-pub(super) struct RegistryRequest {
-    index_entry: tame_index::IndexVersion,
+pub(super) struct RegistryRequest<'a> {
+    index_entry: &'a tame_index::IndexVersion,
 }
 
 #[derive(Debug, Clone)]
@@ -24,7 +24,7 @@ pub(super) struct ProjectRequest<'a> {
 
 #[derive(Debug, Clone)]
 pub(super) enum RequestKind<'a> {
-    Registry(RegistryRequest),
+    Registry(RegistryRequest<'a>),
     LocalProject(ProjectRequest<'a>),
 }
 
@@ -212,7 +212,7 @@ pub(crate) struct CrateDataRequest<'a> {
 
 impl<'a> CrateDataRequest<'a> {
     pub(crate) fn from_index(
-        index_entry: tame_index::IndexVersion,
+        index_entry: &'a tame_index::IndexVersion,
         default_features: bool,
         extra_features: BTreeSet<Cow<'a, str>>,
         build_target: Option<&'a str>,
--- a/src/data_generation/request.rs
+++ b/src/data_generation/request.rs
@@ -13,8 +13,8 @@ use super::generate::GenerationSettings;
 use super::progress::{CallbackHandler, ProgressCallbacks};
 
 #[derive(Debug, Clone)]
-pub(super) struct RegistryRequest<'a> {
-    index_entry: &'a tame_index::IndexVersion,
+pub(super) struct RegistryRequest {
+    index_entry: tame_index::IndexVersion,
 }
 
 #[derive(Debug, Clone)]
@@ -24,7 +24,7 @@ pub(super) struct ProjectRequest<'a> {
 
 #[derive(Debug, Clone)]
 pub(super) enum RequestKind<'a> {
-    Registry(RegistryRequest<'a>),
+    Registry(RegistryRequest),
     LocalProject(ProjectRequest<'a>),
 }
 
@@ -212,7 +212,7 @@ pub(crate) struct CrateDataRequest<'a> {
 
 impl<'a> CrateDataRequest<'a> {
     pub(crate) fn from_index(
-        index_entry: &'a tame_index::IndexVersion,
+        index_entry: tame_index::IndexVersion,
         default_features: bool,
         extra_features: BTreeSet<Cow<'a, str>>,
         build_target: Option<&'a str>,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -556,6 +556,13 @@ note: skipped the following crates since they have no library target: {skipped}"
                 let start = std::time::Instant::now();
                 let name = selected.current_crate_data.name.clone();
 
+                let current_data_request =
+                    generate_data_request(config, &*current_loader, &selected.current_crate_data)
+                        .map_err(|err| log_terminal_error(config, err))?;
+                let baseline_data_request =
+                    generate_data_request(config, &*baseline_loader, &selected.baseline_crate_data)
+                        .map_err(|err| log_terminal_error(config, err))?;
+
                 let data_storage = generate_crate_data(
                     config,
                     generation_settings,
@@ -563,6 +570,8 @@ note: skipped the following crates since they have no library target: {skipped}"
                     &*baseline_loader,
                     &selected.current_crate_data,
                     &selected.baseline_crate_data,
+                    &current_data_request,
+                    &baseline_data_request,
                 )
                 .map_err(|err| log_terminal_error(config, err))?;
 
@@ -745,6 +754,15 @@ impl WitnessGeneration {
     }
 }
 
+fn generate_data_request<'a>(
+    config: &mut GlobalConfig,
+    loader: &'a dyn rustdoc_gen::RustdocGenerator,
+    crate_data: &rustdoc_gen::CrateDataForRustdoc<'a>,
+) -> Result<Option<data_generation::CrateDataRequest<'a>>, TerminalError> {
+    loader.generate_data_request(config, crate_data)
+}
+
+#[expect(clippy::too_many_arguments)]
 fn generate_crate_data(
     config: &mut GlobalConfig,
     generation_settings: data_generation::GenerationSettings,
@@ -752,12 +770,14 @@ fn generate_crate_data(
     baseline_loader: &dyn rustdoc_gen::RustdocGenerator,
     current_crate_data: &rustdoc_gen::CrateDataForRustdoc<'_>,
     baseline_crate_data: &rustdoc_gen::CrateDataForRustdoc<'_>,
+    current_data_request: &Option<data_generation::CrateDataRequest<'_>>,
+    baseline_data_request: &Option<data_generation::CrateDataRequest<'_>>,
 ) -> Result<DataStorage, TerminalError> {
     let current_crate = current_loader.load_rustdoc(
         config,
         generation_settings,
         data_generation::CacheSettings::ReadWrite(()),
-        current_crate_data,
+        current_data_request,
     )?;
 
     let baseline_crate_name = &baseline_crate_data.name;
@@ -768,7 +788,7 @@ fn generate_crate_data(
             config,
             generation_settings,
             data_generation::CacheSettings::ReadWrite(()),
-            baseline_crate_data,
+            baseline_data_request,
         )?;
 
         // The baseline rustdoc JSON may have been cached; ensure its rustdoc version matches
@@ -791,7 +811,7 @@ fn generate_crate_data(
                 config,
                 generation_settings,
                 data_generation::CacheSettings::WriteOnly(()),
-                baseline_crate_data,
+                baseline_data_request,
             )?;
 
             assert_eq!(
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -371,24 +371,20 @@ impl Check {
         &self,
         config: &mut GlobalConfig,
         source: &RustdocSource,
-    ) -> anyhow::Result<Box<dyn rustdoc_gen::RustdocGenerator>> {
+    ) -> anyhow::Result<rustdoc_gen::RustdocGenerator> {
         let target_dir = self.get_target_dir(source)?;
         Ok(match source {
             RustdocSource::Rustdoc(path) => {
-                Box::new(rustdoc_gen::RustdocFromFile::new(path.to_owned()))
+                rustdoc_gen::RustdocFromFile::new(path.to_owned()).into()
             }
             RustdocSource::Root(root) => {
-                Box::new(rustdoc_gen::RustdocFromProjectRoot::new(root, &target_dir)?)
+                rustdoc_gen::RustdocFromProjectRoot::new(root, &target_dir)?.into()
             }
             RustdocSource::Revision(root, rev) => {
                 let metadata = manifest_metadata_no_deps(root)?;
                 let source = metadata.workspace_root.as_std_path();
-                Box::new(rustdoc_gen::RustdocFromGitRevision::with_rev(
-                    source,
-                    &target_dir,
-                    rev,
-                    config,
-                )?)
+                rustdoc_gen::RustdocFromGitRevision::with_rev(source, &target_dir, rev, config)?
+                    .into()
             }
             RustdocSource::VersionFromRegistry(version) => {
                 let mut registry = rustdoc_gen::RustdocFromRegistry::new(&target_dir, config)?;
@@ -396,7 +392,7 @@ impl Check {
                     let semver = semver::Version::parse(ver)?;
                     registry.set_version(semver);
                 }
-                Box::new(registry)
+                registry.into()
             }
         })
     }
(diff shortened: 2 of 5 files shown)


src/rustdoc_gen.rs
```diff
@@ -455,13 +468,32 @@ impl RustdocGenerator for RustdocFromProjectRoot {
                 }
             }
         }).into_terminal_result()?;
+
+        Ok(Some(generate_data_request(
+            config,
+            CrateSource::ManifestPath { manifest },
+            crate_data,
+        )))
```
obi1kenobi: I notice this part is the same across all three implementations that actually generate data.

It's always:
```rust
        Ok(Some(generate_data_request(
            config,
            // *some* `CrateSource` variant
            crate_data,
        )))
```

Can we keep refactoring this stuff down to where we just return `CrateSource` here, and then plug in the `config` and `crate_data` somewhere else in a common layer across all three implementations? That layer can then do the `Ok(Some(generate_data_request(..)))` wrapping too, if needed.

src/rustdoc_gen.rs
```diff
@@ -455,13 +468,32 @@ impl RustdocGenerator for RustdocFromProjectRoot {
...
+        config: &mut GlobalConfig,
+        generation_settings: super::data_generation::GenerationSettings,
+        cache_settings: super::data_generation::CacheSettings<()>,
+        data_request: &Option<CrateDataRequest<'_>>,
+    ) -> Result<VersionedStorage, TerminalError> {
+        let Some(data_request) = data_request else {
+            return Err(anyhow::anyhow!("`data_request` was unexpectedly `None`"))
+                .into_terminal_result();
+        };
+
```
obi1kenobi: This method looks like it has the same shape for all the implementations of `RustdocGenerator` aside from the one that doesn't actually generate anything.

In all three remaining ones, it does essentially an `unwrap_or_err()` on `data_request` followed by a call to `generate_rustdoc()`.

This means we haven't really boiled this down to quite the right minimal abstraction yet.

src/rustdoc_gen.rs
```diff
@@ -519,15 +551,23 @@ fn extract_tree(tree: gix::Id<'_>, target: &std::path::Path) -> anyhow::Result<(
 }
 
 impl RustdocGenerator for RustdocFromGitRevision {
+    fn generate_data_request<'data>(
+        &'data self,
+        config: &mut GlobalConfig,
+        crate_data: &CrateDataForRustdoc<'data>,
+    ) -> Result<Option<CrateDataRequest<'data>>, TerminalError> {
+        self.path.generate_data_request(config, crate_data)
+    }
```
obi1kenobi: Having to name the lifetime of `&self` is weird and tends to cause lots of trouble with lifetime variance. Why did this become necessary, and is it feasible to avoid it?

src/data_generation/request.rs
```diff
@@ -24,7 +24,7 @@ pub(super) struct ProjectRequest<'a> {
 
 #[derive(Debug, Clone)]
 pub(super) enum RequestKind<'a> {
-    Registry(RegistryRequest<'a>),
+    Registry(RegistryRequest),
```
obi1kenobi: This is a bit of a strange change: now one of the variants stores references, and the other is by value. I'm not sure what the motivation was, so maybe it's totally necessary — I just wanted to raise it as a discussion point.

obi1kenobi (commented):
I agree this is a solid intermediate point but it still needs some more boiling down and refactoring. This is definitely my least favorite code, so hopefully we can help you get in and out ASAP :)

GlitchlessCode:
All those pieces of feedback were basically my thoughts too, hence why a draft. I'll get on improving as much of that as I can.

src/data_generation/request.rs
GlitchlessCode: Yeah for this one it's kinda tricky. Previously this wasn't an issue, the `RegistryRequest<'a>`, which took reference to an `IndexVersion`, was just taking that `&'a IndexVersion` from the `IndexKrate`. More specifically, the `load_rustdoc` function for a registry request would get an `IndexKrate`, then iterate over all the contained crates (using `.iter()` to iterate over refs) until it found the requested crate (filtering by version), and would give that versioned crate, which is still just referring to the contents of the `IndexKrate` as a `&IndexVersion`. This used to be fine since it would immediately consume the `RegistryRequest<'a>` to create the rustdoc, since this was all in the same function. However, now, the `IndexKrate` that is referred to by `RegistryRequest<'a>` is owned by the function, in this case `generate_data_request`, and then tries to return a `CrateDataRequest<'a>` which contains the `RegistryRequest<'a>`, but that leaves a dangling reference, since it's a reference to the `IndexVersion` inside the `IndexKrate`, and thus, evidently, we get errors. I changed up the system to make the `IndexVersion` owned (just by using `into_iter()` instead of `iter()` on the `IndexKrate`), but that meant that `RegistryRequest` didn't make sense to take the value by reference. 

Is it possible that there may be some way to keep it how it was with some more refactoring? Maybe? That's just how I brute-forced it for now to get something that compiles.

src/rustdoc_gen.rs
GlitchlessCode: The "necessity" for this was that the `ManifestPath` variant on `CrateSource` expects an `&Manifest`. The only reason this isn't a problem here, unlike what  I mentioned below for the `RegistryRequest` situation, rather than being owned by the function, the `Manifest` is owned by `self`, so as long as `self` exists, then we can refer to the manifest. Now, this could be theoretically solved by moving manifest-based requests to also use an owned value. This would make `RequestKind` no longer depend on any additional lifetimes, but of course, means passing stuff by value. Additionally, at least with the `RegistryRequest`, the `IndexKrate` that was causing the problem was being regenerated each and every time, so there was not much of a difference between owning the `IndexVersion` and borrowing it. However, here, we'd need to probably clone the `&Manifest` to make it owned, or `.remove` it from `self.manifests`, though I'm not sure if a manifest can be reused, so cloning would probably be safer/break less stuff down the line. I don't really know how hot this code path is, so cloning may or may not be okay.

So to summarize, this comment, and the one below will probably be resolved at the same time by the same thing, whatever that may be, since they both deal with the consequences of borrowing values, just one was borrowing a value tied to `&self`, and one was borrowing a value that was tied to a value owned by the function.

GlitchlessCode:
There's quite a bit that's changed here, so here's a quick rundown:

- `RustdocGenerator` has been converted from a trait to an enum, with 4 variants, one for each generator type.
- `CoupledRustdocGenerator` enum has been added. While definitely a little unecessary and verbose, I found it was the best way to couple all of the data like crate data and the owned `IndexKrate` to its respective rustdoc generator. Technically that data could be retrieved and managed in `lib.rs` instead, but this felt a little but contained, even if it gives us more control over the data compared to prviously.
- Once again, we return a `Result<Option<CrateDataRequest>, TerminalError>`, because that's what logically makes sense considering it's both a fallible operation, and for a rustdoc file source, is non-existant, thus is optional. 
- `generate_rustdoc` now is called in exactly one central place

GlitchlessCode:
The error printing for snapshot tests has also been updated. More specifically, snapshot tests used to format errors with `Display`, now they format them with `Debug`. This is to reflect the actual behaviour of `cargo-semver-checks`, which is to debug print an `anyhow::Error`, since that gets full context, rather than the first error only. The only reason this is a part of this PR is that this PR also adds some new wrapping context to some snapshotted errors, which were triggering test fails, since it was only printing the outermost context, rather than the full error.

src/rustdoc_gen.rs
```diff
@@ -285,28 +294,222 @@ fn generate_rustdoc(
...
+    ProjectRoot {
+        generator: &'a RustdocFromProjectRoot,
+        crate_data: &'a CrateDataForRustdoc<'a>,
+    },
+    // This is a wrapper around a ProjectRoot
+    GitRevision {
+        generator: &'a RustdocFromGitRevision,
+        crate_data: &'a CrateDataForRustdoc<'a>,
+    },
+    // A Registry needs the crate data, as well as the list of crates from crates.io
```
obi1kenobi: nit: presumably the list of releases, not crates?

src/lib.rs
```diff
@@ -556,13 +552,35 @@ note: skipped the following crates since they have no library target: {skipped}"
...
+                    config,
+                    &selected.current_crate_data,
+                )
+                .map_err(|err| log_terminal_error(config, err))?;
+                let baseline_loader = rustdoc_gen::CoupledRustdocGenerator::couple_data(
+                    &baseline_loader,
+                    config,
+                    &selected.baseline_crate_data,
+                )
+                .map_err(|err| log_terminal_error(config, err))?;
```
obi1kenobi: I notice we're passing the `current_crate_data` and `baseline_crate_data` here, and then taking the two loaders and passing them together with `current_crate_data` and `baseline_crate_data` into `generate_crate_data()`.

Then `generate_crate_data()` has three pairs of current/baseline data, which all have to be in sync — if any of them gets passed backwards, that's a bug.

Can we have the coupled types include their corresponding crate data, so we don't have to pass it separately later? It seems like we're going to use that data together anyway, so we might as well hang onto it.

src/rustdoc_gen.rs
```diff
@@ -285,28 +294,222 @@ fn generate_rustdoc(
...
+            RustdocGenerator::GitRevision(generator) => Ok(Self::GitRevision {
+                generator,
+                crate_data,
+            }),
+
+            RustdocGenerator::Registry(generator) => {
+                let krate = generator.get_krate(config, crate_data).map_err(|err| {
+                    terminal_context(
+                        err,
+                        "Error while retrieving index of crates from `RustdocFromRegistry`",
```
obi1kenobi: Is this diagnostic meant to be user-facing?

Asking because `RustdocFromRegistry` is not a user-facing concept. It makes sense to you and me, but for users it's completely opaque, confusing, and not helpful.

So if this is meant to be user-facing, we might want to choose a different phrasing here.

src/rustdoc_gen.rs
```diff
@@ -285,28 +294,222 @@ fn generate_rustdoc(
...
+                        "Error while getting crate source from `RustdocFromRegistry`",
+                    )
+                })?,
+        };
+
+        // This should never error since the match case above always return early when there is no attached crate_data
+        let crate_data = self
+            .get_crate_data()
+            .ok_or(anyhow::anyhow!("The `crate_data` property on a `CoupledRustdocGenerator` was unexpectedly `None` while generating a data request"))
+            .into_terminal_result()?;
```
obi1kenobi: If we think this should never error:
- Is it feasible to express that via the type system, e.g. via an exhaustive match of some kind? (Possibly not, and that's fine — just worth thinking it over and double-checking.)
- If we're confident it shouldn't ever error (or it's a bug), this should be an `unreachable!()` and cause a crash, so that we _definitely_ get an error report. The panic crash output has instructions for opening an issue to report it.

src/rustdoc_gen.rs
```diff
@@ -285,28 +294,222 @@ fn generate_rustdoc(
...
+        data_request: &Option<CrateDataRequest<'_>>,
+    ) -> Result<VersionedStorage, TerminalError> {
+        if let Self::File { generator } = self {
+            return generator.load_rustdoc();
+        }
+
+        let data_request = data_request
+            .as_ref()
+            .ok_or(anyhow!("The `data_request` argument for loading the rustdoc from a `CoupledRustdocGenerator` was unexpectedly `None`"))
+            .into_terminal_result()?;
```
obi1kenobi: If we're not willing to accept a `None` value for `data_request` here, let's not accept it in the function signature. Otherwise we're allowing code that "looks correct" to actually fail at runtime.

Let's try to force the `Self::File` path to call a separate method instead.

test_outputs/snapshot_tests/cargo_semver_checks__snapshot_tests__multiple_ambiguous_package_name_definitions-output.snap
```diff
@@ -3,11 +3,14 @@ source: src/snapshot_tests.rs
 expression: result
 ---
 --- error ---
-package `example` is ambiguous: it is defined by in multiple manifests within the root path test_crates/manifest_tests/multiple_ambiguous_package_name_definitions
+Error while getting crate source from `RustdocFromProjectRoot`
```
obi1kenobi: nit: our error messages always start with lowercase letters, because that makes them look better when listing multiple causes

GlitchlessCode:
This version updates the system to use a generic state system, with transitions from state to state. This makes invalid state unrepresentable, since instead of dealing with optional values, data is always coupled to what other data is necessary for that state to progress. I'm personally actually really happy with where this is at now.

GlitchlessCode:
Arguably, the generic system is a little excessive, but
1. It allows for fewer additional structs, before I implemented that, I was going to be submitting a version with two separate structs instead of the `StatefulRustdocGenerator`, which would have been identical to each other, other than the type of coupled state data
2. Technically still leaves room for further refactors or expansion. If more data may need to be pulled out, you can add more states in between, before, or after the existing states. This is isn't exactly likely to happen, but it's a useful possibility to have

src/rustdoc_gen.rs
```diff
@@ -285,28 +294,230 @@ fn generate_rustdoc(
...
+
+        Ok(Self {
+            coupled_state: coupled_data,
+            crate_data,
+        })
+    }
+
+    /// Ready a [`CoupledState`] for rustdoc generation, extracting necessary internal data,
+    /// and creating an appropriate data request
+    pub(crate) fn ready_generator(
```
obi1kenobi: nit: `ready` is a bit unfortunate as a word, because it's unclear whether it's being used as a verb or noun, and in the absence of context it'll be assumed to be a noun because that's more common.

So this method seemed like "get a ready generator" and not like "prepare the generator."

src/rustdoc_gen.rs
```diff
@@ -285,28 +294,230 @@ fn generate_rustdoc(
...
+                data_request,
+            },
+            crate_data,
+        })
+    }
+}
+
+impl<'a> StatefulRustdocGenerator<'a, ReadyState<'a>> {
+    // TODO: Use the data request
+    #[expect(dead_code)]
```
obi1kenobi: What needs to happen for this to be used?

src/rustdoc_gen.rs
```diff
@@ -318,18 +529,15 @@ impl RustdocFromFile {
...
-        _generation_settings: super::data_generation::GenerationSettings,
-        _cache_settings: super::data_generation::CacheSettings<()>,
-        _crate_data: &CrateDataForRustdoc<'_>,
-    ) -> Result<VersionedStorage, TerminalError> {
+    pub(crate) fn load_rustdoc(&self) -> Result<VersionedStorage, TerminalError> {
         trustfall_rustdoc::load_rustdoc(&self.path, None)
-            .map_err(anyhow::Error::from)
+            .with_context(|| {
+                format!(
+                    "Error loading rustdoc for `RustdocFromFile` from `{:?}`",
```
obi1kenobi: Let's make all of these context messages lowercase, and let's make them not reference internal code things like `RustdocFromFile` and such. We want error context to be written in user-facing terms, or else it's just confusing.

GlitchlessCode: I missed that one while rewriting my errors, my bad!

src/rustdoc_gen.rs
GlitchlessCode: This is to be used by the witness generation system. This will be used in the very next witness generation PR.

src/rustdoc_gen.rs
GlitchlessCode: Good point, I'll change the wording on that!

Issue #1355, opened 2025-07-02
Specify default CLI params in Cargo.toml's metadata


nyurik:
### Describe your use case

I use [release-plz](https://release-plz.dev/) to automate release process. Internally, it uses semver-checks to compute the next version. My crate requires custom features to be properly validated, which I cannot do either via release-plz configuration (it should also be fixed), nor via the `[package.metadata.cargo-semver-checks]` section in Cargo.toml.

### Describe the solution you'd like

I think support for `--only-explicit-features` (bool) and `--features` (str[]) in the manifest's metadata would solve this problem.

### Alternatives, if applicable

Ideally, release-plz should also provide a way to pass through the config values, but this path is less desirable because that means that every system that internally uses semver-checker would also need to do that - so easier to configure it via Config.toml rather than every wrapper to support it.

obi1kenobi:
I'm not sure this is the right abstraction. There's an alternative that would make sense, but it'd be a lot more work to build.

The workflow for many complex crates requires running a matrix of `cargo-semver-checks` jobs: with different feature combinations, on different build targets (e.g. `x86-64` / `aarch64`, Linux / Windows) etc. This is because in any one of those runs, `cargo-semver-checks` only sees the items that conditional compilation allows to be present with those features on that target triple.

As such, it doesn't make sense to me to have a _single_ `cargo-semver-checks` config in `Cargo.toml`. If going down this route, that would be a better feature to add into `release-plz` than here.

But an extension of this idea would make sense here: a notion of "profiles" kind of similar to `cargo` — named combinations of settings like features / target triple etc. that users can invoke by name. I think that would make sense, but it would be quite a bit of work to design, implement, document, and test. That would be very nice to have! But I'll be honest that it's probably not an immediate priority given we have very limited development bandwidth and have many higher-impact things into which we could invest that bandwidth (e.g. more lints!).

Of course, this could change if a commercial sponsor wanted to fund the development of such a feature.

PR #1356, opened 2025-07-02
Allow crates to set metadata for some CLI params

nyurik:
implement #1355

This is a WIP to handle several params from the metadata config, allowing users to specify values without using the CLI params

### DIFF
--- a/src/manifest.rs
+++ b/src/manifest.rs
@@ -76,9 +76,17 @@ pub(crate) struct MetadataTable {
 /// or `[workspace.metadata]`.
 #[derive(Debug, Clone, Deserialize)]
 #[non_exhaustive]
+#[serde(rename_all = "kebab-case")]
 pub(crate) struct SemverChecksTable {
     /// Holds the `lints` table, if it is declared.
     pub(crate) lints: Option<LintTable>,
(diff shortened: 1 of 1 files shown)


nyurik:
Feedback is welcome on the best approach to combine the CLI and the config metadata.  My understanding is that currently the code:
* parses CLI and creates `cargo_semver_checks::Check` and runs `cargo_semver_checks::Check::check_release` 
* inside the `check_release`, parses `manifest` and `workspace.manifest` keys from the `Cargo.toml` files, and creates the list of lints to override

I guess i will have to combine the `Check` instance with the metadata inside the `check_release` - by checking if it is set to heuristic, and if it is, override it with the metadata values if set.

obi1kenobi:
I just had a chance to take another look at this PR. Unfortunately, it doesn't seem to address the concerns I shared in the issue [here](https://github.com/obi1kenobi/cargo-semver-checks/issues/1355#issuecomment-3029736711). I think we'll need to evolve the config in a different direction instead, as I mentioned there.

I'll close the PR for the time being, since the solution for that issue will be in a different direction. To speed up the timeline for solving that issue, the most impactful thing community members can do is get their employers to fund `cargo-semver-checks` maintenance and development.

PR #1371, opened 2025-07-22
Parse each query three times per query-lint pair (instead of six times)

CLIDragon:
**Purpose of Change**
Currently for each lint-crate pair, we check it three times for `(new, new)`, `(new, old)` and `(old, old)` crates. For each of these combinations, the lint is parsed twice.

**Describe the Solution**
Add a local cache into `check_query_execution`. This reduced the parses to three times per query-lint pair, plus once per lint.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -626,7 +626,7 @@ mod tests {
         crate_pair_name: &String,
         indexed_crate_new: &VersionedIndex<'_>,
         indexed_crate_old: &VersionedIndex<'_>,
-        indexed_query: Arc<IndexedQuery>,
+        parsed_query: Arc<IndexedQuery>,
     ) -> (String, Vec<BTreeMap<String, FieldValue>>) {
         let adapter = VersionedRustdocAdapter::new(indexed_crate_new, Some(indexed_crate_old))
             .expect("could not create adapter");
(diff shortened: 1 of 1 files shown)


src/query.rs
```diff
@@ -624,6 +626,7 @@ mod tests {
         crate_pair_name: &String,
         indexed_crate_new: &VersionedIndex<'_>,
         indexed_crate_old: &VersionedIndex<'_>,
+        parsed_query: Arc<IndexedQuery>,
```
obi1kenobi: I'm not sure we need ownership here. We might be able to get away with just `&IndexedQuery` and avoid a bunch of cloning.

src/query.rs
```diff
@@ -713,9 +714,15 @@ mod tests {
         indexed_crate: &VersionedIndex<'_>,
         crate_pair_name: &String,
         crate_version: &str,
+        indexed_query: Arc<IndexedQuery>,
```
obi1kenobi: I would move this arg next to `semver_query` both here and in the other function, because it's directly tied to `semver_query` in a non-obvious way. If they get out of sync, we'll have a bad time.

I'd also leave a comment explaining the connection between them.

CLIDragon:
The change I would like to make is to introduce a `run_query_with_indexed_query` function to `VersionedRustdocAdapter` so that the indexed query can be used (Alternatively, I would like to expose the inner adapter from `VersionedRustdocAdapter`. However, I don't think this is possible). This would also make it so that `semver_query` is an unnecessary argument since the variables can be extracted from the `IndexedQuery` and the string is no longer neeeded.

obi1kenobi (approved):

obi1kenobi:
I don't think you'd be able to expose the inner adapter. Adding the new method seems worth a shot, though I think you'll have to add a direct dependency on `trustfall_core`. Not a dealbreaker, just a bit annoying.

Issue #1279, opened 2025-06-05
New lints: `#[must_use]` removed


obi1kenobi:
> I also think I am inadvertently suggesting another lint: If a maintainer removes a must_use, in version `1.2.3` and someone writes code without removing it, that person cannot downgrade to 1.2.2 without clippy breaking. Therefore, removing a `must_use` would also be a minor bump.  

 _Originally posted by @schneems in [#1278](https://github.com/obi1kenobi/cargo-semver-checks/issues/1278#issuecomment-2945356722)_

This would be in the same category as non-breaking API additions: it should be allow-by-default and opt-in.

We'll need variants of the lint for all places where `#[must_use]` may appear. Use the existing "`#[must_use]` removed" lints to guide where a corresponding removal lint should exist.

- [x] `struct_must_use_removed`: #1392 
- [x] `inherent_method_must_use_removed`: #1516
- [x] `function_must_use_removed`: #1400 
- [x] `trait_must_use_removed`: #1515
- [x] `union_must_use_removed`: #1399

~~Rust 1.89 has a bug around `#[must_use]` that makes the attribute not appear as expected in rustdoc JSON. You'll likely need to add each new lint here to the list of exemptions for Rust 1.89 tests [here]( https://github.com/obi1kenobi/cargo-semver-checks/blob/074e0fdb22494fa7b19bbc55f350073d8f53bd85/src/query.rs#L848-L867)~~ Our minimum supported Rust version is 1.90+ so the 1.89 bug is no longer a concern.

nk9:
Given that the code for this lint has been merged, can this issue be closed?

obi1kenobi:
No — there are many more lints to write here. Updated the top post with a list.

nk9:
I'd like to give this a try, but I'm new to trustfall. I was thinking of tackling `struct_must_use_removed` and modeling it on the just-completed `enum_must_use_removed`. Does that seem like a good starting point?

obi1kenobi:
Yeah! Between `struct_must_use_added` and `enum_must_use_removed` you should have everything you need in the query. Also make sure to check out the contributing guide for writing new lints, there are some useful scripts you should make use of.

nk9:
[...]

PR #1392, opened 2025-08-21
Add `struct_must_use_removed`

nk9:
Part of #1279.


obi1kenobi:
This seems to be failing only the `#[must_use]` snapshot, and passing on the others we were discussing.

`#[must_use]` has a bug in Rust 1.89 specifically, so lints targeting it are exempted from Rust 1.89 (current stable) tests: https://github.com/obi1kenobi/cargo-semver-checks/blob/074e0fdb22494fa7b19bbc55f350073d8f53bd85/src/query.rs#L848-L867

I'd recommend adding this lint to the exemptions list, which will make the Rust stable tests pass. Then we can see if any other issues arise.

obi1kenobi:
Out of curiosity, does this pass the test suite on your local machine?

Because these tests which we were discussing should not have been affected by anything this PR did:
```
    query::tests_lints::safe_function_requires_more_target_features
    query::tests_lints::safe_function_target_feature_added
    query::tests_lints::safe_inherent_method_requires_more_target_features
    query::tests_lints::safe_inherent_method_target_feature_added
    query::tests_lints::trait_method_target_feature_removed
    query::tests_lints::unsafe_function_requires_more_target_features
    query::tests_lints::unsafe_function_target_feature_added
    query::tests_lints::unsafe_inherent_method_requires_more_target_features
    query::tests_lints::unsafe_inherent_method_target_feature_added
    query::tests_lints::unsafe_trait_method_requires_more_target_features
    query::tests_lints::unsafe_trait_method_target_feature_added
```

Their issues are around a different bug, one concerning `#[target_feature]` and not `#[must_use]`.

PR #1396, opened 2025-08-26
Replace tuple with a LintResult struct

kornelski:
Fixes a `type_complexity` warning. Keeps overrides together with the results.

I've also changed `filter_map` to `for_each` to avoid relying on rayon's iterator not being lazy.


obi1kenobi (approved):
Thank you!

PR #1400, opened 2025-09-01
Add `function_must_use_removed`

nk9:
More work on #1279


test_crates/function_must_use_removed/old/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+#[must_use]
+fn MustUsePrivateFunction() {}
+
+
+// This function was removed in the new version of the crate with its attribute.
+// It should NOT be reported by this rule to avoid duplicate lints.
+// It should be reported as a removed pub item that is part of the crate's API.
+
+#[must_use]
+pub fn MustUseRemovedFunction() {}
```
obi1kenobi:
```suggestion
pub fn MustUseRemovedFunction() {}

```
Please make sure all text files end in a single trailing newline, to ensure UNIX utilities work correctly with them. There's almost certainly an editor setting you could toggle to make this happen automatically.

test_outputs/witnesses/function_missing.snap
```diff
@@ -38,3 +38,8 @@ hint = 'function_missing::will_be_removed_fn(...);'
 filename = 'src/lib.rs'
 begin_line = 6
 hint = 'function_missing::pub_use_removed_fn(...);'
+
+[["./test_crates/function_must_use_removed/"]]
+filename = 'src/lib.rs'
+begin_line = 46
+hint = 'function_must_use_removed::MustUseRemovedFunction(...);'
```
obi1kenobi:
```suggestion
hint = 'function_must_use_removed::MustUseRemovedFunction(...);'

```
Out of curiosity, did you edit this file by hand?

If so, please don't do that in the future — let `insta` and `cargo-insta` edit it as appropriate.

src/lints/function_must_use_removed.ron
```diff
@@ -0,0 +1,69 @@
+SemverQuery(
+    id: "function_must_use_removed",
+    human_readable_name: "function #[must_use] removed",
+    description: "A function is no longer marked with #[must_use]",
+    required_update: Minor,
+    lint_level: Allow,
+
+
```
obi1kenobi:
```suggestion
```

test_crates/function_must_use_removed/old/src/lib.rs
nk9: Indeed, I have this set on Sublime, but I've been using RustRover recently and somehow that setting isn't on by default. Fixed now!

test_outputs/witnesses/function_missing.snap
nk9: Ah, OK. I made that change because it seemed to be necessary to get the tests to pass. Perhaps I actually needed to run the rustdoc regeneration script. Will avoid this in future.

obi1kenobi: Yes, every time you change any test crates, the rustdoc needs to be regenerated :+1:

PR #1404, opened 2025-09-01
Simpler required/suggested bump collecting

kornelski:
`required_versions` doesn't need to be a `Vec`.

I've also renamed some fields and variables to better match user-facing terminology (queries -> checks).


kornelski:
Nightly tests fail due to rustdoc v56

kornelski:
I've initially used a tuple for counting minor/major bumps, but having `.0 += 1` and such wasn't readable.

I'm not sure what other alternative would be there. `required_versions.iter().filter(|x| *x == &RequiredSemverUpdate::Minor).count()` is pretty wasteful - it allocates a vec only to count items in it. 

I've dropped the witness change commit.

PR #1414, opened 2025-09-04
[Witness Generation] Add empty witness crate generation

GlitchlessCode:
# Commit Notes

+ Adds `rand` and lazy `run_id` generation to config
+ Adds simple empty witness crate generation
+ Adds `target` directory finding
+ Adds `baseline` and `current` data retrieval which is necessary for future witness testing
+ Modifies unstable flag to be boolean based, rather than requiring a path, since `target` is now used instead
+ Is accompanied by the writer's apology for such a bizarre delay in project development

# Comments

Well, `n`th time's the charm I guess? I'm back working on this, ideally much more consistently. I strangely feel more interested and encouraged only now that my life is filling with extra responsibilities.

I'm pretty happy with how this generates the *empty* witness crates, now to do the actual generation, which ideally should be fairly simple (hardest part should be the `Cargo.toml`, since that needs version info, and features and whatnot). There are one or two things in this PR that I feel may need touch-ups, but overall I'm happy with it.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,7 +1,8 @@
 use anstream::{AutoStream, ColorChoice};
 use anstyle::{AnsiColor, Color, Reset, Style};
 use clap::ValueEnum;
-use std::{collections::HashSet, io::Write};
+use rand::Rng;
+use std::{collections::HashSet, io::Write, sync::LazyLock};
 
 use crate::templating::make_handlebars_registry;
 
@@ -9,6 +10,10 @@ use crate::templating::make_handlebars_registry;
 pub struct GlobalConfig {
     level: Option<log::Level>,
     handlebars: handlebars::Handlebars<'static>,
+    /// The unique ID for any particular run
+    ///
+    /// This value is lazily initialized on first access
+    run_id: LazyLock<String>,
     /// Minimum rustc version supported.
     ///
     /// This will be used to print an error if the user's rustc version is not high enough.
@@ -37,6 +42,13 @@ impl GlobalConfig {
         Self {
             level: None,
             handlebars: make_handlebars_registry(),
+            run_id: LazyLock::new(|| {
+                let rng = rand::rng();
+                rng.sample_iter(rand::distr::Alphanumeric)
+                    .take(16)
+                    .map(char::from)
+                    .collect()
+            }),
             minimum_rustc_version: semver::Version::new(1, 87, 0),
             stdout: AutoStream::new(Box::new(std::io::stdout()), stdout_choice),
             stderr: AutoStream::new(Box::new(std::io::stderr()), stderr_choice),
@@ -46,6 +58,10 @@ impl GlobalConfig {
 
     pub fn handlebars(&self) -> &handlebars::Handlebars<'static> {
         &self.handlebars
+    }
+
+    pub fn run_id(&self) -> &str {
+        &self.run_id
     }
 
     pub fn minimum_rustc_version(&self) -> &semver::Version {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -578,6 +578,12 @@ note: skipped the following crates since they have no library target: {skipped}"
                     .prepare_generator(config)
                     .map_err(|err| log_terminal_error(config, err))?;
 
+                let witness_data = witness_gen::WitnessGenerationData::new(
+                    baseline_loader.get_data_request(),
+                    current_loader.get_data_request(),
+                    current_loader.get_target_root(),
+                );
+
                 let data_storage = generate_crate_data(
                     config,
                     generation_settings,
@@ -593,6 +599,7 @@ note: skipped the following crates since they have no library target: {skipped}"
                     self.release_type,
                     &selected.overrides,
                     &self.witness_generation,
+                    witness_data,
                 )?;
                 config.shell_status(
                     "Finished",
@@ -753,9 +760,9 @@ pub struct WitnessGeneration {
     /// Whether to print witness hints, short examples that show why a change is breaking,
     /// while not necessarily buildable standalone programs.  See [`Witness::hint_template`].
     pub show_hints: bool,
-    /// Optional directory to write full witness examples to.  If this is `None`, full witnesses
-    /// will not be generated.  See [`Witness::witness_template`].
-    pub witness_directory: Option<PathBuf>,
+    /// Whether to generate witness programs, longer, fully valid and compilable examples
+    /// of a breaking change. See [`Witness::witness_template`] and [`Witness::witness_query`].
+    pub generate_witnesses: bool,
 }
 
 impl WitnessGeneration {
@@ -765,7 +772,7 @@ impl WitnessGeneration {
     pub const fn new() -> Self {
         Self {
             show_hints: false,
-            witness_directory: None,
+            generate_witnesses: false,
         }
     }
 }
--- a/src/witness_gen.rs
+++ b/src/witness_gen.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::{BTreeMap, btree_map},
-    path::Path,
+    fs,
+    path::{Path, PathBuf},
     sync::Arc,
     time::Duration,
 };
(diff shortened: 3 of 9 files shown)


GlitchlessCode:
I really don't know how everything just decided to up and stop working, I'm gonna figure this out, then take it out of draft.

GlitchlessCode:
I figured it out, this'll take a little bit. Turns out I wasn't paying attention when `LintResults` got added to the codebase.

GlitchlessCode:
Oh, and that the witnesses were added to tests
I'll work on fixing that up tomorrow

GlitchlessCode:
There, that seems fixed now!

src/lib.rs
```diff
@@ -753,9 +760,9 @@ pub struct WitnessGeneration {
     /// Whether to print witness hints, short examples that show why a change is breaking,
     /// while not necessarily buildable standalone programs.  See [`Witness::hint_template`].
     pub show_hints: bool,
-    /// Optional directory to write full witness examples to.  If this is `None`, full witnesses
-    /// will not be generated.  See [`Witness::witness_template`].
-    pub witness_directory: Option<PathBuf>,
+    /// Whether to generate witness programs, longer, fully valid and compilable examples
```
obi1kenobi:
```suggestion
    /// Whether to generate witness programs: longer, fully valid and compilable examples
```

src/witness_gen.rs
```diff
@@ -12,9 +13,32 @@ use trustfall::{FieldValue, TransparentValue};
...
 use crate::query::{Witness, WitnessQuery};
 use crate::{GlobalConfig, SemverQuery};
 
+/// Higher level data used specifically in the generation of a witness program. Values of [`None`] will
+/// prevent witness generation, since the data is required for witness generation.
+pub(crate) struct WitnessGenerationData<'a> {
+    baseline: Option<&'a CrateDataRequest<'a>>,
+    current: Option<&'a CrateDataRequest<'a>>,
+    target_dir: Option<&'a Path>,
+}
```
obi1kenobi, line 26: Out of curiosity, would these ever realistically be `None`? At this point you probably remember this code better than I do :)

src/witness_gen.rs
```diff
@@ -146,16 +170,74 @@ fn map_to_witness_text<'query>(
     }
 }
 
+/// Generates a single witness crate
+fn generate_witness_crate(
+    witness_set_dir: &Path,
+    witness_name: &str,
+    index: usize,
+    _witness_text: String,
```
obi1kenobi: Possibly `&str` here instead? I don't think we care about having ownership particularly strongly.

test_outputs/integration_snapshots__z_help.snap
```diff
@@ -22,7 +22,7 @@ Unstable options:
       --witness-hints
           Enable printing witness hints, examples of potentially-broken downstream code
 
-      --witnesses <OUTPUT_DIR>
+      --witnesses
           Enable generating and testing witness programs, full examples of potentially-broken downstream code
```
obi1kenobi, line 26:
```suggestion
          Enable generating and testing witness programs: full examples of potentially-broken downstream code
```

src/witness_gen.rs
```diff
@@ -146,16 +170,74 @@ fn map_to_witness_text<'query>(
...
+        .flat_map(|(semver_query, witness_texts)| {
+            witness_texts
+                .into_iter()
+                .enumerate()
+                .map(|(index, witness_text)| {
+                    generate_witness_crate(&witness_set_dir, &semver_query.id, index, witness_text?)
+                })
+                // This collect is necessary to convert the above synchronous Iter into a ParIter
+                .collect_vec()
+        });
```
obi1kenobi: It looks like you deleted the `.for_each()`, so now this just constructs an iterator, assigns it, and never uses it. Iterators are lazy, so none of this code actually runs — even though it type-checks.

src/config.rs
```diff
@@ -37,6 +42,13 @@ impl GlobalConfig {
         Self {
             level: None,
             handlebars: make_handlebars_registry(),
+            run_id: LazyLock::new(|| {
+                let rng = rand::rng();
+                rng.sample_iter(rand::distr::Alphanumeric)
+                    .take(16)
```
obi1kenobi: 16 might be a bit too much. Try generating a bunch of witnesses in a directory from a few different runs and see what the directory feels like. Are you able to find a given crate from a given run easily? Are all the prefixes helpful, or do they cross into being annoyingly long?

`rand::distr::Alphanumeric` has 62 options, which is about 2^6. The probability of collision here follows the [Birthday Paradox setup](https://en.wikipedia.org/wiki/Birthday_problem), and we'd probably be quite happy with somewhere in the region of 1-in-100k / 1-in-1M chance of collision. If my math is correct, ~6 chars should be plenty for that, though it's late and I'm tired so it's probably best if you check my math :)

obi1kenobi (commented):
Looks pretty good! Excited to have you back :)

src/witness_gen.rs
GlitchlessCode: They can be `None` if the `StatefulRustdocGenerator` is wrapping a `RustdocSource` (basically, if the source is a rustdoc file, it doesn't have some of these values). However, the current and baseline won't always *both* be `None`, which is why I didn't just wrap the whole struct in an `Option`.

src/witness_gen.rs
GlitchlessCode: Good point, I'll make sure to consume the Iterator. I plan to collect it in the end anyways for error reporting, so that won't be an issue at that point

src/config.rs
GlitchlessCode: I felt that 16 might be too much, I was just starting with a safe number. Thanks for pointing that out!

src/witness_gen.rs
obi1kenobi: Neat. Would you mind mentioning that in a comment?

> However, the current and baseline won't always _both_ be `None`

What if both the current and the baseline are `RustdocSource` i.e. just rustdoc JSON files? Say if `--current-rustdoc` and `--baseline-rustdoc` are used together.

src/witness_gen.rs
obi1kenobi: For maintainability (via "obviousness"), it's usually preferable not to rely on _causing_ side effects as a side effect of `.collect()`.

If we're generating and then checking witness crates (a substantial side effect!), let's try to do so in an obvious fashion, in a `.for_each()` or similar?

src/witness_gen.rs
GlitchlessCode: *Then* everything would be `None`, if both are Rustdoc sources.

src/witness_gen.rs
GlitchlessCode: Ahh yeah, fair enough. Good point on that convention. I'll think more on how best to re-adapt this.

src/witness_gen.rs
obi1kenobi: Ah, I mis-parsed your previous message. That's also what I thought. Carry on and don't mind me :)

src/config.rs
```diff
@@ -45,7 +45,7 @@ impl GlobalConfig {
             run_id: LazyLock::new(|| {
                 let rng = rand::rng();
                 rng.sample_iter(rand::distr::Alphanumeric)
-                    .take(16)
+                    .take(6)
```
obi1kenobi, line 58: nit: please leave a note in a comment about the math we did here

If we need to change it in the future, it'll be much easier to tweak the existing math than to re-do it from scratch. It'll take 20s to write and will save us much more than that :)

src/witness_gen.rs
```diff
@@ -186,6 +186,22 @@ fn generate_witness_crate(
...
 }
 
+fn run_single_witness_check(
+    witness_set_dir: &Path,
+    witness_name: &str,
+    index: usize,
+    witness_text: &str,
+    // TODO: Make this Result<WitnessResult>, such that Err means that we experienced some non-witness-checking error
+    // whereas Ok means that the Witness was succesfully checked, and WitnessResult contains the information on wether
+    // it was successful or failed. Currently just leaving it as Result<()> as a placeholder
```
obi1kenobi: This is also fine, but what you could do as well is make the signature `Result<Result<(), ()>>` and leave something like this as a doc comment on the function, explaining what each of the possible values means. Then later we can fill in the inner success and failure types.

src/witness_gen.rs
```diff
@@ -186,20 +186,24 @@ fn generate_witness_crate(
     Ok(crate_path)
 }
 
+/// Runs a single witness check, returning the result of the witness process
+///
+/// Err implies an error during the generation or checking of the witness
+/// Ok implies the witness was successfully generated and checked. The internal Result
+/// indicates if the witness was successful, or if it failed.
+///
+/// ## Note: The unit types in the return type are temporary placeholders until the witness system is complete
```
obi1kenobi, line 195: Suggested rewording:
```suggestion
/// Runs a single witness check, returning the outcome of the witness process.
///
/// The outer `Result` captures the outcome of the witness-generation process.
/// Failures to check a witness such as "out of disk space" or "cargo not found"
/// would be reflected in a top-level `Result::Err` outcome.
///
/// The inner `Result` shows the outcome of checking the witness itself.
///
/// **Note:** The unit types in the return type are temporary placeholders until the witness system is complete.
```

Lack of punctuation is often okay, but it really hurt here. I originally misread this as
```
Err implies an error during the generation or checking of the witness
or implies the witness was successfully generated and checked.
```
which says something very different, even though it's a one-character misread.

I'd recommend practicing using structure (punctuation, newlines, etc.) a bit more intentionally. For example, here `## Note` is an `h2` (second-level heading) element when displayed, which will show up in huge bold font. But that line seems a lot less important _on a private function_ than knowing what the inner and outer result mean. Similarly, `Err` and `Ok` without context could refer to either the inner or the outer `Result`, which is grounds for confusion.

obi1kenobi: I'll merge this PR as-is, without applying my suggestion, so as to keep this moving forward.

If you agree with the suggestion, please open a new PR to apply the change?

Issue #1421, opened 2025-09-04
Custom rendering of lint reports


kornelski:
### Describe your use case

Currently cargo-semver-checks only emits results as a text formatted for a terminal. However, there are environments that support more than plain text and could display a richer UI for the errors. For example, [GitHub Actions annotations](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands) allow attaching error messages to source code view. GitLab supports junit reports and can post Markdown-formatted comments in pull requests.

If cargo-semver-checks exposed reports in a machine-readable format, the reports could be reformatted for other kinds of UIs. 

### Describe the solution you'd like

The library API of cargo-semver-checks could add more details to [`CrateReport`](https://docs.rs/cargo-semver-checks/latest/cargo_semver_checks/struct.CrateReport.html). The CLI version could emit it as JSON.

obi1kenobi:
Broadly speaking, I'd love to have JSON output and a richer library interface. But within the current constraints, that's beyond my ability to support with the financial support the project is getting at the moment. I'm already at the maintenance time limit, and I can't reasonably allocate time to support such a large expansion of the surface area of the tool at the moment.

I'm happy to gratefully accept refactoring PRs in the spirit of #1422 🙏 But in general, I can't afford to merge or publish functionality that increases my maintenance burden until we have more sources of stable and recurring corporate funding for the project. Otherwise I'd just end up burned out and abandoning the project, which isn't in anyone's best interest. As you're also a prolific OSS maintainer, I'm sure you understand the predicament all too well :)

PR #1422, opened 2025-09-04
Separate report printing from generation

kornelski:
For #1421

This doesn't add any public APIs yet, only avoids printing as a side effect.


obi1kenobi (approved):

obi1kenobi:
Looks great, thank you!

PR #1471, opened 2025-10-14
Load rustflags/rustdocflags from cargo config, then augment them.

obi1kenobi:
Ensure we can correctly build crates that specify `--cfg` or other required configuration in their `.cargo/config.toml` directories.


src/data_generation/generate.rs
```diff
@@ -166,6 +168,156 @@ enum CargoUpdateResult {
...
+/// Determine the default target triple configured in the environment.
+fn get_default_build_target_triple(config: &cargo_config2::Config) -> anyhow::Result<String> {
+    // If `CARGO_BUILD_TARGET` is set, it dominates other options.
+    if let Ok(value) = std::env::var("CARGO_BUILD_TARGET") {
+        return Ok(value);
+    }
+
+    // `build.target` from (merged) cargo config following its precedence rules is next.
+    //
+    // Cargo support multiple build targets, but `cargo-semver-checks` does not.
```
obi1kenobi:
```suggestion
    // Cargo supports multiple build targets, but `cargo-semver-checks` does not.
```

Issue #1177, opened 2025-03-10
Tests fail on windows due to writing a different path separator


CLIDragon:
Running `cargo test` on Windows causes test failures due to the `span_filename` field being written as `src\\lib.rs` instead of `src/lib.rs`. One example is in the screenshot included below. 
![Image](https://github.com/user-attachments/assets/f19eae51-711b-4d89-bd47-e6affcf7bf3f)

I suspect this is because rust's default path handling works in a platform-dependent way and it should be easily fixable - see e.g. the [typed-path](https://docs.rs/typed-path/latest/typed_path) crate.

obi1kenobi:
Thanks for opening the issue. I'm unfortunately not convinced it's that easy of a fix, but it's possible I've missed something. If you think it's both an easy fix and cost-effective to fix (meaning, it offers the best positive impact relative to how much time it'll take to fix) then feel free to go for it.

The `PathBuf` we show there originates from the `rustdoc-types` crate's [`Span` type](https://docs.rs/rustdoc-types/latest/rustdoc_types/struct.Span.html), and comes to `cargo-semver-checks` via [the `VertexKind::Span` variant in `trustfall-rustdoc-adapter`](https://github.com/obi1kenobi/trustfall-rustdoc-adapter/blob/26c3d4009640f4e446064fb32bf32aa6fc935c80/src/adapter/vertex.rs#L36).

Just a note that ensuring `cargo test` continues to pass on Windows would likely also require setting up Windows `cargo test` in CI, which isn't trivial. GitHub's Windows runners are relatively underpowered, so compilation takes a long time and will need a decent bit of work to make fast enough to not be a drag on development. The [`uv` project](https://github.com/astral-sh/uv) has done quite a bit of work on making their Windows CI builds fast, so that's probably a good source of ideas we should consider adopting in our own CI.

obi1kenobi:
Also, just to avoid any doubt with folks just coming across this issue: `cargo-semver-checks` already has end-to-end Windows tests in CI that ensure the tool runs correctly. You can semver-checks your packages on Windows just fine.

This issue is pointing out that the unit test suite isn't portable across operating systems. This is unfortunate, but not a significant correctness concern since we aren't testing any OS-related functionality. If an OS-specific divergence in behavior happens, it's almost certainly a bug in rustdoc, cargo, or another Rust tool and not in `cargo-semver-checks`.

PR #1497, opened 2025-11-06
Remove expression from `insta` output in tests.

CLIDragon:
**Purpose of Change**
Reduce test runtime.

**Describe the change**
Do not generate the expression in `insta` anymore. Since generating the expression requires waiting for a lock on the source file the generation takes ~10-15% of the test runtime, but doesn't provide meaningful information the test name doesn't.

### DIFF
--- a/test_outputs/query_execution/pub_api_sealed_trait_method_receiver_mut_ref_became_ref.snap
+++ b/test_outputs/query_execution/pub_api_sealed_trait_method_receiver_mut_ref_became_ref.snap
@@ -1,7 +1,6 @@
 ---
 source: src/query.rs
 assertion_line: 1048
-expression: "&query_execution_results"
 ---
 {
   "./test_crates/trait_method_receiver_sealedness/": [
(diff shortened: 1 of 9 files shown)


obi1kenobi:
Nice find! Out of curiosity, does this require a change in the snapshot output files? Because I didn't see it in the PR, and we probably should update them if the output is going to look any different (even if the current files work fine).

CLIDragon:
It looks like it will. Unfortunately due to #1177 I'm not able to generate the correct snapshots on Windows and I don't have a linux environment set up right now.

obi1kenobi:
I regenerated the snapshots on my machine and pushed them to the upstream `insta-remove-expression` branch. If that looks good, feel free to cherry-pick my commit to your branch and update the PR — I'll be happy to merge it.

CLIDragon:
LGTM. `snapshot_kind: text` is also removed, but that's due to an `insta`-wide change (https://github.com/mitsuhiko/insta/issues/676).

CLIDragon:
I updated the snapshots for the new lints manually, so this should be good to merge.

obi1kenobi:
Awesome, thank you!

PR #1515, opened 2025-11-24
Add lint trait_must_use_removed

Nalin-khanna:
This PR is part of the issue #1279


test_crates/trait_must_use_removed/new/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[must_use]
+pub trait MustUseMessageTraitToMustUseTrait {}
+
+#[must_use = "Baz"]
+pub trait MustUseMessageTraitToMustUseMessageTrait {}
+
+// This trait is private and should NOT be reported by this rule.
+
+trait MustUsePrivateTrait {}
```
obi1kenobi: nitpick: it's nice to end files in a single trailing newline, to improve compatibility with UNIX tooling that expects all lines end in a newline character
```suggestion
trait MustUsePrivateTrait {}

```

If you use VSCode or similar editors, the following settings are good to enable:
```
```

test_crates/trait_must_use_removed/old/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+// This trait is private and should NOT be reported by this rule.
+
+#[must_use]
+trait MustUsePrivateTrait {}
+
+// This trait existed in the old version and is removed in the new one.
+// It should NOT be reported by this rule to avoid duplicate lints.
+
+#[must_use]
+pub trait MustUseRemovedTrait {}
```
obi1kenobi: same nit as above
```suggestion
pub trait MustUseRemovedTrait {}

```

test_crates/trait_must_use_removed/new/src/lib.rs
Nalin-khanna: Got it ! Thanks for helping out

PR #1523, opened 2025-12-18
Make the -q command-line option useful.

geeknoid:
As of version 0.45.0, using the -q option on the
command-line nearly completely eliminates all console output, which makes it not particularly useful.

This PR makes it so when using the -q option,
any lint violations are still printed to the console. This is the mode I'd like to use in our CI system.


obi1kenobi:
Thanks for looking into this. Could you share some before-and-after comparisons that demonstrate the difference in behavior, to make this PR easier to review?

geeknoid:
Well, I ran checks before with the -q option and got basically no output from the tool, whether or not there were warnings:

```
cargo semver-checks -q
```

With my change, doing this gives you:

image

The above screen shot is taken from this PR: https://github.com/microsoft/oxidizer/pull/157. In this PR, you'll see the horrible regex I'm running to achieve the above output. With the changes I'm proposing here, I can remove those horrible regexes and just use -q.

src/check_release.rs
```diff
@@ -137,85 +137,138 @@ fn print_triggered_lint(
...
-                    "{}\n",
-                    serde_json::to_string_pretty(&pretty_result)?
+                    "{}Failed in:{}",
+                    Style::new().bold(),
+                    Reset
                 )?;
                 Ok(())
             })?;
         }
+        LintLevel::Allow => return Ok(()), // Don't print allowed lints
```
obi1kenobi, line 209: This should be unreachable, it's pointless to run a lint whose level is `Allow`.
```suggestion
        LintLevel::Allow => {
            unreachable!("attempted to print lint whose level is `Allow`: {lint_result:?}");
        }
```

src/check_release.rs
```diff
@@ -137,85 +137,138 @@ fn print_triggered_lint(
...
+                        config.stdout(),
+                        "{}Description:{}\n{}\n{:>12} https://github.com/obi1kenobi/cargo-semver-checks/tree/v{}/src/lints/{}.ron",
+                        Style::new().bold(),
+                        Reset,
+                        &semver_query.error_message,
+                        "impl:",
+                        crate_version!(),
+                        semver_query.id,
+                    )?;
+                }
```
obi1kenobi, line 199: Can we avoid the duplication between the `log_error` and `log_warn` cases, both here and below? This feels like it would regress maintainability a decent bit, since this code is not trivial.

obi1kenobi:
Just checking if you're still interested in getting this PR over the finish line! I'm planning to cut a new release in the next few days :)

geeknoid:
Sorry, I've been pretty busy lately and I'm headed for vacation for a while. So please go ahead, and I'll get it next time around.

obi1kenobi:
Np, enjoy your vacation!

PR #1538, opened 2025-12-25
Exhaustive struct added

Nalin-khanna:
Addresses issue #949 

Lint for - adding a new public API struct that is not #[non_exhaustive] and has no non-public fields

### DIFF
--- a/test_outputs/query_execution/exhaustive_struct_added.snap
+++ b/test_outputs/query_execution/exhaustive_struct_added.snap
@@ -210,16 +210,6 @@ source: src/query.rs
       "visibility_limit": String("public"),
     },
     {
-      "name": String("UnionToStructDocHiddenField"),
-      "path": List([
-        String("union_changed_to_incompatible_struct"),
-        String("UnionToStructDocHiddenField"),
-      ]),
-      "span_begin_line": Uint64(22),
-      "span_filename": String("src/lib.rs"),
-      "visibility_limit": String("public"),
-    },
-    {
       "name": String("UnionWithMultiplePubFields"),
       "path": List([
         String("union_changed_to_incompatible_struct"),


src/lints/exhaustive_struct_added.ron
```diff
@@ -0,0 +1,56 @@
...
+                        visibility_limit @filter(op: "=", value: ["$public"]) @output
+                        name @output
+
+                        attrs @filter(op: "not_contains", value: ["$non_exhaustive"])
+
+                        # Ensure the struct can be constructed outside of its crate
+                        # using a struct literal: it does not have any private fields.
+                        field @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
+                            visibility_limit @filter(op: "!=", value: ["$public"])
+                        }
```
obi1kenobi, line 23: I think we want the following block in there as well:
```suggestion
                        field @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
                            visibility_limit @filter(op: "!=", value: ["$public"])
                        }
                        
                        # Ensure that all fields are public API.
                        # Otherwise constructing the struct with a literal is not public API either.
                        field @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
                            public_api_eligible @filter(op: "!=", value: ["$true"])
                        }
```

Otherwise structs like the following will be reported as public API and exhaustive when they are not really:
```rust
pub struct Example {
    pub a: i64,

    #[doc(hidden)]
    pub b: i64,
}
```

src/lints/exhaustive_struct_added.ron
```diff
@@ -0,0 +1,56 @@
...
+
+                        importable_path {
+                            path @output @tag
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
```
obi1kenobi, line 34: It might be a good idea to lift this block above the `field` ones for slightly better query performance, since it might be more selective and would filter out more structs earlier.

test_crates/exhaustive_struct_added/new/src/lib.rs
```diff
@@ -0,0 +1,27 @@
+#![no_std]
+
+// Existing struct should not trigger the lint
+pub struct ExistingStruct {
+    pub bar: u64,
+}
+
+// New non-exhaustive struct with public fields should trigger the lint
+pub struct NewStruct {
+    pub bar : u64,
```
obi1kenobi: nit: please try to run `cargo fmt` on the test crates
```suggestion
    pub bar: u64,
```

test_crates/exhaustive_struct_added/new/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+// This struct have non-public fields , so it should not trigger the lint.
+pub struct StructWithPrivateField {
+    foo : u64
+}
+
+// This struct is #[non_exhaustive], so it should not trigger this lint
+#[non_exhaustive]
+pub struct NonExhaustiveStruct {
+    bar: u64,
+}
```
obi1kenobi, line 27:
```suggestion
}

// This struct shouldn't trigger the lint because it cannot be constructed
// with a literal within the public API of the crate.
pub struct StructWithHiddenField {
    pub a: i64,

    #[doc(hidden)]
    pub b: i64,
}
```

obi1kenobi (approved):
Nice! Just a couple of nits that I'll fix up here.

If you'd like, feel free to open 3-4 PRs for new lints concurrently. You've clearly mastered the process for writing lints, and I usually find time to review PRs in batches, so having multiple PRs in flight is okay.

test_outputs/query_execution/exhaustive_struct_added.snap
```diff
@@ -0,0 +1,255 @@
...
+    {
+      "name": String("UnionToStructDocHiddenField"),
+      "path": List([
+        String("union_changed_to_incompatible_struct"),
+        String("UnionToStructDocHiddenField"),
+      ]),
+      "span_begin_line": Uint64(22),
+      "span_filename": String("src/lib.rs"),
+      "visibility_limit": String("public"),
+    },
```
obi1kenobi:
```suggestion
```

Nalin-khanna:
> Nice! Just a couple of nits that I'll fix up here.
> 
> If you'd like, feel free to open 3-4 PRs for new lints concurrently. You've clearly mastered the process for writing lints, and I usually find time to review PRs in batches, so having multiple PRs in flight is okay.

Thanks for fixing those up! I’ll open a few more PRs for the new lints as I work through them.

PR #1541, opened 2025-12-29
Non exhaustive struct added

Nalin-khanna:
#949  adding a new #[non_exhaustive] public API struct (one of two complements of the above lint)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,15 +33,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
-name = "android_system_properties"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
-dependencies = [
- "libc",
-]
-
-[[package]]
 name = "anstream"
 version = "0.6.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -216,15 +207,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "block2"
-version = "0.6.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cdeb9d870516001442e364c5220d3574d2da8dc765554b4a617230d33fa58ef5"
-dependencies = [
- "objc2",
-]
-
-[[package]]
 name = "borsh"
 version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


test_outputs/query_execution/non_exhaustive_struct_added.snap
```diff
@@ -0,0 +1,29 @@
...
+      "name": String("UnionToNonExhaustiveStruct"),
+      "path": List([
+        String("union_changed_to_incompatible_struct"),
+        String("UnionToNonExhaustiveStruct"),
+      ]),
+      "span_begin_line": Uint64(33),
+      "span_filename": String("src/lib.rs"),
+      "visibility_limit": String("public"),
+    },
+  ],
```
obi1kenobi: This is an interesting, tricky case: here a `union` became a `struct`. I believe that change will cause a SemVer breakage lint, and then this lint will also trigger as a duplicate, which is undesirable.

This happens because this lint says "the *struct* didn't exist" and ignores non-structs at the same import path. I'll change the `baseline` case to look at `ImplOwner` (struct/enum/union) instead of just `Struct` and the duplicate should go away.

obi1kenobi:
```suggestion
```

src/lints/non_exhaustive_struct_added.ron
```diff
@@ -0,0 +1,62 @@
...
+                        span_: span @optional {
+                            filename @output
+                            begin_line @output
+                        }
+                    }
+                }
+            }
+            baseline @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
+                item {
+                    ... on Struct {
```
obi1kenobi:
```suggestion
                    ... on ImplOwner {
```
See explanation below about the tricky duplicate lint issue.

src/lints/non_exhaustive_struct_added.ron
```diff
@@ -0,0 +1,62 @@
+SemverQuery(
+    id: "non_exhaustive_struct_added",
+    human_readable_name: "#[non_exhaustive] pub struct added",
+    description: "A new #[non_exhaustive] pub struct was added. It cannot be constructed using a struct literal outside of its defining crate, and pattern matching will require a wildcard.",
```
obi1kenobi: I think we usually try to avoid Rust syntax in these fields. Let's play it safe.
```suggestion
    human_readable_name: "non-exhaustive pub struct added",
    description: "A new non-exhaustive pub struct was added. It cannot be constructed using a struct literal outside of its defining crate, and pattern matching will require a wildcard.",
```

test_outputs/query_execution/non_exhaustive_struct_added.snap
Nalin-khanna: Thank you for the review! 
I initially thought we should show both lints in the case when an `ImplOwner` changes to a `struct`, but this makes more sense. I was also wondering why it didn’t show up for the `enum` changing to a `struct` case.

obi1kenobi: It's likely we just didn't have a test case where the enum became a non-exhaustive struct specifically. I think for unions it sometimes matters whether the struct is exhaustive or not, but changing the enum to a struct is likely always breaking.

PR #1543, opened 2026-01-03
Add two new lints regarding additions of structs.

Nalin-khanna:
issue #949 
adding a new public API struct that has a non-public field and no #[non_exhaustive]

Here, I have considered doc-hidden fields as non-public.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -105,9 +105,9 @@ checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "assert_cmd"
-version = "2.1.1"
+version = "2.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bcbb6924530aa9e0432442af08bbcafdad182db80d2e560da42a6d442535bf85"
+checksum = "9c5bcfa8749ac45dd12cb11055aeeb6b27a3895560d60d71e3c23bf979e60514"
 dependencies = [
  "anstyle",
  "bstr",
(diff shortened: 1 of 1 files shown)


src/lints/exhaustive_struct_with_private_fields_added.ron
```diff
@@ -0,0 +1,57 @@
+SemverQuery(
+    id: "exhaustive_struct_with_private_fields_added",
+    human_readable_name: "exhaustive pub struct with private fields added",
+    description: "A new pub struct with private fields was added. Adding fields to this struct in the future will break pattern matching.",
```
obi1kenobi: This isn't quite accurate. The only pattern matching that is within public API is matching that includes a `..` term to encompass non-public-API fields. Naming `#[doc(hidden)]` fields in a pattern match is use of non-public-API and breaking it is fine, while if the fields are not `pub` then the compiler will force use of `..` just like with `#[non_exhaustive]`.

Nalin-khanna:
apologies for the delay! just got done with my uni exams .
I get the distinction and need for separate lints . 
l will go ahead and implement the split into two lints `exhaustive_struct_with_private_fields_added` and `exhaustive_struct_with_doc_hidden_fields_added` and push the updates to this PR shortly.

obi1kenobi:
> apologies for the delay! just got done with my uni exams .

No need to apologize! Thank you for your contributions, and congratulations on being done with exams! I hope they went well.

I didn't get a chance to review this in depth tonight, but I look forward to looking at it ASAP. We'll definitely get this merged before the next `cargo-semver-checks` release which is (aspirationally!) planned for this weekend 🤞

src/lints/exhaustive_struct_with_private_fields_added.ron
```diff
@@ -0,0 +1,57 @@
...
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "zero": 0,
+        "true": true,
+        "non_exhaustive": "#[non_exhaustive]",
+    },
+    error_message: "A new exhaustive pub struct with non-public fields was added.",
```
obi1kenobi:
```suggestion
    error_message: "A new pub struct with non-public fields was added.",
```
Just to avoid possibly confusing the user a bit. Why exhaustiveness is relevant here is something we know, but they don't.

obi1kenobi (approved):
Excellent work as always!

PR #1549, opened 2026-01-17
New lint: struct_no_private_fields_no_longer_non_exhaustive

GrigorenkoPV:
- As suggested in #949
- Allow-by-default
- Name is bikesheddable


GrigorenkoPV:
Sure, makes sense. Closing in favor of #1550 then.

PR #1550, opened 2026-01-18
Add minor lint for removing #[non_exhaustive] from public struct

connor15mcc:
(with the struct having all public fields)

This makes the struct constructible via struct literal, which may or may not have been intended.

An example of something that should trigger this lint:

```diff
-#[non_exhaustive]
pub struct MyStruct {
    pub field: i32,
}
```

Context:
[Unchecked bullet in list of opt-in minor lints](https://github.com/obi1kenobi/cargo-semver-checks/issues/949)

I did my best also to follow the patterns in a similar lint addition (for enums): https://github.com/obi1kenobi/cargo-semver-checks/pull/1186

---

edit: my apologies, I'm just seeing that https://github.com/obi1kenobi/cargo-semver-checks/pull/1549 was recently opened too. I'm happy to close my PR in favor of that once since it was first -- I can take a look through some other open issues!

### DIFF
--- a/src/lints/struct_no_longer_non_exhaustive.ron
+++ b/src/lints/struct_no_longer_non_exhaustive.ron
@@ -1,0 +1,66 @@
+SemverQuery(
+    id: "struct_no_longer_non_exhaustive",
+    human_readable_name: "struct no longer #[non_exhaustive]",
+    description: "A pub struct is no longer #[non_exhaustive]. It can now be constructed using a struct literal outside its crate.",
+    required_update: Minor,
+    lint_level: Allow,
+    reference_link: Some("https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute"),
+    query: r#"
+    {
+        CrateDiff {
+            current {
+                item {
+                    ... on Struct {
+                        visibility_limit @filter(op: "=", value: ["$public"]) @output
+                        name @output
+                        struct_type @output
+
+                        attrs @filter(op: "not_contains", value: ["$non_exhaustive"])
+
+                        # ensure all fields are public (no private fields).
+                        field @fold @transform(op: "count") @filter(op: "=", value: ["$zero"]) {
+                            visibility_limit @filter(op: "!=", value: ["$public"])
+                        }
+
+                        importable_path {
+                            path @output @tag
+                            public_api @filter(op: "=", value: ["$true"])
+                        }
+
+                        span_: span @optional {
+                            filename @output
+                            begin_line @output
+                            end_line @output
+                        }
+                    }
+                }
+            }
+            baseline {
+                item {
+                    ... on Struct {
+                        visibility_limit @filter(op: "=", value: ["$public"])
+                        attrs @filter(op: "contains", value: ["$non_exhaustive"])
+
+                        # ensure all fields are public (no private fields).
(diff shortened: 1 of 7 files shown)


src/lints/struct_no_longer_non_exhaustive.ron
```diff
@@ -0,0 +1,61 @@
+SemverQuery(
+    id: "struct_no_longer_non_exhaustive",
+    human_readable_name: "struct no longer #[non_exhaustive]",
+    description: "A pub struct is no longer #[non_exhaustive]. It can now be constructed using a struct literal outside its crate.",
```
obi1kenobi, line 4: > It can now be constructed using a struct literal outside its crate.

This isn't necessarily true: what if the struct lost `#[non_exhaustive]` but gained a non-`pub` field? The query checks that in the _baseline_ there weren't any non-`pub` fields, but the "current" side might have gained one in the meantime.

I'm now realizing that we might be better off framing these lints differently: the thing to flag is that the struct is constructible with a literal — by virtue of losing `#[non_exhaustive]` in this lint, or by virtue of losing all its non-`pub` fields in a separate lint, while paying close attention to ensure we cover all possible cases without any overlap between lints.

My proposal is this: let's keep this lint as the one for `#[non_exhaustive]` removal and add the check that the "current" side has no non-`pub` fields. Then let's add a separate lint for having and losing all non-`pub` fields and becoming constructible that way, regardless of `#[non_exhaustive]`. I *think* those two don't overlap? Please double-check my thinking here :)

src/lints/struct_no_longer_non_exhaustive.ron
```diff
@@ -0,0 +1,61 @@
...
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "non_exhaustive": "#[non_exhaustive]",
+        "true": true,
+        "zero": 0,
+    },
+    error_message: "A public struct is no longer #[non_exhaustive]. It can now be constructed using a struct literal outside its crate, which previously wasn't possible.",
```
obi1kenobi: See if you like this wording tweak, which seems to be a bit of a shorter way to say the same thing:
```suggestion
    error_message: "A public struct is no longer #[non_exhaustive]. It can newly be constructed using a struct literal outside its crate.",
```

test_crates/struct_no_longer_non_exhaustive/new/src/lib.rs
obi1kenobi, line 1: Please make sure to add some test cases that cover the addition of the new clauses in the query.

src/lints/struct_no_longer_non_exhaustive.ron
connor15mcc: Oh this is interesting, thanks for calling that out. I guess I didn't appreciate that possibility from the baseline -> current comparisons, but it's clear once you've mentioned.

Let me mull it over a bit and I'll try a second pass, hopefully this weekend. I also _think_ those should be disjoint cases, but will pay closer attn when writing + testing.

connor15mcc: I've added the requisite check + tests here, and broken the new lint rule off into a separate PR: https://github.com/obi1kenobi/cargo-semver-checks/pull/1562

I've been building + testing with both applied and agree they shouldn't overlap, covering all the cases that occurred to me.

PR #1551, opened 2026-01-18
Add new minor lint for adding an inherent associated `const` to public API

connor15mcc:
An opt-in lint for accidental additions to a project's public API via new inherent associated const types. For example:

```diff
pub struct Foo;

impl Foo {
+    pub const MY_CONST: i32 = 42;
}
```

Note: I think my naming of the lint could probably use a little massaging.. I'm not sure the best way to communicate this scenario.

Context: "public API type adding a new public associated const item" https://github.com/obi1kenobi/cargo-semver-checks/issues/949

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -264,7 +264,7 @@ jobs:
         #
         #            Also make sure to update the MSRV in the cargo-semver-checks-action CI:
         #            https://github.com/obi1kenobi/cargo-semver-checks-action/blob/main/.github/workflows/test-action.yml#L18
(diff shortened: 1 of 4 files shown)


src/lints/inherent_associated_pub_const_added.ron
```diff
@@ -0,0 +1,66 @@
...
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+        "zero": 0,
+    },
+    error_message: "A public type's added a new public associated const item",
```
obi1kenobi:
```suggestion
    error_message: "A new public associated const item was added to an existing type.",
```

PR #1562, opened 2026-01-25
Add minor lint for losing all non-`pub` fields

connor15mcc:
An example would simply be:

```diff
pub struct MyStruct {
- field: i32,
}
```

A follow up from: https://github.com/obi1kenobi/cargo-semver-checks/pull/1550/files#r2719626142

### DIFF
--- a/src/lints/struct_no_longer_has_non_pub_fields.ron
+++ b/src/lints/struct_no_longer_has_non_pub_fields.ron
@@ -1,0 +1,68 @@
+SemverQuery(
+    id: "struct_no_longer_has_non_pub_fields",
+    human_readable_name: "struct no longer has non-pub fields",
+    description: "A (exhaustive) pub struct has lost all its non-pub fields and can now be constructed using a struct literal",
+    required_update: Minor,
+    lint_level: Allow,
+    reference_link: Some("https://doc.rust-lang.org/cargo/reference/semver.html#item-add"),
+    query: r#"
+    {
+        CrateDiff {
+            current {
+                item {
+                    ... on Struct {
+                        visibility_limit @filter(op: "=", value: ["$public"]) @output
+
+                        importable_path {
+                            path @output @tag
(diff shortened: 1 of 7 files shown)


src/lints/struct_no_longer_non_pub_fields.ron
```diff
@@ -0,0 +1,68 @@
+SemverQuery(
+    id: "struct_no_longer_non_pub_fields",
```
obi1kenobi: This reads a bit awkwardly. Perhaps `struct_no_longer_has_non_pub_fields` instead?

If you agree, consider using the `scripts/rename_lint.sh` script — it might save you some tedium :)

src/lints/struct_no_longer_non_pub_fields.ron
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "non_exhaustive": "#[non_exhaustive]",
+        "true": true,
+        "zero": 0,
+    },
+    error_message: "A pub struct has lost all its non-pub fields and can now be constructed using a struct literal",
```
obi1kenobi:
```suggestion
    error_message: "A pub struct has lost all its non-pub fields and can now be constructed using a struct literal.",
```

src/lints/struct_no_longer_non_pub_fields.ron
connor15mcc: That worked a treat -- quite lovely!

connor15mcc: (agreed on the name change, thanks for the nudge)

connor15mcc:
Should be all renamed! And certainly, thank _you_ for maintaining such a neat tool.

Let me know if there are any particular issues you think could use a fresh set of eyes! No promises how quickly I'll look, but happy to contribute as I'm able.

obi1kenobi (approved):
Nice work!

PR #1571, opened 2026-02-10
feat: add minor lint `inherent_method_added`

Ansh-699:
## Summary

Adds a new **opt-in minor lint** `inherent_method_added` that detects when a public type gains a new public inherent method. This is an API addition that warrants a minor version bump.

Part of #949 — "Opt-in minor lints for added functionality" (specifically the unchecked item: "Adding a new method to a type's own `impl` block (inherent method)").

## What the lint detects

The lint fires when a public type (struct / enum / union) gains a new public method in its inherent `impl` block — both instance methods (`&self`) and static methods (no receiver).

### Triggers (should fire)
- New public instance method on a public struct
- New public static method on a public struct
- New public method on a public enum

### Does NOT trigger
- Private methods added to a public type
- Methods added to a private type
- Methods that already existed in the baseline

## Implementation

**Modeled after** `inherent_associated_pub_const_added` — the query structure is identical, just swapping `associated_constant` for `method`.

- `required_update: Minor` and `lint_level: Allow` (opt-in, per #949 requirements)
- Uses `public_api_eligible` filter to exclude non-public methods
- Baseline check: `@fold @transform(op: "count") @filter(op: "=", value: ["$zero"])` confirms the method did not exist in the baseline

## Files changed (7 files)

| File | Action |
|---|---|
| `src/lints/inherent_method_added.ron` | New lint definition |
| `src/query.rs` | Register lint in `add_lints!()` macro |
| `test_crates/inherent_method_added/old/Cargo.toml` | Baseline test crate manifest |
| `test_crates/inherent_method_added/old/src/lib.rs` | Baseline code |
| `test_crates/inherent_method_added/new/Cargo.toml` | Current test crate manifest |
| `test_crates/inherent_method_added/new/src/lib.rs` | Code with new methods |
| `test_outputs/query_execution/inherent_method_added.snap` | Snapshot |

## Verification

- `cargo test inherent_method_added` — ✅ 1 passed
- `cargo clippy --all-targets --no-deps -- -D warnings --allow deprecated` — ✅ clean
- `cargo fmt --check` — ✅ clean
- Test crates compile: `RUSTFLAGS="-A dead_code -A unused" cargo check` — ✅ both old and new

### DIFF
--- a/src/lints/method_parameter_count_changed.ron
+++ b/src/lints/method_parameter_count_changed.ron
@@ -128,5 +128,5 @@ SemverQuery(
         "zero": 0,
     },
     error_message: "A publicly-visible method now takes a different number of parameters, not counting the receiver (self) parameter.",
-    per_result_error_template: Some("{{join \"::\" path}}::{{method_name}} now takes {{unpack_if_singleton current_parameter_count}} parameters instead of {{old_parameter_count}}, in {{multiple_spans non_matching_span_filename non_matching_span_begin_line}}"),
+    per_result_error_template: Some("{{join \"::\" path}}::{{method_name}} takes {{old_parameter_count}} parameters in {{span_filename}}:{{span_begin_line}}, but now takes {{unpack_if_singleton current_parameter_count}} parameters in {{multiple_spans non_matching_span_filename non_matching_span_begin_line}}"),
 )


src/lints/inherent_method_added.ron
```diff
@@ -0,0 +1,66 @@
...
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "true": true,
+        "zero": 0,
+    },
+    error_message: "A new public method was added to an existing public type's inherent impl.",
```
obi1kenobi: small wording tweak:
```suggestion
    error_message: "A new inherent public method was added to an existing public type.",
```

obi1kenobi (approved):
Thanks for the lint! Just like in the other PR, please ask your LLM of choice to be more succinct when writing the PR description — the vast majority of it is unnecessary.

PR #1576, opened 2026-02-16
Make -q command-line option show lint violations

geeknoid:
When using -q (quiet mode), lint violations are now printed using the appropriate log level based on severity: log_error for Deny-level lints and log_warn for Warn-level lints, instead of log_info which gets suppressed in quiet mode.

Adds is_warn(), log_warn(), and log_at_lint_level() methods to GlobalConfig. The log_at_lint_level() method dispatches to the correct log level based on LintLevel, avoiding code duplication between Deny and Warn cases. It panics on LintLevel::Allow since allowed lints should never be printed.

This is a redo of PR #1523


obi1kenobi:
Thanks!

PR #1579, opened 2026-02-22
Add minor lint: inherent_method_no_longer_unsafe

MarcosAsh:
Adds a new Minor/Allow lint that detects when a public inherent method or associated fn is no longer unsafe. From the tracking list in #949.

The query mirrors inherent_method_unsafe_added with the unsafe filters inverted, using the same 3-fold impl pattern to handle trait migration and method removal edge cases.

Test crate covers associated fn, self method, and enum associated fn becoming safe, plus negative cases for stays-unsafe, already-safe, removed-entirely, and private-struct.

### DIFF
--- a/src/lints/inherent_method_no_longer_unsafe.ron
+++ b/src/lints/inherent_method_no_longer_unsafe.ron
@@ -1,0 +1,105 @@
+SemverQuery(
+    id: "inherent_method_no_longer_unsafe",
+    human_readable_name: "pub method is no longer unsafe",
+    description: "A method or associated fn is no longer unsafe to call. Callers that use an `unsafe` block may get `unused_unsafe` warnings.",
+    required_update: Minor,
+    lint_level: Allow,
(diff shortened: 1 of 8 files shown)


src/lints/inherent_method_no_longer_unsafe.ron
```diff
@@ -0,0 +1,105 @@
+SemverQuery(
+    id: "inherent_method_no_longer_unsafe",
+    human_readable_name: "pub method is no longer unsafe",
+    description: "A method or associated fn is no longer unsafe to call. Callers that use an `unsafe` block may get `unused_unsafe` warnings.",
+    required_update: Minor,
+    lint_level: Allow,
+    reference_link: Some("https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html#calling-an-unsafe-function-or-method"),
```
obi1kenobi, line 7: This doc appears to have moved to a new address recently. Looks like a chapter was added to the book.
```suggestion
    reference_link: Some("https://doc.rust-lang.org/book/ch20-01-unsafe-rust.html#calling-an-unsafe-function-or-method"),
```

obi1kenobi:
Thanks for the PR!

PR #1581, opened 2026-02-24
Add minor lint: repr(C) added on public API type

h-wal:
Implements the repr(C)-added lint from #949.

### DIFF
--- a/.github/workflows/add_new_rust_to_matrix.yml
+++ b/.github/workflows/add_new_rust_to_matrix.yml
@@ -117,7 +117,7 @@ jobs:
           sed -i "s/$CURRENT_UP/$NEXT_UP/g" .github/workflows/ci.yml
 
       - name: upload ci.yml artifact for use in PR
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: ci.yml
           path: .github/workflows/ci.yml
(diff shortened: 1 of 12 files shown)


test_crates/repr_c_added/new/src/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+pub enum Bar {
+    A,
+    B,
+}
+
+#[repr(C)]
+pub union Baz {
+    pub x: u32,
+    pub y: i32,
+}
```
obi1kenobi: Leaving a trailing newline at the end of files is nice for interoperability with various Unix tools.

If you use a VSCode-like editor, please consider adding the following lines to your settings:
```
```

obi1kenobi:
```suggestion
}

```

test_crates/repr_c_added/old/src/lib.rs
```diff
@@ -0,0 +1,16 @@
...
+
+pub enum Bar {
+    A,
+    B,
+}
+
+pub union Baz {
+    pub x: u32,
+    pub y: i32,
+}
```
obi1kenobi:
```suggestion
}

```

test_crates/repr_c_added/new/src/lib.rs
h-wal: I've added these to my editor settings. Thanks for the suggestion.

PR #1589, opened 2026-03-05
Add minor lint:  inherent_method_now_const

h-wal:
added the lint to detect public inherent methods that change from non-const to const.

### DIFF
--- a/test_crates/inherent_method_now_const/new/src/lib.rs
+++ b/test_crates/inherent_method_now_const/new/src/lib.rs
@@ -1,6 +1,6 @@
 #![no_std]
 
-//TRUE POSITIVES: these methods are now const, but were not const in the old version, so should be reported.
+// TRUE POSITIVES: these methods are now const, but were not const in the old version, so should be reported.
 
 pub struct PubStruct;
 
@@ -43,7 +43,7 @@ impl PubUnion {
     }
 }
 
-//FALSE POSITIVES: these methods are const in both versions, so should NOT be reported.
+// FALSE POSITIVES: these methods are const in both versions, so should NOT be reported.
 
 pub struct AlreadyConst;
 
--- a/test_crates/inherent_method_now_const/old/src/lib.rs
+++ b/test_crates/inherent_method_now_const/old/src/lib.rs
@@ -1,6 +1,6 @@
 #![no_std]
 
-//TRUE POSITIVES: these methods become const in the new version
+// TRUE POSITIVES: these methods become const in the new version
 
 pub struct PubStruct;
 


src/lints/inherent_method_now_const.ron
```diff
@@ -0,0 +1,73 @@
...
+    id: "inherent_method_now_const",
+    human_readable_name: "pub method in inherent impl has newly become const",
+    description: "A public method in a public type's inherent impl has newly become `const`, allowing it to be used in a `const` context. Removing `const` later would be a breaking change.",
+    required_update: Minor,
+    lint_level: Allow,
+    reference_link: Some("https://doc.rust-lang.org/reference/const_eval.html#const-functions"),
+    query: r#"
+    {
+        CrateDiff {
+            baseline{
```
obi1kenobi: Please make sure to format the queries in all lints for readability -- e.g. make sure there's a space before `{` etc.

h-wal: Done, thanks for the feedback, I have formatted the lint query for readability (spaces before {, etc.) and added trailing newlines to all test crate files.

test_crates/inherent_method_now_const/new/src/lib.rs
```diff
@@ -0,0 +1,115 @@
+#![no_std]
+
+//TRUE POSITIVES: these methods are now const, but were not const in the old version, so should be reported.
```
obi1kenobi:
```suggestion
// TRUE POSITIVES: these methods are now const, but were not const in the old version, so should be reported.
```

test_crates/inherent_method_now_const/new/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+impl PubUnion {
+    /// Union associated fn became const, but was not const in the old version, so should be reported.
+    pub const fn new(val: u32) -> Self {
+        PubUnion { f1: val }
+    }
+}
+
+//FALSE POSITIVES: these methods are const in both versions, so should NOT be reported.
```
obi1kenobi:
```suggestion
// FALSE POSITIVES: these methods are const in both versions, so should NOT be reported.
```

test_crates/inherent_method_now_const/old/src/lib.rs
```diff
@@ -0,0 +1,112 @@
+#![no_std]
+
+//TRUE POSITIVES: these methods become const in the new version
```
obi1kenobi:
```suggestion
// TRUE POSITIVES: these methods become const in the new version
```

obi1kenobi (commented):
Please make sure to self-review the PR in the GitHub UI after you push it. The change of environment in terms of where you're looking at the code has a huge effect on making it possible to see issues you would have missed otherwise. For example, I bet you would have caught the typos below if you looked at GitHub, even though I imagine you didn't notice them in your editor.

h-wal:
Got it, thanks! I’ll make sure to review the PR in GitHub before submitting next time.

PR #1592, opened 2026-03-07
Add minor lint: trait_associated_const_with_default_added

h-wal:
Added a new lint trait_associated_const_with_default_added that reports when a non-sealed public trait gains a new associated constant with a default value.


obi1kenobi (commented):
I just realized something — we have a gap in the lint coverage here.

Adding an associated const can be breaking in more ways than we track currently.
- it's breaking if the trait isn't sealed and the const doesn't have a default, because downstream impls are broken
- it's also breaking if the trait is [dyn-compatible](https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility), regardless of whether the trait is sealed or whether the const has a default

[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=606033e62ec9383d3b6bd39c3f7983d5)
```rust
pub trait Foo {
    // Uncommenting this line breaks the function below.
    // const N: i64 = 0;
}

fn bar(_: &dyn Foo) {}
```

We have two issues now:
- We currently have a lint for the former but not the latter. 
- This minor-lint actually partially overlaps with the latter case, because it doesn't check for dyn-compatibility.

There are two tasks to do, which you're welcome to take on if you'd like:
- first, fixing up this lint to only fire for traits that were already previously non-dyn-compatible
- then, add one or more major deny-by-default lints + adjust the existing one so that all the breaking cases are covered without overlaps -- no case triggers more than one lint

h-wal:
Thanks for catching that, I’ll fix it.
I’ll (1) restrict the minor lint to traits that were already non–dyn-compatible, and (2) add major deny-by-default lint(s) for the dyn case and adjust things so every breaking case is covered with no overlap.

obi1kenobi:
Awesome, thanks!

obi1kenobi:
Just checking, are you still working on this PR? No worries if you got busy and aren't able to finish it -- it happens :)

I'll keep it open for one week to give you a chance to reply, after that I'll close it if I haven't heard back.

PR #1593, opened 2026-03-07
Add minor lint: trait_associated_type_added_to_sealed_trait

h-wal:
Adds a new lint trait_associated_type_added_to_sealed_trait that reports (lint level - allow)  when a sealed public trait gains a new associated type (with or without a default).


src/lints/trait_associated_type_added_to_sealed_trait.ron
```diff
@@ -0,0 +1,60 @@
+SemverQuery(
+    id: "trait_associated_type_added_to_sealed_trait",
```
obi1kenobi: Same rename request as in the other PR, if you wouldn't mind. Let's put `sealed_trait` at the front instead of at the end.

h-wal:
Done, The lint is also now renamed to sealed_trait_associated_type_added

src/lints/sealed_trait_associated_type_added.ron
```diff
@@ -0,0 +1,60 @@
+SemverQuery(
+    id: "sealed_trait_associated_type_added",
+    human_readable_name: "pub sealed trait gained new associated type",
+    description: "A sealed public trait added a new associated type. This is not a breaking change since downstream crates cannot implement the trait.",
```
obi1kenobi, line 4: slight rephrase to avoid confusion around whether this addition sealed the trait or not
```suggestion
    description: "A sealed public trait added a new associated type. This is not a breaking change since downstream crates couldn't have implemented the sealed trait.",
```

test_crates/sealed_trait_associated_type_added/new/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+// Non-sealed trait gained a type with default.
+// This is handled by `trait_associated_type_with_default_added`.
+pub trait NonSealedWillGainTypeWithDefault {
+    type Bar = bool;
+
+    fn make_me_non_dyn_compatible() -> Self;
+}
+
+// Was sealed in old, lost its seal, and gained a type.
+// Our lint does NOT fire because the trait is no longer sealed in current.
```
obi1kenobi, line 71: I think the lint should fire here. The trait no longer being sealed is unrelated -- it's important that external crates couldn't have `impl`'d the trait in the past to make sure this isn't a major breaking change. Whether they can do so now is not important here.

obi1kenobi:
Just checking, are you still working on this PR? No worries if you got busy and aren't able to finish it -- it happens :)

I'll keep it open for a few more days to give you a chance to reply, after that I'll close it if I haven't heard back.

PR #1595, opened 2026-03-07
Fix partial_ord_enum_variants_reordered false positive

### LINKED ISSUE #1244: partial_ord_enum_variants_reordered Does not seem to allow inserting new variants in a non_exhaustive enum
### Which lint or lints are the issue

Let me begin by saying that this `partial_ord_enum_variants_reordered` lint is awesome 👏 

I manage a project that has an enum where we add variants periodically. To make propagating this in our ecosystem easier, we use `#[non_exhaustive]` which forces our code to ignore "unknown-to-them" variants.

When inserting new variants in the enum, the `partial_ord_enum_variants_reordered` seems to FP and says that the ordering of the variant has changed. I don't think that is true in a `#[non_exhaustive]` context since all the existing/unchanged variants of the enum still order amongst themselves the same which means any existing code using these variants will not break.

### Known issues that might be causing this

- [ ] Is the flagged item defined in another crate, or a re-export of such an item? ([#638](https://github.com/obi1kenobi/cargo-semver-checks/issues/638))

Inserting a new variant in an enum like :
```rust
#[derive(PartialEq, PartialOrd)]
#[non_exhaustive]
pub MyEnum {
    A,
    B,
    Z,
}
```

### Actual Behaviour

Adding a variant to the enum will trigger the lint
```rust
#[derive(PartialEq, PartialOrd)]
#[non_exhaustive]
pub MyEnum {
    A,
    B,
    // This is a new variant. All other variants still order the same relative to each other
    T,
    Z,
}
```

### Expected Behaviour

The rule should not trigger on `#[non_exhaustive]` enums when only new variants are added

### Verbose Lint Output

N/A

N/A

### Build Configuration

_No response_

### Additional Context

_No response_

MarcosAsh:
## Summary

  Fixes #1244.

  The partial_ord_enum_variants_reordered lint produced false positives when new variants were inserted into a #[non_exhaustive] enum without changing the relative order of existing variants. For example, going from {A, B, Z} to {A, B, T, Z} would incorrectly warn that Z moved from position 3 to 4.

  ## Implementation 

  - Restricts the existing lint to exhaustive enums (where absolute position changes always indicate a real reorder).
  - Adds a companion lint partial_ord_non_exhaustive_enum_variants_reordered that uses a pairwise inversion check: it only fires
  when two variants that both existed in the baseline have swapped their relative order.

 ## Test plan

  - Added #[non_exhaustive] test cases to the existing test crate (insertion-only and reorder scenarios, confirming the original
  lint no longer fires on them).
  - Created a new test crate for the pairwise lint covering: genuine reorders, insertion-only (no trigger), insertion + reorder,
  exhaustive enums (no trigger), private/doc-hidden/manual-impl enums (no trigger), enum becoming #[non_exhaustive] with reorder, and mixed variant types.
  - Verified end-to-end with a manual test reproducing the exact #1244 scenario.


obi1kenobi:
A few questions for you:
 
> ## Implementation
> * Restricts the existing lint to exhaustive enums (where absolute position changes always indicate a real reorder).

Is this true? What happens if I make a major breaking change by adding a new variant to the exhaustive enum, in exactly the same way: `{A, B, Z}` to `{A, B, T, Z}`?

> * Adds a companion lint partial_ord_non_exhaustive_enum_variants_reordered that uses a pairwise inversion check: it only fires
>   when two variants that both existed in the baseline have swapped their relative order.

Consider this case: `{A, V, W, X, Y, Z}` to `{V, W, X, Y, Z, A}`. How many times does the new lint fire here? How many times do we want it to fire?

Separately, I'm concerned that this PR:
- seems to be a straight-up copy-paste from an LLM tool
- claims to fix an issue
- ignores [substantive discussion present in the issue](https://github.com/obi1kenobi/cargo-semver-checks/issues/1244#issuecomment-2777423948)
- doesn't actually fix the problem it claims to fix, nor addresses the points raised in the issue's discussion

As you may have read, [we have a zero-tolerance policy for low-quality AI generated contributions](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md#making-your-first-contribution):
> [!IMPORTANT]
> **We have a zero-tolerance policy for low-quality AI-generated contributions.**
>
> It's very easy to use AI to generate low-quality "contributions" that are not carefully considered and do not meet the quality bar for inclusion in the codebase.
> Such "contributions" do not benefit the project's stakeholders — in fact, they are a drain on maintainers' and reviewers' time and energy.
> Instead, the "contributions" serve to boost the contributor's GitHub statistics or achieve other personal goals, at the expense of the project and its stakeholders.
>
> Contributions that appear to be low-quality AI slop will be rejected without further explanation. Further contributions by the same authors will also be rejected without review or explanation, as the good faith assumption has already been breached once.
> We express no opinion on AI assistance *in general*, and do not ban AI tools outright — we merely require that you take personal responsibility for the contribution's quality instead of relying on "the AI thought it was a good idea."

This PR has me on the fence. What do you think?

MarcosAsh:
For question 1, limiting the original lint to exhaustive enums doesn't help because the same false positive pattern happens if a variant is added without changing the order. I didn't think that through.

For question 2, the pairwise check fires 5 times for that case (once for each pair with A), which is obviously not what we want.

I also missed your comment in #1244 about how to use @transform on properties to figure out adjusted positions :
`(current_position - count(added at lower positions) + count(removed at lower positions) != previous_position)`.
That method avoids both of these issues. It only fires once for each variant that has been reordered, and it works no matter how thorough it is. Before suggesting a different option, I should have taken part in that discussion.

I did use AI tools to help me validate my approach, but I relied too much on it. I did just miss that part in the discussion that was clearly looking back now important. I will make sure that this isn't the case in the future.

I'd like to end this and come back with an attempt that builds on the @transform method. Is the Trustfall feature out now, or is it still not out? I'm happy to build out the test corpus you talked about in the issue if it's not ready yet. That would be helpful no matter how the final version turns out.

obi1kenobi:
Thank you for your candid reply.

The Trustfall feature isn't out yet, unfortunately. The test corpus would be very helpful!

PR #1639, opened 2026-05-18
Add support for supressing unreleased items breakage

gstvg:
`cargo-semver-checks` compare changes across two branches, so, in per-PR workflows, we either check with main and risk flagging breakage on unreleased items, or check with the latest release, and PRs against a project whose main already has known breakage relative to its last release will keep flagging the same pre-existing findings on every PR

This PR adds two new `-Z unstable-options` flags that together enable a new workflow: save findings of main, subtract them from the PR run

- `--report-output <PATH>` — serialize the full `CrateReport`'s as JSON. When set, the binary exits 0 even with findings present (provided no required-witness errors), so CI can produce a baseline without failing.
- `--baseline-report <PATH>` — load a previously saved report and subtract any lint findings already present in it from the current run, ignoring source-location drift (`span_filename/span_begin_line/span_end_line`).

The pair of flags here lets you:

In main, run `cargo semver-checks --baseline-rev <last-release-tag> -Z unstable-options --report-output baseline.json`, and maybe save it in cache.
Then, in PR branch, run `cargo semver-checks --baseline-rev <last-release-tag> -Z unstable-options --baseline-report baseline.json`, which only fails if the PR introduces new breakage on top of what main already had.
The two flags are mutually exclusive (`conflicts_with`).

We can make this in a single invocation with one flag like `--subtract-rev <last_release_tag>`, but then we can't cache the first step

For now the PR contains a single, minimal test. I would like to know if maintainers have interest in the feature and have a preference for how this should be tested before adding new ones. Thanks!

# How the subtraction works

The subtraction/removal happens after running the semver queries and witness checks, where we have a finished `Vec<LintResult>`. If `--baseline-report <file>`  is set, we decode a `CrateReport` from the file, match lint results from the current run and from the file by lint id, and remove from the current run query results every result that is also present in the crate report from the file, by checking equality of both `BTreeMap<Arc<str>, FieldValue>`, but ignoring the keys `span_filename`, `span_begin_line` and `span_end_line`. This part, the core of the PR, is very small (~25 lines), and is included below:

<details>

```rust
    // semver queries and witness checks executed above
    let baseline_report: Option<CrateReport> = todo!();
    let mut lint_results: Vec<LintResult> = todo!();

    if let Some(baseline_report) = baseline_report {
        let mut lint_map: HashMap<String, &mut LintResult> = lint_results
            .iter_mut()
            .map(|lint| (lint.semver_query.id.clone(), lint))
            .collect();

        for old_lint in &baseline_report.lint_results {
            let Some(new_lint) = lint_map.get_mut(&old_lint.semver_query.id) else {
                continue;
            };

            // note building a query results set requires adding Hash to FieldValue in trustfall
            for old_result in &old_lint.query_results {
                new_lint
                    .query_results
                    .retain(|new_result| !btreemap_eq_ignoring_span_keys(new_result, old_result));
            }
        }
    }
    // compute bumps below

type QueryResults = BTreeMap<Arc<str>, FieldValue>;

fn btreemap_eq_ignoring_span_keys(a: &QueryResults, b: &QueryResults) -> bool {
    let not_span = |(k, _): &(&Arc<str>, _)| {
        !["span_filename", "span_begin_line", "span_end_line"].contains(&k.as_ref())
    };
    a.iter().filter(not_span).eq(b.iter().filter(not_span))
}
```

</details>

False positives are possible if more span fields are added, or the existing renamed, without updating `btreemap_eq_ignoring_span_keys`

False negatives are possible if a query returns the exact same values (ignoring spans) for two different items. Suppose a project has two functions named `foo`, one in path `foo::bar` and another in `bar::foo`. If the query returned only the name of the function, `foo`, but not the paths, the removal of one function in the baseline report would supress the removal of the other function. I haven't checked every query but I believe none would trigger this.

### DIFF
--- a/src/check_release.rs
+++ b/src/check_release.rs
@@ -1,5 +1,5 @@
 use std::cmp::Ordering;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 use std::io::Write as _;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
@@ -119,7 +119,7 @@ fn get_minimum_version_change(version: &semver::Version) -> VersionChange {
 }
 
 /// Intermediate state in [`run_check_release`]
-#[derive(Debug)]
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
 pub(crate) struct LintResult {
     pub semver_query: SemverQuery,
     pub query_results: Vec<BTreeMap<Arc<str>, FieldValue>>,
@@ -260,6 +260,7 @@ fn print_triggered_lint(
     Ok(())
 }
 
+#[allow(clippy::too_many_arguments)]
 pub(super) fn run_check_release(
     config: &mut GlobalConfig,
     data_storage: &DataStorage,
@@ -268,6 +269,7 @@ pub(super) fn run_check_release(
     overrides: &OverrideStack,
     witness_generation: &WitnessGeneration,
     witness_data: witness_gen::WitnessGenerationData,
+    baseline_report: Option<&CrateReport>,
 ) -> anyhow::Result<PendingCrateReport> {
     let current_version = data_storage.current_crate().crate_version();
     let baseline_version = data_storage.baseline_crate().crate_version();
@@ -387,6 +389,26 @@ pub(super) fn run_check_release(
         &mut lint_results,
     );
 
+    if let Some(baseline_report) = baseline_report {
+        let mut lint_map = lint_results
+            .iter_mut()
+            .map(|lint| (lint.semver_query.id.clone(), lint))
+            .collect::<HashMap<_, _>>();
+
+        for old_lint in &baseline_report.lint_results {
+            let Some(new_lint) = lint_map.get_mut(&old_lint.semver_query.id) else {
+                continue;
+            };
+
+            // note building a query results set requires adding Hash to FieldValue in trustfall
+            for old_result in &old_lint.query_results {
+                new_lint
+                    .query_results
+                    .retain(|new_result| !btreemap_eq_ignoring_span_keys(new_result, old_result));
+            }
+        }
+    }
+
     let checks_duration = checks_start_instant.elapsed();
 
     let mut required_bumps = Bumps { major: 0, minor: 0 };
@@ -424,6 +446,15 @@ pub(super) fn run_check_release(
         report,
         witness_run_report,
     })
+}
+
+type QueryResults = BTreeMap<Arc<str>, FieldValue>;
+
+fn btreemap_eq_ignoring_span_keys(a: &QueryResults, b: &QueryResults) -> bool {
+    let not_span = |(k, _): &(&Arc<str>, _)| {
+        !["span_filename", "span_begin_line", "span_end_line"].contains(&k.as_ref())
+    };
+    a.iter().filter(not_span).eq(b.iter().filter(not_span))
 }
 
 fn print_report(
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,7 +17,7 @@ use clap::ValueEnum;
 use data_generation::{DataStorage, IntoTerminalResult as _, TerminalError};
 use directories::ProjectDirs;
 use itertools::Itertools;
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 
 use std::collections::{BTreeMap, HashSet};
 use std::io::Write as _;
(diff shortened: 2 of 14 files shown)


obi1kenobi:
Hi! Thanks for the suggestion, apologies I'm only just now looking at it — I was at RustWeek and I'm only just catching up.

Could you say more about how the subtraction is meant to work mechanically? What is being compared, equated, and then subtracted?

I'm open to this idea but I have reservations about whether it's possible to make the output sound. I fear that the subtraction will itself introduce tricky false positives or false negatives. Is that the case? Or is there neat trick in play that ensures subtracting two `--baseline-rev v1.2.3` runs from each other ends up being identical to a proper `--baseline-rev origin/main` run?

I think the best starting point for this feature is not code, but a design doc tackling those questions. I'd be more than happy to read that doc if you're open to writing it — just please make sure to [review our contributions policy](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md#making-your-first-contribution) before starting.

gstvg:
@obi1kenobi No worries, happy to know you went to RustWeek! How the subtraction works is very simple, I hope this comment and updating the PR description will be enough. 

The subtraction (perharps "removal of existing lints" would be a better description) happens after running the semver queries and witness checks, where we have a finished `Vec<LintResult>`. If `--baseline-report <file>`  is set, we decode a `CrateReport` from the file, match lint results from the current run and from the file by lint id, and remove from the current run query results every result that is also present in the crate report from the file, by checking equality of both `BTreeMap<Arc<str>, FieldValue>`, but ignoring the keys `span_filename`, `span_begin_line` and `span_end_line`. This part, the core of the PR, is very small (~25 lines), and is included below:

<details>

```rust
    // semver queries and witness checks executed above
    let baseline_report: Option<CrateReport> = todo!();
    let mut lint_results: Vec<LintResult> = todo!();

    if let Some(baseline_report) = baseline_report {
        let mut lint_map: HashMap<String, &mut LintResult> = lint_results
            .iter_mut()
            .map(|lint| (lint.semver_query.id.clone(), lint))
            .collect();

        for old_lint in &baseline_report.lint_results {
            let Some(new_lint) = lint_map.get_mut(&old_lint.semver_query.id) else {
                continue;
            };

            // note building a query results set requires adding Hash to FieldValue in trustfall
            for old_result in &old_lint.query_results {
                new_lint
                    .query_results
                    .retain(|new_result| !btreemap_eq_ignoring_span_keys(new_result, old_result));
            }
        }
    }
    // compute bumps below

type QueryResults = BTreeMap<Arc<str>, FieldValue>;

fn btreemap_eq_ignoring_span_keys(a: &QueryResults, b: &QueryResults) -> bool {
    let not_span = |(k, _): &(&Arc<str>, _)| {
        !["span_filename", "span_begin_line", "span_end_line"].contains(&k.as_ref())
    };
    a.iter().filter(not_span).eq(b.iter().filter(not_span))
}
```

</details>

False positives are possible if more span fields are added, or the existing renamed, without updating `btreemap_eq_ignoring_span_keys`

False negatives are possible if a query returns the exact same values (ignoring spans) for two different items. Suppose a project has two functions named `foo`, one in path `foo::bar` and another in `bar::foo`. If the query returned only the name of the function, `foo`, but not the paths, the removal of one function in the baseline report would supress the removal of the other function. I haven't checked every query but I believe none would trigger this.

> Or is there neat trick in play that ensures subtracting two --baseline-rev v1.2.3 runs from each other ends up being identical to a proper --baseline-rev origin/main run?

I may have misuderstood your question, so please correct me if I'm wrong, but they aren't going to be identical and that's the goal. Running with `--baseline-rev origin/main` flag changes to unreleased items, for example, in https://github.com/apache/datafusion/pull/22106#issuecomment-4423494347, the breakage flagged in `HigherOrderSignature` is irrelevant to versioning because it's a new item introduced in https://github.com/apache/datafusion/pull/21679, after the then latest release branch (53.0.0) was cut (so the item has not been published). If we used `--baseline-rev v53.0.0` to supress this, after a single breaking change is introduced in main, every subsequent PR would be flagged as breaking until the next release. The subtraction allow to use the `--baseline-rev v53.0.0`, while supressing changes that are already present in main. So the run with subtraction should returns *less* items compared to `--baseline-rev origin/main`

I'm okay and expected that this would require more discussions. Regarding the contributing policy, I read it before opening the PR, and assume the only part applicable to this feature is regarding AI? If that's the case, the only part which is mostly AI written is the single test, the rest is mostly hand written, since I wanted to understand the codebase myself, and the required changes looked small and the codebase was easy to follow

Finally, even if you agree the current approach, please don't review it yet. I realized that we can use a single run with `--subtract-rev <lastest_release_tag>` instead of two runs with an intermediate file with even higher cache usage, but it only makes sense to implement it if the PR as whole make sense as well. Thank you!

PR #1650, opened 2026-06-13
Update gix from 0.78 to 0.84, and enable SHA256 support

musicinmybrain:
This updates the `gix` dependency to the latest version. No source-code changes are required, but we need to enable at least one hash feature (traditionally: `sha1`) since we are disabling default features. In a follow-up commit, I also enabled the `sha256` feature; given https://git-scm.com/docs/hash-function-transition, it seems like a good idea to support that.

(I’m considering packaging this in Fedora, and this PR is part of that investigation.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -230,6 +230,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "block-buffer"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


obi1kenobi (approved):
Excellent, thank you for the PR!

Let me know if I can help with the Fedora packaging investigation. Happy to answer any questions etc.

If knowing release plans is useful for that investigation, I'm planning to make the next cargo-semver-checks release shortly after Rust 1.97 becomes stable (~July 9th), most likely over the first subsequent weekend.

PR #1652, opened 2026-06-18
Allow invocation from full binary name instead of cargo subcommand

alerque:
This project has continually annoyed me by making it hard to invoke the right tooling from specialized environments with hard coded tool paths. Using the indented Cargo wrapper to invoke the subcommand as `cargo semver-checks` works fine, but if you hard code a path to a binary like `/usr/bin/cargo-semver-checks` it refuses to run because it doesn't have the same argument structure and the top level `cargo` subcommand isn't found. This makes it difficult to do conditional execution such as finding the binary and then executing the binary you found or hard coding a full path to it somewhere out of the shell `$PATH`.

This fixes that by allowing invocation with either the full binary name *or* the cargo wrapper. This is similar to how many other Cargo tools work. For example `cargo deny` works this way (except they check for and *remove* the extra subcommand instead of checking for it's absence and adding it, but this was less disruptive to the rest of the CLI argument handling.

https://github.com/EmbarkStudios/cargo-deny/blob/main/src/cargo-deny/main.rs#L265-L269

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -36,7 +36,15 @@ fn main() {
     }
     .write_global();
 
-    let Cargo::SemverChecks(args) = Cargo::parse();
+    // Catch possible invocations from development tooling that use `cargo run` or with the full
+    // binary name `cargo-semver-checks`, and normalize the expected subcommand so that it works the
+    // same way as the cargo wrapper `cargo semver-checks` style invocation.
+    let mut raw_args: Vec<_> = env::args_os().collect();
+    if raw_args.get(1).is_none_or(|arg| arg != "semver-checks") {
+        raw_args.insert(1, "semver-checks".into());
+    }
+
+    let Cargo::SemverChecks(args) = Cargo::parse_from(raw_args);
 
     let feature_flags = HashSet::from_iter(args.unstable_features.clone());
 


src/main.rs
```diff
@@ -30,7 +30,18 @@ fn main() {
     }
     .write_global();
 
-    let Cargo::SemverChecks(args) = Cargo::parse();
+    // Catch possible invocations from development tooling that use `cargo run` or with the full
```
obi1kenobi, line 39: In a development environment, I use `cargo run -- semver-checks <...>` and that seems to work fine. Based on this comment, I gather that doesn't work in your use case?

src/main.rs
```diff
@@ -30,7 +30,18 @@ fn main() {
...
+    // binary name `cargo-semver-checks`, and normalize the expected subcommand so that it works the
+    // same way as the cargo wrapper `cargo semver-checks` style invocation.
+    let mut raw_args: Vec<_> = env::args().collect();
+    if raw_args.first().is_some_and(|bin_name| {
+        (bin_name.ends_with("cargo-semver-checks") || bin_name.ends_with("cargo-semver-checks.exe"))
```
obi1kenobi: Out of curiosity, do we need this to use pure `.ends_with()` or can we afford to be a bit more strict?

obi1kenobi (commented):
> This makes it difficult to do conditional execution such as finding the binary and then executing the binary you found or hard coding a full path to it somewhere out of the shell `$PATH`.

I have to admit this is the first time I've heard of this situation, and now I'm really curious. Could you share more about your use case?

src/main.rs
alerque: It can/will be a full path, so the ending segment is the only thing we can reasonably use. For example adding `dbg!(&raw_args);` and using `cargo run` shows:

```
[src/main.rs:37:5] &raw_args = [
    "/home/caleb/.local/share/cargo/debug/cargo-semver-checks",
    "--help",
]
```

The more strict alternative would be a slightly more expensive cast to a `Path`, letting that do some segmenting, and then matching on the filename part. Given we're mangling the args *prior* to actual parsing I think this will catch most cases. If people change the case of the binary from any normal situation they won't benefit.

src/main.rs
alerque: That works *if* you know to run it. Also it's inconsistent with regular usage and many other tools. The `cargo run` case is primarily developer facing though. The more important use case is is a shell script (Makefile, Justfile, etc.) that detects whether a tool is available, allows the user to make it available via a custom path, then tries to use it from that path. E.g. a Justfile like this:

```justfile
cargo-semver-checks := require('cargo-semver-checks')

check:
  {{ cargo-semver-checks }}
```

That will fail because the command you discover on the path or specified externally with an argument at run time is not the one you need to actually run.

It turns out you can add the extra subcommand here too, but that become repetitively redundant (pun intended). Even I didn't realize that was possible until working on this PR. It's especially frustrating that the `--help` message isn't even for the right program!

```console
$ cargo-semver-checks --help
Usage: cargo

Options:
  -h, --help     Print help
  -V, --version  Print version
```

There is no discoverability here that even helps people find a work around.

I would suggest that:

```bash
# These two being equivalent is messy and unexpected:
cargo semver-checks
cargo-semver-checks semver-checks

# These two should be 100% equivalent invocations:
cargo semver-checks
cargo-semver-checks
```

obi1kenobi:
I've been thinking about this problem on and off in preparation for the next `cargo-semver-checks` release, and I wanted to get your thoughts on something.

I'm not sure `.ends_with()` is buying us much, and yet it feels unnecessarily brittle. For example, it would not support directly invoking a versioned binary such as `cargo-semver-checks-v0.48` when A/B testing several versions side by side.

Also, `env::args()` panics if any process argument is not valid Unicode, whereas `clap` normally parses `OsString` arguments via `args_os()`. That could affect non-Unicode executable or manifest paths on Unix.

Since Cargo documents that it passes `semver-checks` as `argv[1]`, what do you think about normalizing solely based on whether that argument is present? I'm not sure we need to look at `argv[0]` at all, given that we know it's "us" in some form or another.

It would look something like this:
```rust
let mut raw_args: Vec<_> = env::args_os().collect();

if raw_args
    .get(1)
    .is_none_or(|arg| arg != "semver-checks")
{
    raw_args.insert(1, "semver-checks".into());
}

let Cargo::SemverChecks(args) = Cargo::parse_from(raw_args);
```

If this seems reasonable and you're open to updating the PR to this approach, I'd be happy to merge it.

alerque:
That sounds like a plausible simplification. I'll look into it.

alerque:
Yes as far as I'm concerned that make perfect sense. Checking for the case where cargo will add the extra subcommand makes a lot more sense than trying to check for all the cases where it won't. I guess I came at this with some brain fog from having reviewed other implementations. This simplification can probably be submitted to a few other projects ;-)

obi1kenobi (approved):

obi1kenobi:
Awesome, thanks for flagging this and for the help with fixing it!

Issue #1659, opened 2026-06-30
Breakage in items of reachable but not nameable type


obi1kenobi:
Sibling of #1658, except instead of a trait accessible via a supertrait, we have a type that is accessible by virtue of being returned from a public API function like so:
```rust
mod sealed {
    pub struct S;
}

pub fn example() -> sealed::S {
    sealed::S
}
```

We'll need a huge suite of lints here, covering cases where the type is a struct / enum / union as well as covering its associated items.

I was not able to come up with witnesses for the following cases, so it's possible they aren't breaking:
- breakage of inherent associated consts
- breakage of inherent associated functions without a receiver
- breakage of enum variants

Most of the analogous open questions from #1658 are open here as well. It's likely that we neither need to, nor should, attempt to equate the pub-in-priv types, and instead should equate them as "the value returned from such-and-such public API function" because _in principle_ the type of that function _might be possible_ to be changed without causing breakage.

AdemVessell:
For the enum-variant question, a numeric cast seems to give a small downstream
breaking example without naming the returned type.

Library:

```rust
mod sealed {
    pub enum E { A, B }
}

pub fn example() -> sealed::E { sealed::E::A }
```

Downstream, compiled as a separate crate:

```rust
fn main() {
    let value = upstream::example();
    let _ = value as isize;
}
```

The caller compiles before, but fails with E0605 if `B` becomes `B(u8)`, or if
`C(u8)` is added. Both libraries still compile. I checked this on Rust 1.95.0
and nightly-2026-04-01, edition 2021. Trying to name `upstream::sealed::E`
fails with E0603.

As controls, adding or renaming a unit variant keeps this caller compiling;
changing `B` to `B()` also works because the enum remains fieldless. So this
is specifically loss of numeric castability, not evidence that every change
to an inaccessible variant breaks downstream code.

The [cast rule](https://doc.rust-lang.org/reference/items/enumerations.html#casting)
is already known, and the existing discriminant lint covers the related
importable-enum case. I have not run the full checker on this example. I also
do not have a witness for the inherent-constant or receiverless-function cases.

AdemVessell:
An update to my earlier comment: I have now run the checker on this example and put together a narrow fix locally. Would an adapter change followed by a lint for this cast case be a useful first contribution toward this issue?

[...]

Issue #1658, opened 2026-06-30
Breakage in items of reachable but not nameable supertrait


obi1kenobi:
Example minified from https://github.com/time-rs/time/issues/793
```rust
mod sealed {
    pub trait Sealed {
        fn example(&self) { ... }
    }
}

pub trait Visible: sealed::Sealed {}
```

Now modify:
```diff
  mod sealed {
      pub trait Sealed {
-          fn example(&self) { ... }
+          fn example(&self, value: &[u8]) { ... }
      }
  }
```

Callers are broken as follows:
```rust
fn witness(value: impl upstream::Visible) {
    value.example()
}
```
because `example` is in scope as a supertrait method of an in-scope trait, where the supertrait is `pub`.

It's unclear whether `example` is public API here or not. There are arguments on both sides. The "not public API" argument is that the method is not explicitly documented, because it's a member of a pub-in-priv item and for example, you couldn't use UFCS to call it directly. The opposing argument is that the method is not `#[doc(hidden)]` and it's available on a public trait, which by default inherits and makes available all its supertraits' items under default visibility (i.e. matching the subtrait's own visibility, `pub` in this case).

To my knowledge, there's no normative text or clearly applicable precedent that explicitly tackles this case. **This is why SemVer is hard. Do not flame the maintainers who ran into this and are now helping fix it going forward!**

For my taste, the opposing argument wins narrowly, making `Sealed::example` public API. I find it analogous to returning a pub-in-priv type from a pub function, where pub items associated with that type can be public API even though the type itself cannot be named directly. I also find the argument of "the downstream user wrote code that does not touch any `#[doc(hidden)]`" reasonably compelling. **EDIT:** The "this is public API" side is getting some traction, so I'm calling this public API.

To avoid duplicate lints, `cargo-semver-checks` lints here should look for associated items on _non-importable but public API eligible supertraits of public API traits_. A reminder that "public API eligible" means either deprecated or not `#[doc(hidden)]` [due to existing precedent](https://predr.ag/blog/checking-semver-for-doc-hidden-items/). The supertrait whose items are broken [must be `pub`](https://github.com/rust-lang/rust/issues/158654#issuecomment-4863199967).

We'll need lints for the following kinds of breakage, at minimum:
- [ ] associated type removed
- [ ] associated const removed
- [ ] method removed
- [ ] method argument count changed
- [ ] method receiver changed (see existing lints, we'll need more than one here)
- [ ] method no longer returns a value
[...]

PR #1676, opened 2026-07-17
Add lint: pub type no longer implements pub trait

ivelieu:
Hello! This is my first PR here, I read the contributing doc and picked one of the lints on the list at https://github.com/obi1kenobi/cargo-semver-checks/issues/5 "type no longer implements pub trait". There were a few options I considered:
- Whether both the trait and the type should be public, which I'm assuming yes.
- How generics and lifetimes should be added. There are three kinds of trait impls as listed in the schema:

> - inherent impls: `impl Foo`
> - explicit trait implementations: `impl Bar for Foo`
> - blanket implementations: `impl<T> Bar for T`

The first one isn't relevant, but for a blanket implementation, it's not clear to me if the lint should flag

1. Only for explicit trait implementations (smallest coverage)
2. For explicit impls, or if any blanket impl is removed for some `T` (which could be trait bounded) (largest coverage)
3. For explicit impls, or if all blanket impls are removed

The reason why option 3 could make sense, is I can think of situations where `T` is changed from a type to a more permissive trait bound, which option 2 would flag as a false positive. But that kind of reasoning would only be possible with type checking. The readme says that false positive reported semver violations are a bug, leading me to think option 3 is best even though it does miss some coverage.

Also, if `T` is a trait bound and not a type, and it's possible for some of the trait bounds to be satisfied by a public type, then removing the impl could be breaking even if the trait bounds themselves are private. That seems a bit intractable to me, and implicates that option 3 could also produce false positives.

So this is a pretty tricky edge case, I'd appreciate advice on which option would be best, with the understanding the current implementation is option 3 since that won't produce false positives and has better coverage than option 1, but it can be changed to 1 or 2.

- Whether the trait needs to be defined in the same crate. The current schema says:

> Currently, item information for items from a foreign crate is not generally available,
  so if the trait is defined in a different crate then this edge won't have any instances.

So I'd think that means matching on `ImplementedTrait.trait_id` and providing the additional trait information as an optional if it's available. I tested this and found there are a few existing lints this will overlap with if this is set: `derive_trait_impl_removed` which is removing derive traits, and the `partial_ord_*_reordered` lints which are incidentally removing Ord/PartialOrd to do some of its checks. These traits are public, and the specific removal of the trait is not checked by those lints, so it is a different use case that makes sense to have checked against.

So I've excluded derive traits from the check. If the trait data is available, my current implementation matches against it to ensure it is a public trait.

`auto_trait_impl_removed` also had some overlap with this lint, so I've excluded anything tagged with `#[automatically_derived]`, and as a second layer also [synthetic trait impls](https://docs.rs/rustdoc-types/latest/rustdoc_types/struct.Impl.html#structfield.is_synthetic).

`sized_impl_removed` also had overlap, which I believe should be included because the offending trait, [`SizedTypeProperties`](https://doc.rust-lang.org/src/core/mem/mod.rs.html#1375), is a blanket impl, not explicitly derived, not `#[automatically_derived]`, not synthetic, not auto and not checked against in the current output of any other lints. Since it's a subtrait of `sized`, I've excluded it explicitly.

`sized_impl_removed` is also flagged for `From` and `Into`, and the `instantiated_name` doesn't seem to display trait bounds as the output I'm seeing is `String("Into<U>")`. I'd appreciate advice if there is a way to include the trait bounds in the output. Personally, I would expect this specific `From` and `Into` to be marked synthetic, but it's not.

According to [the rust unstable book](https://doc.rust-lang.org/beta/unstable-book/language-features/negative-impls.html#semver-interaction), removing a negative trait impl is semver breaking, so I don't have a check to exclude them.


obi1kenobi:
Hi! Out of curiosity, how much of this PR's code and description did you write by hand, vs generate via an LLM? Also, which model did you use?

ivelieu:
All by hand. (I did set up the grpahql LSP, but I don't have a formatter configured yet.) No AI and I will never use AI on this or any other project.

ivelieu:
Just saw from the CI I forgot to run sort_add_lists.sh after renaming the lint while I was working on it, I've ran it now.

obi1kenobi:
Could I try to convince you to shelve this lint for now and start with something smaller in scope (ideally one labeled `E-mentor`), then work back up to this if you're particularly curious?

This is an extremely difficult lint to write, and not a good one to get started with. There are lots of tricky design decisions to be made; you've already flagged some of them, but there are more. For example, what do we do with generic traits like `MyTrait<'a, T>` and generic types `MyType<'a, T>`? What happens if we have `impl<'a> MyTrait<'a, T> for MyType<'a, T>` and it goes to `impl<'a> MyTrait<'a, T> for MyType<'static, T>`?

As another example, I'm pretty sure that your lint as written would return false-positives if a crate switches from a manual impl of a trait to a `#[derive]` of the same trait. We'd also have all the usual problems of cross-crate info and type-checking lints etc. that many lints suffer from.

Overall I salute the ambition to ship this difficult lint. But I don't think `cargo-semver-checks` can deliver on a good user experience with this lint today, and I'd like to help you get it there while also polishing up your lint-writing skills by writing some slightly less challenging lints first. The ones described in #949 are an easy first start, and #1658 and #1659 are great if you want a challenge that's still feasible today.

ivelieu:
> Could I try to convince you to shelve this lint for now and start with something smaller in scope

Sure thing. I noticed some type mistakes in the test crate which I've fixed, and I'll respond to the points you've raised and go no further with this one for now. By the way, I have a handful of unrelated technical questions just about the project, is there a best place on the zulip (e.g. the gsoc thread) or elsewhere to ask them?

> what do we do with generic traits like MyTrait<'a, T> and generic types MyType<'a, T>?

Ignore the lifetime and generic parameters on both the trait impl and the type, and only flag if the trait stops being implemented entirely (option 3 from earlier). That's what the current implementation does, by checking matches on `bare_name` but showing the `instantiated_name` in failed lint checks. As I mentioned, I can see option 2 would be useful because it has higher coverage, but doing it properly even restricted to only checking lifetimes/generics on the type would still in my opinion require type checking for all the edge cases (and of course is much bigger in scope). The test crate already has a check for the cases of generic types on the trait (`PubGenericTrait`) and lifetime bounds on the type (`PubLifetimeBoundStruct`), I could add tests for other combinations too.

> What happens if we have impl<'a> MyTrait<'a, T> for MyType<'a, T> and it goes to impl<'a> MyTrait<'a, T> for MyType<'static, T>?

The specific version you mention, the lint currently does not flag, I would not expect it to based on what I mentioned in the previous paragraph. One option is having it checked by a separate (hypothetical and currently nonexistent) `impl_mismatched_generic_lifetimes` lint. I don't think it's in scope of `trait_mismatched_generic_lifetimes`. The similar `impl<'a> MyTrait for MyType<'a>` to `impl MyTrait for MyType` with a held `'static` gets flagged by `type_mismatched_generic_lifetimes` and not by this lint.

> I'm pretty sure that your lint as written would return false-positives if a crate switches from a manual impl of a trait to a #[derive] of the same trait.

No it does not, as written. If it did it would flag the test crate in `derive_trait_impl_removed`, and derived traits are explicitly excluded in the lint since they are checked by `derive_trait_impl_removed` (I had a comment in the query which mentioned this too, and I noticed I left a mention of testing against negative traits that I've removed).

For the lifetime changing and derive impl <-> manual impl, I've added tests to show that it passes in a separate commit (they can be squashed later). The lifetime changing tests are correctly flagged by `type_mismatched_generic_lifetimes`.

obi1kenobi:
> > Could I try to convince you to shelve this lint for now and start with something smaller in scope
> 
> Sure thing. I noticed some type mistakes in the test crate which I've fixed, and I'll respond to the points you've raised and go no further with this one for now. By the way, I have a handful of unrelated technical questions just about the project, is there a best place on the zulip (e.g. the gsoc thread) or elsewhere to ask them?

If you have Discord, please feel free to join our server and ask there: https://discord.gg/ChhaVRUR

If not, Zulip is fine but most likely the other folks who know things about the project won't see your questions :)
 
> > I'm pretty sure that your lint as written would return false-positives if a crate switches from a manual impl of a trait to a #[derive] of the same trait.
> 
> No it does not, as written. If it did it would flag the test crate in `derive_trait_impl_removed`, and derived traits are explicitly excluded in the lint since they are checked by `derive_trait_impl_removed` (I had a comment in the query which mentioned this too, and I noticed I left a mention of testing against negative traits that I've removed).
> 
> For the lifetime changing and derive impl <-> manual impl, I've added tests to show that it passes in a separate commit (they can be squashed later). The lifetime changing tests are correctly flagged by `type_mismatched_generic_lifetimes`.

Here's a repro:
```rust
// baseline
use serde::{Serialize, Serializer};

pub struct Unit;

impl Serialize for Unit {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer,
    {
        serializer.serialize_unit_struct("Unit")
    }
}

// becomes current
use serde::Serialize;

#[derive(Serialize)]
pub struct Unit;
```

Here's the current output of this PR:
```
--- failure type_removed_trait: type no longer implements pub trait ---

Failed in:
  [type_removed_trait_serde_manual_to_derive, Unit]::Unit
  no longer implements trait ::Serialize in src/lib.rs:4
```

Aside from the templating issue in the path, this is a false positive.

There are a handful of other issues, some of which are not resolvable today — we'll need to generate [witnesses](https://predr.ag/blog/semver-violations-are-common-better-tooling-is-the-answer/#automated-validation-via-witnesses) to fully fix them, and we're still months away from that. This is why I recommended shelving this for now.

But if you might be interested in contributing to the project in the longer run, I'd be happy to work with you to find a series of PRs that culminates in shipping this lint, or even something more ambitious :)

obi1kenobi:
I'll close this PR just for bookkeeping purposes, but very happy to continue to mentor you toward your push to build and ship this lint with all the edge cases fully covered!

Issue #1660, opened 2026-06-30
Breakage around generic associated types (GATs)


obi1kenobi:
We need to work out the rules for GATs. Inherent GATs are currently unstable (https://github.com/rust-lang/rust/issues/8995) but trait-associated GATs are stable and widely used.

I believe both adding and removing generics are breaking for GATs in public API traits, in the same ways it is for public functions.

Adding and removing bounds should be breaking similarly to how it's breaking for bounds in public API methods (keeping in mind sealed traits etc.).

We need to dig into all the edge cases. There are probably a dozen or so lints here.

ivelieu:
At roughly the same level of granularity the corresponding public function lints act on, I can think of:

The number of generics changed (I guess this would include becoming generic or becoming non-generic)
- `trait_generic_associated_type_requires_different_type_params`
- `trait_generic_associated_type_requires_different_lifetime`
- `trait_generic_associated_type_requires_different_const_params`

The trait bounds of a GAT changed, but the number of bounds didn't change: `trait_generic_associated_type_bounds_changed`

In a trait with at least two lifetimes, the lifetime of a GAT changed from one to the other. I think this would require checking if the lifetimes themselves got renamed in the old/new crate. `trait_generic_associated_lifetime_changed`

The generics got reordered:
- `trait_generic_associated_type_params_reordered`
- `trait_generic_associated_type_lifetime_reordered`
- `trait_generic_associated_type_const_params_reordered`

Am I missing anything? Are these names okay?

Also, should these be submitted as separate PRs, separate commits in one PR, or as stacked PRs?

obi1kenobi:
Directionally yes I think this is correct, and good enough to get started.

A few inline comments, mostly because you've thought this space more than I have, and I'd like to build up my intuition based on what you've discovered:

> The trait bounds of a GAT changed, but the number of bounds didn't change: `trait_generic_associated_type_bounds_changed`

Is it important that the number of bounds didn't change? I didn't notice lints for "the number of bounds changed" and perhaps such lints might be ill-advised from a UX design perspective. Perhaps it's better to have a single "bounds changed" lint to cover all the changes? I'm not sure, some experimentation is probably warranted.

[...]

PR #1683, opened 2026-07-19
Add minor lint: auto trait impl added to pub type

ivelieu:
This is one of the minor lints at https://github.com/obi1kenobi/cargo-semver-checks/issues/949. It's based on `auto_trait_impl_removed` and uses the same hardcoded list of auto traits from core.


src/lints/auto_trait_impl_added.ron
```diff
@@ -0,0 +1,90 @@
...
+                    }
+                }
+            }
+        }
+    }"#,
+    arguments: {
+        "public": "public",
+        "zero": 0,
+        "false": false,
+        "true": true,
```
obi1kenobi: nit: could you use `= true` and `!= true` to avoid having to define both boolean literals? If the field is non-nullable then that should work.

Prior art for this: https://github.com/obi1kenobi/cargo-semver-checks/blob/a3331d0cf9becd894231041e702573329d9691b1/src/lints/copy_impl_added.ron#L46

This is arguably a low-priority, nice-to-have fix for the existing auto trait removal lint too.

test_crates/auto_trait_impl_added/old/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+#![no_std]
+
+use core::cell::RefCell;
+
+// struct gains Sync and RefUnwindSafe
+// when the RefCell private field is replaced
+pub struct PubStruct {
+    s: RefCell<usize>,
+}
```
obi1kenobi: Could you cover the other auto traits as well, just to be sure? You can reuse the `auto_trait_impl_removed` test cases "in reverse" by swapping "old" and "new": https://github.com/obi1kenobi/cargo-semver-checks/blob/a3331d0cf9becd894231041e702573329d9691b1/test_crates/auto_trait_impl_removed/old/src/lib.rs

src/lints/auto_trait_impl_added.ron
```diff
@@ -0,0 +1,90 @@
+SemverQuery(
+    id: "auto_trait_impl_added",
+    human_readable_name: "public type gained auto trait",
+    description: "A public type has gained an auto trait's implementation.",
+    required_update: Minor,
+    lint_level: Allow,
+    reference_link: Some("https://doc.rust-lang.org/cargo/reference/semver.html#item-new"),
```
obi1kenobi: Consider https://doc.rust-lang.org/reference/special-types-and-traits.html#auto-traits as a possible better link. I'm not sure `#item-new` is the most applicable section here.

src/lints/auto_trait_impl_added.ron
```diff
@@ -0,0 +1,90 @@
...
+        "true": true,
+        "auto_traits": [
+            ["core", "marker", "Send"],
+            ["core", "marker", "Sync"],
+            ["core", "marker", "Unpin"],
+            ["core", "panic", "unwind_safe", "RefUnwindSafe"],
+            ["core", "panic", "unwind_safe", "UnwindSafe"],
+        ],
+    },
+    error_message: "A public type has gained an auto trait's implementation. Removing this auto trait in future would be a breaking change.",
```
obi1kenobi: tiny nit: word choice
```suggestion
    error_message: "A public type has gained an auto trait implementation. Removing this auto trait in future would be a breaking change.",
```

test_crates/auto_trait_impl_added/old/src/lib.rs
obi1kenobi, line 1: If you want to go for "extra credit" test coverage, consider adding coverage that proves that neither pub `#[doc(hidden)]` nor private types trigger the lint.

obi1kenobi (approved):
Broadly looks good! A few suggestions on additional good habits to pick up when writing lints — the "auto trait impl removed" lint is one of `cargo-semver-checks`' oldest lints and it predates a lot of these practices. For example, at the time it was written, `cargo-semver-checks` didn't recognize `#[doc(hidden)]` at all, which is why its test coverage never included it.

I'm happy to make the tweaks myself if you'd rather move on to the next thing. I'd recommend https://github.com/obi1kenobi/cargo-semver-checks/issues/1660 which is a bigger project and will require first working out the desired set of lints, then adding some new schema in `trustfall-rustdoc-adapter` to support those lints, and finally writing the lints.

test_crates/auto_trait_impl_added/old/src/lib.rs
ivelieu: added.

src/lints/auto_trait_impl_added.ron
ivelieu: changed. my reasoning was the auto-traits page doesn't explicitly mention semver, but I can imagine the person seeing the error is probably less familiar with auto traits considering it's an opt-in lint.

test_crates/auto_trait_impl_added/old/src/lib.rs
ivelieu: good point, I didn't think of that. It feels a bit bad to be just copy/pasting it, but you're right it covers more test cases. I've left the comments in from the original.

src/lints/auto_trait_impl_added.ron
ivelieu: I've made the changes for both lints. I can see this is also used in four other lints currently (`derive_trait_impl_removed, partial_ord_enum_variants_reordered, partial_ord_enum_struct_variant_fields_reordered, partial_ord_struct_fields_reordered`), all of which have both `false` and `true` defined. should those be changed too?

ivelieu:
Appreciate the detailed review, thank you, I've made the changes you've asked. I can see the nightly build is now failing, the error doesn't make full sense to me, but I don't think it's related to the lint. I'll make a start on https://github.com/obi1kenobi/cargo-semver-checks/issues/1660 in the next few days.

src/lints/auto_trait_impl_added.ron
obi1kenobi: Sure, thank you! Just in a separate PR if possible, so we can merge this one independently.

test_crates/auto_trait_impl_added/old/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+    bar: Rc<RefCell<i64>>,
+}
+
+// The RefUnwindSafeStruct struct checks for RefUnwindSafe being removed.
+// However, the way it's constructed above means that it loses both
+// the UnwindSafe and the RefUnwindSafe traits.
+//
+// Manually put the UnwindSafe trait back in, so that our test case tests
+// only for RefUnwindSafe being removed. UnwindSafe was tested earlier
+// in this file.
```
obi1kenobi: This ended up being a bit of copy-pasta.
```suggestion
// The RefUnwindSafeStruct struct checks for RefUnwindSafe being added.
// However, the way it's constructed above means that it gains both
// the UnwindSafe and the RefUnwindSafe traits.
//
// Manually put the UnwindSafe trait back in, so that our test case tests
// only for RefUnwindSafe being added. UnwindSafe was tested earlier
// in this file.
```

obi1kenobi (approved):
Nice! Fixing a bit of copy-pasta in a comment and merging. Thanks!

obi1kenobi:
The nightly failure is unrelated, I'll take care of it. Just need another nightly to get released, then we'll switch to the newer rustdoc JSON format that was just made, and that'll pick up a fix along the way.

PR #1697, opened 2026-08-11
Validate sort keys for singleton lint results

### LINKED ISSUE #1696: `cargo test`'s span-output validation for lints never executes when every test-crate pair yields fewer than 2 results (61 of 254 lints today)

All references are at `main` @ a1dda5c3.

1. The only enforcement of "a valid query must output `span_filename` + `span_begin_line`, or an `ordering_key`" is the sort-key closure `key_func` at `src/query.rs:1041-1092` (panic arms at `:1081-1089`). It is invoked in exactly one place, `value.sort_unstable_by_key(key_func)` at `src/query.rs:1094`, and empty result vectors are filtered out before that at `src/query.rs:1030`. `grep -rn span_filename src/ --include='*.rs'` shows no other `.rs` site outside `src/query.rs`; `src/check_release.rs` has zero hits.

2. `sort_unstable_by_key` performs no key evaluations for slices shorter than 2, so the closure's validation is dead code for any lint whose crate pairs each produce at most one result:

   ```rust
   // probe.rs
   use std::cell::Cell;
   fn main() {
       for n in 0..4usize {
           let calls = Cell::new(0usize);
           let mut v: Vec<usize> = (0..n).rev().collect();
           v.sort_unstable_by_key(|x| { calls.set(calls.get() + 1); *x });
           println!("len={} key_fn_calls={}", n, calls.get());
       }
   }
   ```

   ```
   $ rustc -O probe.rs -o probe && ./probe   # rustc 1.95.0
   len=0 key_fn_calls=0
   len=1 key_fn_calls=0
   len=2 key_fn_calls=2
   len=3 key_fn_calls=6
   ```

3. Counting result-vector lengths in the committed `test_outputs/query_execution/*.snap` files (count `    {` result-object openings per crate-pair array): **61 of 254 lints have at most one result in every crate pair**, e.g. `attribute_proc_macro_missing` (`[1]`), `exported_function_parameter_count_changed` (`[1]`), `enum_struct_variant_field_added` (`[1, 1, 1, 1]`). For these, the validation has never run.

4. End-to-end: I followed CONTRIBUTING.md's "Adding a new lint" walkthrough (`./scripts/make_new_lint.sh`) on a fork, keeping the scaffold defaults `per_result_error_template: Some("TODO")` and `witness: None`, and wrote a query that outputs neither `ordering_key` nor span data and yields exactly one result in one crate pair (it matches a private-struct rename, so it is semver-neutral, returns nothing on identical crate pairs, and no sibling lint's snapshot is affected). The full CI matrix passes — all 7 `rust-tests` legs (1.93/1.94/1.95/1.96/stable/beta/nightly) green: branch https://github.com/glaziermag/cargo-semver-checks/tree/demo-missing-span-outputs, diff vs upstream main https://github.com/obi1kenobi/cargo-semver-checks/compare/main...glaziermag:cargo-semver-checks:demo-missing-span-outputs, run https://github.com/glaziermag/cargo-semver-checks/actions/runs/31437680941. (The branch also carries two fork-only `ci.yml` accommodations needed to run this CI on a fork at all — release-automation jobs trimmed, and the issue-#902 disable-check pinned to this repo so `gh issue view 902` resolves on forks. Neither affects the test suite.)

### Actual Behaviour

`cargo test` passes for a registered lint that outputs neither `ordering_key` nor both span outputs, whenever each of its test-crate pairs yields ≤1 result.

Where an invalid singleton-shaped lint *does* get caught today, it is caught incidentally with misleading errors, never the documented one: 54 of the 61 singleton-shaped lints have a `per_result_error_template` referencing `{{span_filename}}`/`{{span_begin_line}}`, so stripping spans fails via the strict-mode handlebars error (`could not materialize template`, `src/query.rs:1122-1131` with `set_strict_mode(true)` at `src/templating.rs:147`), and the other 7 have witnesses and fail via `unreachable!("Missing span_filename String, this should be validated above")` at `src/query.rs:1142-1148`. The scaffold-default template `Some("TODO")` references no span fields, so a brand-new lint hits the fully-silent path demonstrated above.

### Expected Behaviour

CONTRIBUTING.md promises this is validated at test time:

- "Adding a new lint": "For almost all lints, make sure your lint outputs `span_filename` and `span_begin_line` in order to be a valid lint" (CONTRIBUTING.md:195-196), with the manifest-lint `ordering_key` exception.
- "Troubleshooting": "If your lint fails with an error similar to the following: … panicked at 'A valid query must output both `span_filename` and `span_begin_line`. …'" (CONTRIBUTING.md:381-397).

Expected: that panic fires for every result of every lint's test, independent of how many results each crate pair yields — e.g. by evaluating the key for every element unconditionally (map results through the key function into a `Vec` first, then sort — a few lines in `src/query.rs`). That would also make the witness path's "this should be validated above" assumption at `src/query.rs:1142-1148` actually hold.

Not applicable — this is a bug in the repo's own test suite rather than the installed CLI. Environments used: probe compiled with rustc 1.95.0 (aarch64-apple-darwin); end-to-end demonstration ran this repo's unmodified `rust-tests` CI matrix (ubuntu-latest, toolchains 1.93/1.94/1.95/1.96/stable/beta) on a fork.

### Build Configuration

None relevant.

### Additional Context

Two adjacent observations in the same block, noted here rather than filed separately:

- The comment at `src/query.rs:1045-1047` says "any output name with the `ordering_key` prefix works in practice", but `src/query.rs:1050` branches on the exact name (`elem.contains_key("ordering_key")`), so a lint outputting only `ordering_key1`/`ordering_key2` would panic as invalid rather than sort by them.
- The current panic text at `src/query.rs:1088` ("must either output an explicit `ordering_key`, or output both…") no longer matches the older message quoted in CONTRIBUTING.md:386 ("must output both…").

I searched for prior reports before filing: `span_filename span_begin_line` (hits #1267, #175, #243 — unrelated to this validation), and sort/validation-phrased searches (no hits).

Disclosure: this issue was researched and drafted with AI assistance. Every cited fact — the line references, the probe output, the 61/254 count, and the CI demonstration — was independently re-derived and verified before filing.

Guflly:
`sort_unstable_by_key` does not call the key function for vectors shorter than two elements, so invalid singleton query outputs skipped the ordering and span validation.

This evaluates each result's key before sorting, which keeps ordering deterministic and makes the existing validation run for every result. The regression covers a singleton result without ordering data.

Tests:
- `cargo test --lib singleton_query_result_without_ordering_data_is_rejected`
- `cargo clippy --lib --tests --no-deps -- -D warnings --allow deprecated`
- `cargo fmt -- --check`

Fixes #1696

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -856,6 +856,77 @@ mod tests {
         .join("\n\n")
     }
 
+    fn sort_by_key_evaluating_each<T, K: Ord>(
+        values: &mut Vec<T>,
+        mut key_func: impl FnMut(&T) -> K,
+    ) {
+        let mut keyed_values: Vec<_> = values
+            .drain(..)
+            .map(|value| (key_func(&value), value))
+            .collect();
+        keyed_values.sort_unstable_by(|(left, _), (right, _)| left.cmp(right));
+        values.extend(keyed_values.into_iter().map(|(_, value)| value));
+    }
+
+    #[derive(Clone, Eq, PartialEq, Ord, PartialOrd)]
+    enum SortKey {
+        Span(Arc<str>, usize),
+        Explicit(Vec<Arc<str>>),
+    }
+
+    fn query_result_sort_key(elem: &BTreeMap<String, FieldValue>) -> SortKey {
+        // Queries should either:
+        // - define `span_filename` and `span_begin_line` values where the lint is being raised,
+        //   which will then define a total order of results for that query on that crate.
+        // - define explicit ordering keys, canonically named `ordering_key`,
+        //   `ordering_key1`, `ordering_key2`, etc., even though any output name
+        //   with the `ordering_key` prefix works in practice. Those keys form a
+        //   composite ordering key by being sorted lexicographically by name,
+        //   then compared lexicographically by their string values, or
+        if elem.contains_key("ordering_key") {
+            let mut ordering_key_names: Vec<_> = elem
+                .keys()
+                .filter(|key| key.starts_with("ordering_key"))
+                .collect();
+            ordering_key_names.sort_unstable();
+            let ordering_keys = ordering_key_names
+                .into_iter()
+                .map(|key| {
+                    let value = elem
+                        .get(key)
+                        .unwrap_or_else(|| panic!("{key} output missing from result"));
+                    Arc::clone(
+                        value
+                            .as_arc_str()
+                            .expect("ordering_key output was not a string"),
+                    )
+                })
+                .collect();
+            SortKey::Explicit(ordering_keys)
+        } else {
+            let filename = elem.get("span_filename").map(|value| {
+                value
+                    .as_arc_str()
+                    .expect("`span_filename` was not a string")
+            });
+            let line = elem
+                .get("span_begin_line")
+                .map(|value: &FieldValue| value.as_usize().expect("begin line was not an int"));
+            match (filename, line) {
+                (Some(filename), Some(line)) => SortKey::Span(Arc::clone(filename), line),
+                (Some(_filename), None) => panic!(
+                    "No `span_begin_line` was returned by the query, even though `span_filename` was present. A valid query must either output an explicit `ordering_key`, or output both `span_filename` and `span_begin_line`. See https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CONTRIBUTING.md for details."
+                ),
+                (None, Some(_line)) => panic!(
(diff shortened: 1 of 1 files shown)


obi1kenobi:
This is an overkill solution for what is arguably not really even a bug. Please don't just have your AI agent send PRs without you reviewing the output first. Helping the project by contributing your good judgment is welcome; contributions of tokens only are not helpful — I also have a source of tokens.

